PR #3, opened 2019-05-28
Interpolate transforms

joelkoz:
Re-opening this PR on the SignalK repo.  I removed the "gauge specific" transforms like KelvinToCelsius.


src/transforms/interpolate.h
```diff
@@ -0,0 +1,51 @@
...
+#include "transforms/transform.h"
+
+
+/**
+ * Interpolate uses a collection of input/output samples that approximate a non-linear curve.
+ * The output is the linear interpolation between the two nearest sample points.
+ * It is used primarily for non-linear analog gauges such as temperature gauges using
+ * Thermocouples 
+ */
+class Interpolate : public OneToOneTransform<float> {
```
mairas: So this class takes a bunch of x,y pairs and linearly interpolates between the closest pair? I wonder could e.g. PiecewiseLinear be a more descriptive name? Interpolation led me to think about interpolating between consecutive input samples and I couldn't understand why you'd want to do that. :-)

PR #12, opened 2019-05-29
Chainable connect from()

joelkoz:
This PR is for a chainable call version of connectFrom() that is added to the SymmetricTransform().

This hacked approach was chosen over alternative solutions for its simplicity of use by end user.


src/transforms/transform.h
```diff
@@ -76,6 +76,37 @@ class SymmetricTransform : public ValueConsumer<T>, public Transform<T> {
...
+  /**
+   * A convenience method that allows up to five producers to be
+   * quickly connected to the input of the ValueConsumer side of this 
+   * transform.  The first producer will be connected to input
+   * channel zero, the second one to input channel 1, etc.
+   * "this" is returned, which allows the ValueProducer side
+   * of this transform to then be wired to other transforms via
+   * a call to connectTo().
+   */
+  SymmetricTransform<T>* connectFrom(ValueProducer<T>* pProducer0,
```
mairas: Did you consider declaring the function variadic (with a variable length argument list) instead? Your approach has the benefit of being more strongly typed.

joelkoz: I did consider that, yes.  I also considered "packed parameters" briefly.  The former would suffer from type checking problems as you mentioned.  The second approach MIGHT work, but it is a feature of C++ that is new to me, and I didn't really have time to digest it.  I admit this is a very hacky approach.  I figured we'd get it into the API, and then I or someone else more well versed in C++'s newer features could refactor later.

PR #20, opened 2019-06-06
Update examples

joelkoz:
Added verbose comments to examples to aid new users


mairas:
Looks good to me! Could @ba58smith review the comments as well? Butch has been active in writing documentation before.

ba58smith:
I will, but it will be a few days. (Hard to stop diving in St. Croix!) Go
ahead and commit Joel's changes and I  will review and comment on them when
I can.

Butch

examples/analog_input.cpp
```diff
@@ -1,26 +1,61 @@
...
+  // Create a "device" that is the source of our data
+  auto* pAnalogInput = new AnalogInput();
+
+
+  // Create a "transform" that can modify the data and/or broadcast it
+  // over the SignalK network.
+  const float multiplier = 1.0;
+  const float offset = 0.0;
+  auto* pTransform = new Linear(sk_path, multiplier, offset, config_path);
+
```
ba58smith: By specifying the multiplier and offset here, and as `const`, does that mean they can't be changed at runtime by the config UI?

joelkoz: No. Each transform has their own local copy of these values. It is the local copies that are used by the transform and retrieved and set by the UI.

Here, those variables are just used as initial values.  The parameters are "passed by value" and saved to the local copies.  So, there is no difference between using a "const float", or just typing the literal value directly in the list of parameters. I only split them out so they could be given a label for learning purposes. In practice, you'd just type: 

`new Linear("my.sk.path", 1.0, 0.0, "my/config/path")`

PR #22, opened 2019-06-09
Schema fixes

mairas:
A bunch of different things. Might be easier to review commit-by-commit.

- Some minor schema bugs fixed (loading `Frequency` configuration, matching expected used keys in `ws_config`, etc). 
- All schema are now saved in PROGMEM.
- `SymmetricTransform::connectFrom` fixed
- Removed the `connect_*to1` helpers and updated the wiring_helpers functions

Once you've checked this out, I'd like to bump the version up to 0.4.0 because the interface is somewhat incompatible after the `connect_*to1` changes.

### DIFF
--- a/src/devices/onewire_temperature.cpp
+++ b/src/devices/onewire_temperature.cpp
@@ -117,7 +117,6 @@ void OneWireTemperature::enable() {
   }
 }
 
-
 void OneWireTemperature::update() {
   dts->sensors->requestTemperaturesByAddress(address.data());
 
@@ -139,19 +138,18 @@ JsonObject& OneWireTemperature::get_configuration(JsonBuffer& buf) {
   return root;
 }
 
+static const char SCHEMA[] PROGMEM = R"({
+    "type": "object",
+    "properties": {
+        "address": { "title": "OneWire address", "type": "string" },
+        "found": { "title": "Device found", "type": "boolean", "readOnly": true },
+        "value": { "title": "Last value", "type" : "number", "readOnly": true }
+    }
+  })";
 
 String OneWireTemperature::get_config_schema() {
-   return R"({
-      "type": "object",
-      "properties": {
-          "address": { "title": "OneWire address", "type": "string" },
-          "found": { "title": "Device found", "type": "boolean", "readOnly": true },
-          "value": { "title": "Last value", "type" : "number", "readOnly": true }
-      }
-   })";
+  return FPSTR(SCHEMA);
 }
-
-
 
 bool OneWireTemperature::set_configuration(const JsonObject& config) {
   if (!config.containsKey("address")) {
--- a/src/devices/onewire_temperature.cpp
+++ b/src/devices/onewire_temperature.cpp
@@ -98,7 +98,7 @@ OneWireTemperature::OneWireTemperature(
     bool success = dts->get_next_address(&address);
     if (!success) {
       debugE("FATAL: Could not find an available OneWire device");
-      failed = true;
+      found = false;
     } else {
       dts->register_address(address);
     }
@@ -106,13 +106,13 @@ OneWireTemperature::OneWireTemperature(
     bool success = dts->register_address(address);
     if (!success) {
       debugE("FATAL: Unable to register OneWire device");
-      failed = true;
+      found = false;
     }
   }
 }
 
 void OneWireTemperature::enable() {
-  if (!failed) {
+  if (found) {
     app.onRepeat(1000, [this](){ this->update(); });
   }
 }
@@ -135,7 +135,7 @@ JsonObject& OneWireTemperature::get_configuration(JsonBuffer& buf) {
   char addr_str[24];
   owda_to_string(addr_str, address);
   root.set("address", addr_str);
-  root.set("failed", failed);
+  root.set("found", found);
   return root;
 }
 
@@ -145,6 +145,7 @@ String OneWireTemperature::get_config_schema() {
       "type": "object",
       "properties": {
           "address": { "title": "OneWire address", "type": "string" },
+          "found": { "title": "Device found", "type": "boolean", "readOnly": true },
           "value": { "title": "Last value", "type" : "number", "readOnly": true }
       }
    })";
--- a/src/devices/onewire_temperature.cpp
+++ b/src/devices/onewire_temperature.cpp
@@ -97,7 +97,7 @@ OneWireTemperature::OneWireTemperature(
     // previously unconfigured device
     bool success = dts->get_next_address(&address);
     if (!success) {
-      debugE("FATAL: Could not find an available OneWire device");
+      debugE("FATAL: Could not find available OneWire devices");
       found = false;
     } else {
       dts->register_address(address);
@@ -105,7 +105,7 @@ OneWireTemperature::OneWireTemperature(
   } else {
     bool success = dts->register_address(address);
     if (!success) {
-      debugE("FATAL: Unable to register OneWire device");
+      debugE("FATAL: OneWire device(s) found, but could not be registered");
       found = false;
     }
   }
--- a/src/net/networking.cpp
+++ b/src/net/networking.cpp
@@ -77,16 +77,16 @@ void Networking::set_hostname(String hostname) {
   this->hostname->set(hostname);
 }
 
+static const char SCHEMA[] PROGMEM = R"({
+    "type": "object",
+    "properties": {
+        "hostname": { "title": "Device hostname", "type": "string" }
+    }
+  })";
 
 String Networking::get_config_schema() {
-   return R"({
-      "type": "object",
-      "properties": {
-          "hostname": { "title": "Device hostname", "type": "string" }
-      }
-   })";
+  return FPSTR(SCHEMA);
 }
-
 
 JsonObject& Networking::get_configuration(JsonBuffer& buf) {
   JsonObject& root = buf.createObject();
--- a/src/net/networking.cpp
+++ b/src/net/networking.cpp
@@ -82,7 +82,7 @@ String Networking::get_config_schema() {
    return R"({
       "type": "object",
       "properties": {
-          "hostname": { "title": "Network SSID", "type": "string" }
+          "hostname": { "title": "Device hostname", "type": "string" }
       }
    })";
 }
--- a/src/net/networking.cpp
+++ b/src/net/networking.cpp
@@ -45,10 +45,10 @@ void Networking::setup(std::function<void(bool)> connection_cb) {
   wifi_manager->setDebugOutput(false);
   #endif
   AsyncWiFiManagerParameter custom_hostname(
-    "hostname", "Set hostname", this->hostname->get().c_str(), 20);
+    "hostname", "Set ESP device hostname", this->hostname->get().c_str(), 20);
   wifi_manager->addParameter(&custom_hostname);
 
-  if (!wifi_manager->autoConnect("Unconfigured Sensor")) {
+  if (!wifi_manager->autoConnect("Unconfigured SensESP Device")) {
     debugE("Failed to connect to wifi and config timed out.");
     ESP.restart();
   }
@@ -80,7 +80,7 @@ void Networking::set_hostname(String hostname) {
 static const char SCHEMA[] PROGMEM = R"({
     "type": "object",
     "properties": {
-        "hostname": { "title": "Device hostname", "type": "string" }
+        "hostname": { "title": "ESP device hostname", "type": "string" }
     }
   })";
 
--- a/src/net/ws_client.cpp
+++ b/src/net/ws_client.cpp
@@ -91,16 +91,16 @@ void WSClient::on_receive_delta(uint8_t * payload) {
   debugD("Received payload: %s", (char*)payload);
 }
 
-bool WSClient::get_mdns_service(String& host, uint16_t& port) {
+bool WSClient::get_mdns_service(String& server_address, uint16_t& server_port) {
   // get IP address using an mDNS query
   int n = MDNS.queryService("signalk-ws", "tcp");
   if (n==0) {
     // no service found
     return false;
   } else {
-    host = MDNS.IP(0).toString();
-    port = MDNS.port(0);
-    debugI("Found server with IP/Port: %s:%d", host.c_str(), port);
+    server_address = MDNS.IP(0).toString();
+    server_port = MDNS.port(0);
+    debugI("Found server %s (port %d)", server_address.c_str(), server_port);
     return true;
   }
 }
@@ -113,17 +113,17 @@ void WSClient::connect() {
 
   connection_state = connecting;
 
-  String host = "";
-  uint16_t port = 80;
+  String server_address = "";
+  uint16_t server_port = 80;
 
-  if (this->host.length() == 0) {
-    get_mdns_service(host, port);
+  if (this->server_address.length() == 0) {
+    get_mdns_service(server_address, server_port);
   } else {
-    host = this->host;
-    port = this->port;
+    server_address = this->server_address;
+    server_port = this->server_port;
   }
 
-  if ((host.length() > 0) && (port > 0)) {
+  if ((server_address.length() > 0) && (server_port > 0)) {
     debugD("Websocket client starting");
   } else {
     // host and port not defined - wait for mDNS
@@ -133,20 +133,20 @@ void WSClient::connect() {
 
   if (this->polling_href != "") {
     // existing pending request
-    this->poll_access_request(host, port, this->polling_href);
+    this->poll_access_request(server_address, server_port, this->polling_href);
     return;
   }
 
   if (this->auth_token == NULL_AUTH_TOKEN) {
     // initiate HTTP authentication
     debugD("No prior authorization token present.");
-    this->send_access_request(host, port);
+    this->send_access_request(server_address, server_port);
     return;
   }
-  this->test_token(host, port);
+  this->test_token(server_address, server_port);
 }
 
-void WSClient::test_token(const String host, const uint16_t port) {
+void WSClient::test_token(const String server_address, const uint16_t server_port) {
 
   // FIXME: implement async HTTP client!
   WiFiClient client;
@@ -154,7 +154,7 @@ void WSClient::test_token(const String host, const uint16_t port) {
 
   debugD("Testing token");
 
-  String url = String("http://") + host + ":" + port + "/signalk/v1/api/";
+  String url = String("http://") + server_address + ":" + server_port + "/signalk/v1/api/";
   http.begin(client, url);
   String full_token = String("JWT ") + auth_token;
   http.addHeader("Authorization", full_token.c_str());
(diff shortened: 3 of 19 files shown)


src/devices/onewire_temperature.cpp
```diff
@@ -98,26 +98,25 @@ OneWireTemperature::OneWireTemperature(
...
       debugE("FATAL: Could not find an available OneWire device");
-      failed = true;
+      found = false;
     } else {
       dts->register_address(address);
     }
   } else {
     bool success = dts->register_address(address);
     if (!success) {
       debugE("FATAL: Unable to register OneWire device");
```
ba58smith: To make troubleshooting more efficient, how about making this message:
`FATAL: OneWire device(s) found, but could not be registered`
That tells the end user exactly what happened, and no one has to dig into it to see if the problem was that a device wasn't found in the first place, or was found, but couldn't be registered.

src/net/networking.cpp
```diff
@@ -77,17 +77,17 @@ void Networking::set_hostname(String hostname) {
   this->hostname->set(hostname);
 }
 
+static const char SCHEMA[] PROGMEM = R"({
+    "type": "object",
+    "properties": {
+        "hostname": { "title": "Device hostname", "type": "string" }
```
ba58smith: In SigkSens, and now SensESP, I've always been confused by the use of "hostname". Sometimes it seems like it's the SSID of the wifi network that the MCU is trying to connect to. Sometimes it seems like it's the name of the MCU. Sometimes it seems like it's the Signal K server. Now, it's been changed to "Device hostname". What device? The MCU? In SensESP, something of the class `Device` is a sensor, not an MCU, but I think in this case, hostname is referring to the MCU. But maybe not. It's all very confusing, all through the code.

src/net/ws_client.cpp
```diff
@@ -335,22 +335,24 @@ JsonObject& WSClient::get_configuration(JsonBuffer& buf) {
...
+static const char SCHEMA[] PROGMEM = R"({
+    "type": "object",
+    "properties": {
+        "sk_host": { "title": "SignalK Host", "type": "string" },
+        "sk_port": { "title": "SignalK host port", "type": "integer" },
+        "client_id": { "title": "Client id", "type": "string", "readOnly": true },
+        "token": { "title": "Server authorization token", "type": "string" },
+        "polling_href": { "title": "Server authorization polling href", "type": "string", "readOnly": true }
+    }
+  })";
```
ba58smith: Here's a further point of confusion with hostname: the use of the term "SignalK Host". If that term is referring to the SignalK Server, then why not use "SignalK Server"? Even in this one object (above), the word "host" and "server" are used to refer to what appears to be the same thing.

src/transforms/interpolate.cpp
```diff
@@ -9,7 +9,7 @@ Interpolate::Sample::Sample(float input, float output) : input{input}, output{ou
 }
 
 Interpolate::Sample::Sample(JsonObject& obj) : input{obj["input"]}, output{obj["output"]} {
```
ba58smith: Does Interpolate return an average of all the samples taken, or does it do some more complex math that actually interpolates a value somewhere in the range? I ask only in the interest of simplicity - if it returns an average, why not called it "Average" instead of "Interpolate"?

src/net/networking.cpp
joelkoz: +1 on this.  I obviously had it wrong also as I called it SSID in the original schema.  I just did some searching, and as near as I can tell, its SUPPOSED to be the domain name used in the mDSN system.  That is, if you call your device "foo", then you can find your sensor at http://foo.local.  I THINK.  It think it is ALSO the host name that would show up if you scan your list of network devices.  But - I don't think this works.  I just used a network scan utility, and my mDNS and "host name" both came up as "esp8266-2809f3", despite the fact that I set my host name and rebooted my device.

I wonder if this feature is even working?

src/devices/onewire_temperature.cpp
mairas: Will fix!

src/net/networking.cpp
mairas: The hostname setting _was_ working, at least for mDNS and display purposes. Also some of the system paths like `sensors.$HOSTNAME.freeMem` are updated based on the hostname. I'll verify this properly before merging.

As for the naming, a fair point. What should we call the MCU? I'm not sure "MCU" is clear enough. As Brian pointed out, "device" is not good. "Sensor" is somewhat overloaded, too. "Microcontroller" or "ESP"?

src/net/ws_client.cpp
mairas: OK, that's pretty horrible. Will fix!

src/transforms/interpolate.cpp
mairas: It's definitely doing linear interpolation, but on an axis other than what I expected. To me the name would imply something comparable to `MovingAverage`, i.e. taking adjacent samples of an observable and doing the interpolation in time domain.

How about "CalibrationCurve" or "CorrectionCurve"? You're basically defining a pointwise-linear curve and doing interpolation on that curve.

src/net/ws_client.cpp
```diff
@@ -335,22 +335,24 @@ JsonObject& WSClient::get_configuration(JsonBuffer& buf) {
...
-          "polling_href": { "title": "Server authorization polling href", "type": "string" }
-      }
-   })";
+  return FPSTR(SCHEMA);
 }
 
 
 bool WSClient::set_configuration(const JsonObject& config) {
-  String expected[] = {"sk_host", "sk_port", "token"};
+  String expected[] = {"sk_host", "sk_port", "token", "client_id", "polling_href"};
```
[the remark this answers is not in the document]
mairas: Yeah ok. I was just assuming you shouldn't be trying to set undefined attributes in ArduinoJson. Will fix.

src/transforms/transform.h
```diff
@@ -92,21 +92,21 @@ class SymmetricTransform : public ValueConsumer<T>, public Transform<T> {
                                      ValueProducer<T>* pProducer3 = NULL,
                                      ValueProducer<T>* pProducer4 = NULL) {
 
-      this->connectFrom(pProducer0, 0);
+      pProducer0->connectTo(this, 0);
```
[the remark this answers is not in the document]
mairas: I don't think `connectFrom` with that signature was defined any longer. It didn't compile for me, anyway. I guess it'd be better to implement the `connectFrom` version, though. I'll have a look.

src/wiring_helpers.cpp
```diff
@@ -17,165 +17,88 @@
...
-    f1, f2,
-    diff
-  );
 
-  // this is the same as above, but averaged over 10 s
-  seapp->connect_1to1<Difference, MovingAverage>(
-    diff,
-    new MovingAverage("propulsion.left.fuel.averageRate", 10, 1., "/sensors/fuel_average_rate")
-  );
+  diff->connectFrom(freq1, freq2);
```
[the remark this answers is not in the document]
mairas: Why not. Will fix.

src/net/networking.cpp
joelkoz: Perhaps "Network name"?.  But, now that you mention its ALSO used in synthesizing signal K paths, then "Device name" may in fact be the most appropriate, but simply needs to be documented somewhere.

Json Schema supports a "description" field, which is supposed to be a "long description" vs. the shorter "title" field.  I'll test to see if Json Editor supports it.  If it does, we can use "description" as a way to give a detailed description of all of these fields.

src/net/ws_client.cpp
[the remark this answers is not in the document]
joelkoz: Yeah - there is a difference between "these fields are required" and "these fields are not part of the schema."  

I know I have seen "Json Schema validators" that can check an object against a schema.  You can mark fields in the schema as required or not.  The validator can then reject objects that contain keys that are NOT in the schema at all.  I can look around for a library to do this.  If I find it, its probably better to locate it in the save_configuration() high level method, and remove these localized validations altogether.

src/transforms/interpolate.cpp
joelkoz: Just so we are all on the same page - the samples have NOTHING to do with the actual values flowing thru the system.  The "samples" that make up the curve are static.  They are fixed values supplied by the programmer (though they can be modified by the web UI).  As actual values are presented to interpolate, all you get is a "this is the y value for your x input."  That input value does not change future values of the curve.

That being said, while it is possible to use Interpolate with just two sample points (thus being much more like "Linear"), its most common use case is for sure a curve, so the word "Curve" should probably be in the title.  Maybe we name this "LinearCurve" or "CurveInterpolator" or something like that?

src/transforms/transform.h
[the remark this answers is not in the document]
joelkoz: connectFrom() is still there - its defined in ValueConsumer.  The multi-input connectFrom() had to be defined in transform.h because it only works on SymmetricTransforms.

The overloaded connectFrom() should work and not get confused over function signatures because one is ValueConsumer::connectFrom(ValueProducer, int), and the other is connectFrom(ValueProducer, ValueProducer, ValueProducer...)

src/net/networking.cpp
ba58smith: To hopefully help resolve this, here are each of the things we need to have clear names for, in the code, in the documentation, and in the UI:

1. An individual sensor - something that takes a reading of some physical thing and sends that reading to the microcontroller. I propose we call that a "sensor", and that we don't call anything else a "sensor".

2. The microcontroller. I called this an MCU in SigkSens because it was eventually supposed to work with a lot of different microcontrollers. But in SensESP, I think it's safe to use the term "ESP", because that's clearly the target microcontoller. They'll all be ESP8266's or ESP32's for the foreseeable future.

3. The wifi network that the microcontroller connects to. "SSID" is good, but there is a small potential for confusing the SSID of the network that the microcontroller connects to with the SSID "Unconfigured Sensor". (And, BTW, can we change that to "Unconfigured ESP"?) So maybe "serverSSID", or "networkSSID", or something with a bit more resolution than just SSID?

4. The Signal K server. How about "signalKServer", or "skServer", or some variation on that?

"hostname" is clearly an ambiguous term. I think we should eradicate it from SensESP, except where some other library (Wifi Manager, etc.) uses it and we have no choice.

src/transforms/interpolate.cpp
ba58smith: What's the use case? To be able to get a value from an analog gauge with an output that's not linear? (As an example.) i.e., at 0.0V, the oil pressure is 0 psi, at 5.0V the oil pressure is 60 psi, but at X measured points in between 0.0V and 5.0V, the psi reading of the gauge isn't on that same line?

src/net/networking.cpp
joelkoz: Good stuff.  @ba58smith is right - we need to keep in mind that it will be quite common to have multiple sensors connected to a single MCU.

First, there are technically two more things:

The "network name" that device broadcasts to the network its connected to (this is actually referred to as "hostname", and is rarely seen by humans that are not administering the network), and the mDNS name the device broadcasts to allow you to connect to it without having to discover the IP address (e.g. the http://"someName.local").

Right now, "hostname" is used for the network name, mDNS name, and parts of the signalK path that the MCU broadcasts for its free memory, uptime, etc.

I think its OK for certain settings to do "double duty" or even tripple duty.  For example, I think it makes perfect sense for the MCU name to be used as both the network name, the mDNS name.  It also makes sense to have the MCU name in those signalK paths that are specific to the entire device.

And now that I've typed all that using "MCU", I vote for "MCU" as the name of the board vs. ESP or something like that.  Everything I've seen in online discussions about these boards refer to them as MCU, so I think that is the de facto standard already.

src/transforms/interpolate.cpp
joelkoz: The water temperature sender on my generator changes its resistance value in relationship to the temperature.  This relationship is non-linear however.  As it heats up, the difference in ohms "per degree" gets less and less.  My understanding is that this is how ALL thermistors and thermocouplers work.

If you look at transform/TemperatureInterpreter.h, you'll see this use case.  I put that transform in there from my AnalogGauge project.  It may be more appropriate to move it to the examples directory however.  I have no idea how common the values of my particular sender are.

joelkoz: Actually - I just remembered - I put most of the main.cpp from my AnalogGauge project into examples/temperature_sender.cpp to show how many of the transforms are used.  Take a peek at that, and note that the TemperatureInterpreter it references inherits from Interpolate.

src/net/networking.cpp
ba58smith: Agree on using the MCU name as its network name, its mDNS name, and especially including it in the SK path. I already have two Wemos D1 Minis on my network, and being able to see which one of them the "systemHz" and "freemem" are coming from is important.

Also agree on using "MCU" when we refer to the ESP8266, ESP32, or whatever microcontroller someone is using. It will always be correct, even if someone does a fork of SensESP for the Arduino Mega.

joelkoz: I have confirmed that adding a "description" field to the Json Schema is indeed supported by the Json Editor.  @mairas once you get this PR checked in, we can open a separate PR for adding a description to all of the schemas (or, at least the ones requiring more explanation).  That will then allow us to document what each of these fields mean and have it presented in the web UI.

joelkoz: Thinking about this a little more: "MCU" it the board one uses, but from a layman's point of view, isn't "Device name" really more appropriate?  You build this device and stick it on your network.  A single device may have one or more sensors.  You "device" is your "temperature sensor" (if that is all the MCU does in an implementation), or its your "Bilge watcher" if it has a bilge pump monitor along with a water level sensor.  As far as explaining the field better, we can use the "description" field of the json schema to supplement the web UI.  We could also change the config path from "/system/networking" to "/system/general" to indicate its more than just a networking configuration.

Any thoughts on this @mairas or @ba58smith?

src/transforms/interpolate.cpp
ba58smith: "CurveInterpolator" works for me.

src/net/networking.cpp
ba58smith: "Device" is so non-descript. Here's one dictionary's definition: "a thing made or adapted for a particular purpose, especially a piece of mechanical or electronic equipment." A boat has a hundred "devices" on it. You might as well call it a  "Gadget" or a "Thingamajig". For someone who doesn't work with SensESP on a regular basis, they'll have to remind themselves what a "Device" is every time they dive back in. OTOH, "MCU" means something, and as you pointed out earlier, it has a rather common meaning in the small electronics world.
To your point about this thing that you build being a "device", that's true - but the part of the device that contains and runs SensESP, and that communicates with the network, and that handles the configuration - that's the microcontroller. My Engine Monitor "device" consists of an MCU, 5 temperature sensor, two voltage sensors, and a digital input sensor for the tachometer. "MCU" is a boring acronym, but it's meaningful. "Device" isn't.

mairas: OK, tested a bit, and I can verify that the hostname is no longer advertised on mDNS. So, something has broken at some point.

I believe this was the document I followed when I originally implemented the feature on SigkSens: https://tttapa.github.io/ESP8266/Chap08%20-%20mDNS.html. It seems to be quite different from the more authoritative examples at https://github.com/esp8266/Arduino/tree/master/libraries/ESP8266mDNS, though. 

Hostname setting is still correctly reflected on the SK paths and e.g. server auth request, so it's not completely broken, IMO. Only partially. :-p

As for the terminology, I think "hostname" has a pretty specific and clear meaning (the human-readable name of a network device). The main problem is that the code has been ambiguous regarding which hostname is being referred to where: SK server hostname, MCU (device?) hostname, or what. I'd like to keep the term but disambiguate it according to context.

My opinions on Brian's naming suggestions: 1. fully agree on sensor. 2. I think many people might not be familiar with "MCU", especially if English is not their first language. Microcontroller is clear but a mouthful. I kinda like "ESP" but maybe we can also use "ESP device" if that fits the context better? 3. Wifi network can in my opinion be "Wifi network" or "SSID", or "network SSID", or "network name". I think any of those can be used interchangably, depending on the context. In the code the terminology should be as accurate as possible. I'll rename the SSID "Unconfigured SensESP device", if you're OK with that, but only in a later PR.

src/transforms/interpolate.cpp
mairas: I quite like CurveInterpolator. But, how about PiecewiseLinearCurve? That would be accurate and would suit my old DSP geek sensibilities. :-)

src/net/ws_client.cpp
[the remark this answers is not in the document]
mairas: I removed the polling_href but left token and client_id in place. I think the current implementation would set them to empty strings, if they're not set.

src/transforms/transform.h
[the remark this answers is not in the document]
mairas: OK, I can see it. I don't know why the compiler didn't get it. I agree that calling `connectFrom` would be more consistent. I'll keep it as is to get the PR in and then fix later, if that's OK with you.

src/wiring_helpers.cpp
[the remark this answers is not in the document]
mairas: The signal propagation graph seemed to be so complex that I couldn't easily carve out simple chains. I'll leave it as is.

src/net/networking.cpp
ba58smith: "As for the terminology, I think "hostname" has a pretty specific and clear meaning (the human-readable name of a network device). The main problem is that the code has been ambiguous regarding which hostname is being referred to where: SK server hostname, MCU (device?) hostname, or what. I'd like to keep the term but disambiguate it according to context."

I think that's a good solution (continue to use "hostname" where appropriate, but disambiguate each usage in the code and in the debug() messages).

ba58smith: Hadn't thought about non-native English speakers - my bad. Between "microcontroller", "ESP", and "ESP device", why not keep it simple and go with "ESP"? If someone wants to make a fork for non-ESP micrcontrollers, let THEM deal with what to call it!

ba58smith: "I'll rename the SSID "Unconfigured SensESP device", if you're OK with that..." 

I'm really not trying to be nit-picky, just consistent. If we end up calling the microcontroller an "ESP", then let's use "Unconfigured ESP" for the SSID. If you call it "ESP Device", then make the SSID "Unconfigured ESP Device". If you want to call it a "SensESP Device" (which is a mouthful), then the SSID should be "Unconfigured SensESP Device". IMO, of course. Just pick a name, and use it consistently, that's my point.

src/transforms/interpolate.cpp
joelkoz: I like CurveInterpolator the most of the suggestions put forth thus far.  I've not done any DSP work, thus PiecewiseLinearCurve isn't clear to me.

mairas:
I went through the networking code and renamed `host` and `port` variables to `server_address` and `server_port`. That should go a long way reducing the confusion.

Merging now.

src/transforms/interpolate.cpp
ba58smith: +1 for CurveInterpolator.

ba58smith:
"what about "this", which is a common term to C++ programmers. how about this_hostname to represent the MCU's own hostname?"

I don't think that's clear. You could end up with `this->this_hostname` in your code, and also `myDevice->this_hostname`, and also `myNetwork->this_hostname`. Arrggghhh....

PR #24, opened 2019-06-10
Moved TemperatureInterpreter to the examples directory

joelkoz:


examples/temperature_sender.cpp
```diff
@@ -6,7 +6,38 @@
...
-#include "transforms/temperatureinterpreter.h"
+#include "transforms/interpolate.h"
+
+
+
+/**
+ * A custom transform that takes a resistance value on ohms and returns the estimated 
+ * temperature in Kelvin.  Sample data in this example were taken from a Westerbeke generator
+ * temperature gauge.
+ */
```
ba58smith: I think the comment should mention that the sample data can be changed in the Config UI. Otherwise, people are likely to think they need to modify a source file and recompile for every instance of Interpolate. You might also mention something about the number of samples: you included 10, but is there any magic in that number? How many are enough? How many are too many? Is there a limit on how many you can use? (Maybe this should be in interpolate.h, as well, or instead of, here?)

joelkoz: Interpolate has no samples of its own. If you use it directly, it is worthless unless you use the web UI first to fill it up.

As for the number of samples - the correct answer depends on the actual curve you are approximating, as well as how accurate you want the results to be.  In this example, this descendant class defines samples that represent various points on the temperature gauge that I pulled from my boat.  I sampled values that included the lowest value that would make the gauge move, the highest value that had a detectable move before the gauge "pegged", as well as six visible positions on the gauge (and what resistance was required to make the gauge move to those six positions).  That is a total of eight actual values.  The other two are basically "zero" and "infinity" (or, at least a value so high the value would never be over it). Those values should mean Interpolate ALWAYS finds two points to use.

I actually have no idea if I needed that many samples or not.  Maybe I need more on the high end of the curve to make it more accurate.  I figured it was enough as it modeled what the real gauge was capable of displaying.

I can add some comments to Interpolate.h to answer the other fine details - two is the minimum for something to be functional, and "as many as you want without exhausting memory" is the upper limit.

ba58smith: "Interpolate has no samples of its own. If you use it directly, it is worthless unless you use the web UI first to fill it up."
But as a new user of SensESP, how do I know that, and how do I know where to add the samples, and how many to add, and so on? That's my point. I've looked at every file I can think of that's associated with this, and I don't know how anyone would understand how to use this, unless they wanted to use your data in temperatureinterpreter.cpp, which is almost certainly not going to work for very many analog gauges. So whether it's in the example, or tin temperatureinterpreter.h, or interpolate.h - I think it needs to be somewhere.

joelkoz: That is what this PR is.  I removed "TemperatureInterpreter.h" and .cpp from the main code base entirely.  I then added it directly into the "temperature_sender.cpp" file, which is in the examples directory.  Take a look at the [entire file now in context](https://github.com/SignalK/SensESP/blob/4e137afb3274f0903e5964125a04fd381229af39/examples/temperature_sender.cpp) and let me know what you think.

ba58smith: I look at temperature_sender.cpp and I think "In order to do something like this myself, I need to create a Class that descends from Interpolate, and its constructor needs to have TEN calls to addSample() in it, and the values need to be in my main.cpp file, which means I have to edit the source code and rebuild / reload every time I want to do anything with the samples."

I would have no idea that I could use a different number of samples, and no clue that I could load the samples in the Config UI. In fact, I still don't know what I'd see if I compiled this example, and then started the Config UI. Is it going to present me with 10 fields for entering data? What if I want to enter 15? (If that would be obvious by looking at the UI, sorry - I haven't gotten to that point yet.) I'm sure the comments in the example, or in interpolate.h, could make this much more clear.

joelkoz: > In order to do something like this myself, I need to create a Class that descends from Interpolate

That part is 100% true.  Not all classes in the framework are going to be simple to use.  You can use this code without pre-loading it up with samples.  If one don't understand how a voltage divider works, how its used to integrate with a sender, and how this code takes advantage of it, I don't think a few comments in the source code is going to help much.  It may warrant a completely separate tutorial.

As far as the UI is concerned - yes - it will present you a list of the samples.  You can remove or add an individual sample at will, as its defined as an "array of samples", and the editor know how to deal with arrays in general.

ba58smith: Now that I know that, do you mind if I take a stab at writing a comment that might be enough to let someone use Interpolate, or a descendant of it?

joelkoz: Absolutely.  I was going to suggest that.  I think you have something in mind for stuff like this, so  I'd rather answer questions for you, give you whatever explanations you need, then have you write the comments or documentation necessary.  I tend to take certain knowledge for granted.

Note that for the library code (Interpolate in this instance), there are special format specifiers used for generating the API documentation.  It follows the same formatting as JavaDocs if you ever saw or used those.  For example, instead of saying:

`Parameter X is used for bla bla`

You can type on its own line:

`@param X Bla bla`

and Doxygen (the utility we use to generate the API documentation) will understand you are talking about one of the parameters and will format and cross reference appropriately.  Take a look at ValueConsumer.h, the set_input() method for an example.  Complete [specs are here](http://www.doxygen.nl/manual/docblocks.html).

I am going to go ahead and merge this PR into master, then you can make a branch from that with your comments.  Really, feel free to do that to ANY of the classes you find.

ba58smith: Got it.

PR #25, opened 2019-06-10
docs: enhance debug msgs in the startup process

ba58smith:
Added several new debug() messages, and enhanced several more, all in the startup process. I changed one instance of the word "sensor" to "MCU", but if we decide to use "Device" instead of "MCU", I'll change it then.

### DIFF
--- a/src/net/networking.cpp
+++ b/src/net/networking.cpp
@@ -54,7 +54,7 @@ void Networking::setup(std::function<void(bool)> connection_cb) {
     ESP.restart();
   }
 
-  debugD("Connected to wifi, SSID: %s", WiFi.SSID().c_str());
+  debugI("Connected to wifi, SSID: %s", WiFi.SSID().c_str());
   connection_cb(true);
 
   if (should_save_config) {
--- a/src/net/networking.cpp
+++ b/src/net/networking.cpp
@@ -29,6 +29,7 @@ Networking::Networking(String config_path)
 void Networking::check_connection() {
   if (WiFi.status() != WL_CONNECTED) {
     // if connection is lost, simply restart
+    debugD("Wifi disconnected: restarting...");
     ESP.restart();
   }
 }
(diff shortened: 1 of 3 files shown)


src/net/ws_client.cpp
```diff
@@ -189,7 +189,7 @@ void WSClient::send_access_request(const String host, const uint16_t port) {
   JsonObject& req = buf.createObject();
   req["clientId"] = client_id;
   req["description"] =
-    String("SensESP sensor: ") + sensesp_app->get_hostname();
```
mairas: As mentioned in the other PR, I think MCU might not be the bestest term. In this particular context, I'd use "SensESP device".

ba58smith: OK, I'll make that change. But I don't know how. :-(
Can it be made right here in the PR somehow, and have that find its way back to all the places it needs to go for Git / GitHub's purpose?
Or do I need to update the file on my local computer, and create a commit for it, and add that to the PR? I have to say, I've had more trouble using Git / GitHub than I have with the SensESP source code!

src/net/networking.cpp
```diff
@@ -49,11 +50,11 @@ void Networking::setup(std::function<void(bool)> connection_cb) {
   wifi_manager->addParameter(&custom_hostname);
 
   if (!wifi_manager->autoConnect("Unconfigured Sensor")) {
-    debugE("Failed to connect to wifi and config timed out.");
+    debugE("Failed to connect to wifi and config timed out. Restarting...");
     ESP.restart();
   }
 
-  debugI("Connected to Wifi.");
+  debugD("Connected to wifi, SSID: %s", WiFi.SSID().c_str());
```
joelkoz: Any reason why this was changed from from an "information" msg to a "debug" message?  For the "Disconnected - restarting" on line 32, I think it should be an info message, if not a warning message (i.e. debugW).  I'd want to see that I lost my wifi connection at a specific time even in production. The "Connected to SSI" would be handy to see even if lower level debug information is turned off, so I'd vote for an "information" level (i.e. debugI).

ba58smith: Changed back to debugI().

src/net/ws_client.cpp
ba58smith: Change made.

PR #29, opened 2019-06-12
docs: enhance debug msgs in the startup process (#25)

ba58smith:
* Enhance debug msgs

* After review, one change each for @mairas and @joelkoz


ba58smith:
I could never resolve the conflicts to GitHub's satisfaction, so I'm starting over with a new branch.

PR #32, opened 2019-06-14
feature: Signalk output consumer

joelkoz:
This is the refactor of the code base to separate out the signalK output from the Transforms.  The major changes include:

1. No sk_path parameter on transforms
2. Passthrough() renamed SKOutput
3. All signalk related code moved to signalk folder
4. Transform now defined as having being both a Producer AND Consumer
6. typedefs *Transform are now SymmetricTransforms
7. system/signal_k is now signalk/signalk_delta.h

One advantage of change#4 is that the chainable version of ValueProducer::connectTo() was able to be re-defined as accepting asymmetric transforms. As long as the consumer's input and producer's output match, the other side of the transforms (producer's input and consumer's output) no longer matter.  This allows for things like DigitalInput (which is a ValueProducer&lt;int&gt;) to be connected to Frequency (which is a symmetric transform &lt;float to float&gt;.

One big downside of this refactor we didn't originally consider: the "sk_path" parameter is now housed in an object separate from the transforms.  This means if you want the end user to be able to CHANGE the signal k path using the web UI or the REST API, it needs to have its own config_path SEPARATE from the config_path of the transform.  The old method allowed for a single config path to handle the sk_path as well as any calibration parameters.  This makes the final code much more wordy.

Study carefully the code in wiring_helpers.cpp, as well as the examples temperature_sender.cpp and rpm_counter.cpp


src/signalk/signalk_output.h
```diff
@@ -0,0 +1,68 @@
...
+// to output SignalK data on the SignalK network.
+template <typename T>
+class SignalKOutput : public SignalKSource,
+                      public SymmetricTransform<T> {
+ public:
+  SignalKOutput() : SignalKOutput("") {}
+
+  SignalKOutput(String sk_path, String config_path="")
+    : SignalKSource(sk_path), SymmetricTransform<T>(config_path) {
+    Enable::setPriority(-5);
```
mairas: Just out of curiosity: why a high priority rather than low? Now it first outputs, then polls new values on each event loop iteration.

joelkoz: Absolutely no reason.  Other than note that higher values mean "first", so setting priority to -5 means this gets enabled LAST

PR #35, opened 2019-06-14
fixup: Rename devices to sensors

mairas:
Previously, the terminology related to devices and sensors was not quite consistent. Low-level hardware sensors were called devices, while the whole application running on the hardware was referred to as sensor or device or whatever. As discussed with @joelkoz and @ba58smith, low-level hardware sensors are now called sensors and the whole thing is "ESP device" or just "device" if the context permits.


src/sensors/onewire_temperature.cpp
```diff
@@ -33,24 +33,24 @@ bool string_to_owda(OWDevAddr* addr, const char* str) {
...
     int pin, String config_path)
-    : Device(config_path) {
+    : Sensor(config_path) {
   onewire = new OneWire(pin);
   sensors = new DallasTemperature(onewire);
   sensors->begin();
 
-  int num_devices = sensors->getDeviceCount();
-  debugI("OneWire devices found: %d", num_devices);
+  int num_sensors = sensors->getDeviceCount();
```
joelkoz: Should that method be renamed getSensorCount()?

src/sensors/sensor.h
```diff
@@ -0,0 +1,53 @@
...
+
+
+///////////////////
+// Sensors represent raw hardware thingies.
+
+class Sensor : virtual public Observable, public Configurable, public Enable {
+  public:
+    Sensor(String config_path="");
+
+  // Was used by sensesp_app::enable(). Now handled by Enable.  Deprecate?
```
joelkoz: I see now this method is used by more than the one spot I originally saw it.  We can get rid of this comment.

src/sensors/onewire_temperature.cpp
mairas: Can't - that's a method of `DallasTemperature`. And yes, I did try. :-)

src/sensors/sensor.h
mairas: Removed.

src/sensors/onewire_temperature.cpp
ba58smith: To avoid any future confusion, why not add a comment to the right of that line: 
// getDeviceCount() comes from another library, so can't be changed. We would rename it to getSensorCount() if we could.
Something like that.

src/sensors/onewire_temperature.cpp
```diff
@@ -91,22 +91,22 @@ bool DallasTemperatureSensors::get_next_address(OWDevAddr* addr) {
...
-    : NumericDevice(config_path), dts{dts} {
+    : NumericSensor(config_path), dts{dts} {
   load_configuration();
   if (address==null_ow_addr) {
-    // previously unconfigured device
+    // previously unconfigured sensor
     bool success = dts->get_next_address(&address);
     if (!success) {
-      debugE("FATAL: No more unregistered OneWire devices left");
+      debugE("FATAL: No more unregistered OneWire sensors left");
```
ba58smith: I'm not sure what this debug message means. Does it mean this?:

debugE("FATAL: All OneWire sensors have already been configured. Check the physical wiring of your sensors.");

And would it be good to include the config_path in the debug() message somehow, to make it clear which sensor in main.cpp is causing the problem?

src/sensors/onewire_temperature.cpp
```diff
@@ -91,22 +91,22 @@ bool DallasTemperatureSensors::get_next_address(OWDevAddr* addr) {
...
       found = false;
     } else {
-      debugD("Registered previously unconfigured OneWire device");
+      debugD("Registered previously unconfigured OneWire sensor");
```
ba58smith: I also don't understand what this debug message means. Does it mean that a OneWire sensor that had not previously been configured has now been configured, or has now been registered, or both? And what exactly does "registered" mean?
In any case, I think including the config_path in the debug() message would be very helpful.

src/sensors/onewire_temperature.cpp
```diff
@@ -91,22 +91,22 @@ bool DallasTemperatureSensors::get_next_address(OWDevAddr* addr) {
...
       dts->register_address(address);
     }
   } else {
     bool success = dts->register_address(address);
     if (!success) {
-      debugE("FATAL: Previously saved OneWire device could no longer be found");
+      debugE("FATAL: Previously saved OneWire sensor could no longer be found");
```
ba58smith: "Previously saved..." - Does that mean "Previously registered"? Or does it mean "Previously configured"? Or does it mean "Previously registered and configured"?
In any case, I think including the config_path in the debug() message would be very helpful.

src/system/valueconsumer.h
```diff
@@ -7,8 +7,8 @@
 template <typename T> class ValueProducer;
 
 /**
- *  A ValueConsumer is any piece of code (like a transformation) or device that
- *  accepts data for input. They can accept one or more input values 
+ *  A ValueConsumer is any piece of code (like a transformation) or sensor that
+ *  accepts data for input. They can accept one or more input values
```
ba58smith: Is a sensor ever a ValueConsumer? I thought the wiring was always sensor->ZeroOrMoreTransforms->OneOrMoreOutputs.

src/sensors/onewire_temperature.cpp
mairas: Sure thing.

src/sensors/onewire_temperature.cpp
mairas: Changed to: 

```
debugE("FATAL: Unable to allocate a OneWire sensor for %s. "
             "All sensors have already been configured. "
             "Check the physical wiring of your sensors.", addrstr);
```

src/sensors/onewire_temperature.cpp
mairas: Changed to: `debugD("Registered a new OneWire sensor");`

src/sensors/onewire_temperature.cpp
mairas: Changed to:

```
      debugE("FATAL: OneWire sensor %s at %s is missing. "
             "Check the physical wiring of your sensors.", config_path, addrstr);
```

src/system/valueconsumer.h
mairas: Probably not. I'll remove "... or sensor"

PR #44, opened 2019-06-15
Update RPM example after refactoring

ba58smith:
Make the example very clear and educational to someone looking at a SensESP project for the first time. Add a lot of explanation to the comments. Delete the comment that showed how to wire it up the old way (before the refactor).

### DIFF
--- a/examples/rpm_counter.cpp
+++ b/examples/rpm_counter.cpp
@@ -29,25 +29,31 @@ ReactESP app([]() {
   // If you have multiple sensors connected to your microcontoller (ESP), each one of them
   // will (probably) have its own SignalK path variable. For example, if you have two
   // propulsion engines, and you want the RPM of each of them to go to SignalK, you might
-  // have sk_path_leftEngine = "propulsion.left.revolutions"
-  // and  sk_path_rightEngine = "propulsion.right.revolutions"
+  // have sk_path_portEngine = "propulsion.port.revolutions"
+  // and sk_path_starboardEngine = "propulsion.starboard.revolutions"
   // In this example, there is only one propulsion engine, and its RPM is the only thing
   // being reported to SignalK.
-  const char* sk_path = "propulsion.left.revolutions";
+  const char* sk_path = "propulsion.main.revolutions";
 
 
   // The "Configuration path" is combined with "/config" to formulate a URL
   // used by the RESTful API for retrieving or setting configuration data.
   // It is ALSO used to specify a path to the SPIFFS file system
-  // where configuration data is saved on the MCU board.  It should
-  // ALWAYS start with a forward slash if specified.  If left blank,
+  // where configuration data is saved on the microcontroller.  It should
+  // ALWAYS start with a forward slash if specified. If left blank,
   // that indicates this sensor or transform does not have any
   // configuration to save.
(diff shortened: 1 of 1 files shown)


examples/rpm_counter.cpp
```diff
@@ -25,7 +25,14 @@ ReactESP app([]() {
...
-  // The "SignalK path" identifies this sensor to the SignalK network.
+  // The "SignalK path" identifies the output of the sensor to the SignalK network.
+  // If you have multiple sensors connected to your microcontoller (ESP), each one of them
+  // will (probably) have its own SignalK path variable. For example, if you have two
+  // propulsion engines, and you want the RPM of each of them to go to SignalK, you might
+  // have sk_path_leftEngine = "propulsion.left.revolutions"
+  // and  sk_path_rightEngine = "propulsion.right.revolutions"
+  // In this example, there is only one propulsion engine, and its RPM is the only thing
+  // being reported to SignalK.
   const char* sk_path = "propulsion.left.revolutions";
```
mairas: The description is very good, but what do you think about the path itself? Having "left" there for a single-engine boat irritates me more and more. Should we change it to "propulsion.0.revolutions" (or 1 if you prefer)?

ba58smith: I didn't know where "left" came from. I think I assumed it was in the SK list of already-defined paths. (It's not - I just looked.)  I don't like it, either - who calls it "left" or "right"? If there's only one, I might call it "mainEngine" (because I have a main engine, and the only other engine is the generator's engine). But if there's only one, then "main" doesn't seem quite right, either. But is a number any better? (Not everyone thinks you start counting things at 0, you know.) I don't know - don't really much care. Suggest something, and I'll use it.

ba58smith: Based on a brief "poll" on Slack, I'm going with "main" instead of "left. I'll also change left and right to port and starboard in my comment.

examples/rpm_counter.cpp
```diff
@@ -53,23 +69,15 @@ const float multiplier = 1.0 / 97.0;
...
 
-  // 2. Connect the producer directly to the consumer
-  auto* sensor = new DigitalInputCounter(D5, INPUT_PULLUP, RISING, read_delay);
 
-  sensor->connectTo(new Frequency(multiplier, "/sensors/engine_rpm/calibrate"))
-        ->connectTo(new SKOutputNumber(sk_path, "/sensors/engine_rpm/sk"));
-
-
-  // Start the SensESP application running
+  // Start the SensESP application running. It constantly monitors the interrupt pin, and every
```
joelkoz: Just so we are on the same page, its all the lines ABOVE that makes the code monitor the interrupt pin and send sigk data to the network.  The sensesp_app->enable() call simply sets everything in motion.  This comment sort of implies that its the call here that does that.  I guess that is true if you are describing the app, but all throughout here you are describing what each line of code does.  Wouldn't this be better to just say "set all of the above setup in motion" or something like that?

ba58smith: Yeah, I was trying to summarize what `sensesp_app->enable();` is all about - setting everything defined above in motion, which means monitoring the interrupt pin and sending the frequency to SK every read_dealy ms. I'm going to merge this, and revisit it to make it more clear.

PR #48, opened 2019-06-17
feature: Angle correction transform

mairas:
Implemented an angle correction transform (take an angle value and add a constant offset to it, wrapping the result value back into legal range).

Also fix GPSInput arguments and add optional reset_pin for forcing the GPS modules to reset.

### DIFF
--- a/src/transforms/angle_correction.cpp
+++ b/src/transforms/angle_correction.cpp
@@ -1,0 +1,55 @@
+#include "angle_correction.h"
+
+AngleCorrection::AngleCorrection(float offset, float min_angle, String config_path) :
+    NumericTransform(config_path),
+      offset{ offset }, min_angle{ min_angle } {
+  load_configuration();
+}
+
+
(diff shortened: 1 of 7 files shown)


src/transforms/angle_correction.cpp
```diff
@@ -0,0 +1,55 @@
...
+  root["min_angle"] = min_angle;
+  root["value"] = output;
+  return root;
+}
+
+static const char SCHEMA[] PROGMEM = R"###({
+    "type": "object",
+    "properties": {
+        "offset": { "title": "Constant offset (radians)", "type": "number" },
+        "min_angle": { "title": "Minimum angle value (radians, typically 0 or -3.15159265", "type": "number" },
```
joelkoz: Json Schema (and the Json Editor) supports a "description" property for detailed descriptions.  How about trying to limit "title" to just "Minimum angle value" and then put the rest of the helpful text in a "description" property

mairas: Good tip! Fixed.

Issue #27, opened 2019-06-11
docs: add comments to interpolate.h and temp_sender.cpp


ba58smith:
See all the comments for `examples/temperature_sender.cpp` in PR #24 for guidance.

ba58smith:
See PR#50 that addresses this issue.

PR #50, opened 2019-06-18
CurveInterpolator and temp_sensor example comments

ba58smith:
Addresses issue #27 
Slight modification to the comment describing the CurverInterpolator transform, and a lot of additional comments in the example that uses it, temperature_sensor.cpp.

### DIFF
--- a/examples/temperature_sender.cpp
+++ b/examples/temperature_sender.cpp
@@ -9,27 +9,28 @@
 #include "transforms/curveinterpolator.h"
 #include "signalk/signalk_output.h"
 
-/**
- * Illustrates a custom transform that takes a resistance value in ohms and returns the estimated
- * temperature in Kelvin. Sample data in this example were taken from a Westerbeke generator
- * temperature sender and gauge.
- * Note that you will never instantiate CurveInterpolator in your own SensESP
- * project - you will always need to create a descendant class of it, like
- * TemperatureInterpreter in this example. The constructor needs to do only two things:
- * call clearSamples(), then call addSample() at least twice. (This class can't function
- * without at least two samples.) You can call addSample() as many times as you like,
- * and the more samples you have, the more accurately this transform will emulate the
- * analog sensor. Note, however, that since all transforms inherit from class
- * Configurable, you can make its data configurable in the Config UI, and that means
- * not only that you can edit (at runtime) whatever samples you add in the constructor,
- * but also that you can add more samples at runtime.
- * 
- * One of your samples should be at the lowest value for the input,and one should be at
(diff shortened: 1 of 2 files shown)


examples/temperature_sender.cpp
```diff
@@ -60,34 +72,59 @@ ReactESP app([] () {
...
-  // Translating the number returned by AnalogInput into a temperature requires several transformations.
-  // Wire them up in sequence...
-  pAnalogInput->connectTo(new AnalogVoltage()) ->
-                connectTo(new VoltageDividerR2(R1, Vin, "/gen/temp/sender")) ->
-                connectTo(new TemperatureInterpreter("/gen/temp/curve")) ->
-                connectTo(new Linear(1.0, 0.0, "/gen/temp/calibrate")) ->
-                connectTo(new SKOutputNumber(sk_path, "/gen/temp/sk"));
+  // Translating the number returned by AnalogInput into a temperature, and sending it to SignalK,
+  // requires several transforms. Wire them up in sequence...
+  pAnalogInput->connectTo(new AnalogVoltage()) ->  // convert the value from the AnalogIn pin into a voltage
```
mairas: As a code style comment, I would prefer not having comments inside a statement because it impairs code readability. I realize it's not as illustrative in this case, but I'd still just describe the flow in a separate comment block above the wiring statement.

mairas:
For consistency, maybe use just one commenting style for the larger comment blocks? Probably /* */ without the per-line asterisks would be most readable?

And yes, I know I tend to violate this all the time, using line comments instead...

joelkoz:
> For consistency, maybe use just one commenting style for the larger comment blocks? Probably /* */ without the per-line asterisks would be most readable?

I personally prefer the /** */ stype with an asterisk per line for two reasons: 1) That is the JavaDoc style (which I know isn't a good reason, esp since this is C++), but more importantly 2) - The unique comment marker emphasizes that the comments are different than just only old comment as they are intended to be transformed into the online documentation.  As an added benefit, when you start a comment with "/**", the other "asterisks at the start of each line" are actually inserted automatically by VSCode.

Those would be my argument for using JavaDoc style comments.  You are the project founder, however, and can dictate what you feel is right.  I don't feel strongly enough about it that my feathers would get ruffled if you decide to go with a more traditional style.  I am pretty sure Doxygen can pick up either style.

examples/temperature_sender.cpp
ba58smith: Will do.

ba58smith: Done.

PR #52, opened 2019-06-18
feature: add Enable::getClassName, use it in EnableAll()

ba58smith:
Add more detailed info to the startup messages in the Serial Monitor, showing each sensor and transport that's being enabled. Output looks like this. I don't know what the leading integer is, nor the trailing "fE" and "E" in some cases. That's what comes back from `return typeid(*this).name();`
```␛[0m(I) (enable) WS client enabled
␛[0m(D) (enableAll) Enabling sensor or transform: 6Linear
␛[0m(D) (enableAll) Enabling sensor or transform: 22TemperatureInterpreter
␛[0m(D) (enableAll) Enabling sensor or transform: 16VoltageDividerR2
␛[0m(D) (enableAll) Enabling sensor or transform: 8SKOutputIfE
␛[0m(D) (enableAll) Enabling sensor or transform: 13AnalogVoltage
␛[0m(D) (enableAll) Enabling sensor or transform: 11AnalogInput
␛[0m(D) (enableAll) Enabling sensor or transform: 8SKOutputI6StringE
␛[0m(D) (enableAll) Enabling sensor or transform: 9IPAddrDev
␛[0m(D) (enableAll) Enabling sensor or transform: 8SKOutputIfE
␛[0m(D) (enableAll) Enabling sensor or transform: 6Uptime
␛[0m(D) (enableAll) Enabling sensor or transform: 8SKOutputIfE
␛[0m(D) (enableAll) Enabling sensor or transform: 7FreeMem
␛[0m(D) (enableAll) Enabling sensor or transform: 8SKOutputIfE
␛[0m(D) (enableAll) Enabling sensor or transform: 8SystemHz
␛[0m(I) (enable) All sensors and transforms enabled```

### DIFF
--- a/src/system/enable.cpp
+++ b/src/system/enable.cpp
@@ -1,4 +1,6 @@
 #include "enable.h"
+#include "sensesp_app.h"
+
 
 std::priority_queue<Enable*> Enable::enableList;
 
@@ -11,6 +13,7 @@ Enable::Enable(uint8_t priority) : priority{priority} {
 void Enable::enableAll() {
     while (!enableList.empty()) {
         auto& obj = *enableList.top();
+        debugD("Enabling sensor or transform: %s", obj.getClassName());
         obj.enable();
         enableList.pop();
     } // while
--- a/src/system/enable.cpp
+++ b/src/system/enable.cpp
@@ -10,10 +10,16 @@ Enable::Enable(uint8_t priority) : priority{priority} {
 
 
 void Enable::enableAll() {
+    debugD("Enabling all required sensors and transforms");
     while (!enableList.empty()) {
         auto& obj = *enableList.top();
+        #if __GXX_RTTI
         debugD("Enabling sensor or transform: %s", obj.getClassName());
+        #endif
         obj.enable();
         enableList.pop();
     } // while
+    #if ! __GXX_RTTI
+    debugD("To see each sensor and transform listed here, add 'build_unflags = -fno-rtti' to platformio.ini");
+    #endif
 }
--- a/src/system/enable.cpp
+++ b/src/system/enable.cpp
@@ -1,7 +1,6 @@
 #include "enable.h"
 #include "sensesp_app.h"
 
-
 std::priority_queue<Enable*> Enable::enableList;
 
 
--- a/src/system/enable.cpp
+++ b/src/system/enable.cpp
@@ -10,14 +10,7 @@ Enable::Enable(uint8_t priority) : priority{priority} {
 
 
 const char* Enable::getClassName() { 
-    static int counter = 1;
-    #if __GXX_RTTI
-    return typeid(*this).name();
-    #else
-    char s[4];
-    sprintf(s, "%d", counter++);
-    return s;
-    #endif
+   return this->className;
 }
 
 
@@ -29,7 +22,4 @@ void Enable::enableAll() {
         obj.enable();
         enableList.pop();
     } // while
-    #if ! __GXX_RTTI
-    debugD("To see the name of each sensor and transform, add 'build_unflags = -fno-rtti' to platformio.ini");
-    #endif
 }
--- a/src/system/enable.cpp
+++ b/src/system/enable.cpp
@@ -9,11 +9,6 @@ Enable::Enable(uint8_t priority) : priority{priority} {
 }
 
 
-const char* Enable::getClassName() { 
-   return this->className;
-}
-
-
 void Enable::enableAll() {
     debugD("Enabling all required sensors and transforms");
     while (!enableList.empty()) {
--- a/src/system/enable.cpp
+++ b/src/system/enable.cpp
@@ -9,17 +9,27 @@ Enable::Enable(uint8_t priority) : priority{priority} {
 }
 
 
+const char* Enable::getClassName() { 
+    static int counter = 1;
+    #if __GXX_RTTI
+    return typeid(*this).name();
+    #else
+    char s[4];
+    sprintf(s, "%d", counter++);
+    return s;
+    #endif
+}
+
+
 void Enable::enableAll() {
     debugD("Enabling all required sensors and transforms");
     while (!enableList.empty()) {
         auto& obj = *enableList.top();
-        #if __GXX_RTTI
         debugD("Enabling sensor or transform: %s", obj.getClassName());
-        #endif
         obj.enable();
         enableList.pop();
     } // while
     #if ! __GXX_RTTI
-    debugD("To see each sensor and transform listed here, add 'build_unflags = -fno-rtti' to platformio.ini");
+    debugD("To see the name of each sensor and transform, add 'build_unflags = -fno-rtti' to platformio.ini");
     #endif
 }
--- a/src/system/enable.h
+++ b/src/system/enable.h
@@ -25,7 +25,9 @@ class Enable {
         virtual void enable() {}
      
 
-        virtual const char* getClassName() { return typeid(*this).name(); }   
+        #if __GXX_RTTI
+        virtual const char* getClassName() { return typeid(*this).name(); }
+        #endif
 
 
         const int8_t getEnablePriority() { return priority; }
--- a/src/system/enable.h
+++ b/src/system/enable.h
@@ -27,7 +27,7 @@ class Enable {
         void setClassName(const char* newClassName) { className = newClassName; }
         
         
-        virtual const char* getClassName();
+        virtual const char* getClassName() { return className; };
 
 
         const int8_t getEnablePriority() { return priority; }
--- a/src/system/enable.h
+++ b/src/system/enable.h
@@ -4,7 +4,6 @@
 #include <queue>
 #include <stdint.h>
 #include <typeinfo>
-// BAS #include "sensesp_app.h"
 
 /**
  * Classes that implement "Enable" will have their enable() method
--- a/src/system/enable.h
+++ b/src/system/enable.h
@@ -3,7 +3,10 @@
 
 #include <queue>
 #include <stdint.h>
+
+#if __GXX_RTTI
 #include <typeinfo>
+#endif
 
 /**
  * Classes that implement "Enable" will have their enable() method
@@ -25,6 +28,9 @@ class Enable {
         virtual void enable() {}
 
 
+        void setClassName(const char* newClassName) { this->className = newClassName; }
+        
+        
         virtual const char* getClassName();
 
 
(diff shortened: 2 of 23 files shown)

### CODE REVIEW

**Comment by ba58smith** [comment]
I just realized that this change requires the addition of `build_unflags = -fno-rtti ` to platformio.ini. Is that a problem?

**Comment by ba58smith** [comment]
I will see if I can find a foolproof way to detect whether `no-rtti` is active or not, and if I can, I'll use that around the call to getClassName().

File: src/system/enable.h [hunk 1]
```diff
@@ -22,6 +23,9 @@ class Enable {
          * to add runtime initialization code to your class
          */
         virtual void enable() {}
+     
+
+        virtual const char* getClassName() { return typeid(*this).name(); }
```
joelkoz: Idea: If you can't get a compile time check for the no-rtti flag, maybe you could just use an #if DEBUG.  I'm pretty sure "DEBUG" (or something like it) is defined if debug code is desired. You are only using this in a debugD() statement, so if debug info is not being included, then there is no "need" for this method to return a useful value.

ba58smith: I think I may have found a way to check the status of rtti, but if not, your suggestion seems like it will work, too. Thx.

ba58smith: Actually, I don't think your suggestion will work, because in some situation, debugging will be enabled, and this method will attempt to compile, and it won't w/o `rtti` being enabled. So I have to find a foolproof way of checking to see if rtti is enabled, and only if it is can I define this method, and call it in the debugD() line in EnableAll().

ba58smith: I think I found it: `#ifdef _CPPRTTI` means "If rtti is on".

ba58smith: ... but, it doesn't work. I've tried `#ifdef _CPPRTTI` and `#if _CPPRTTI` and `#if _GXX_RTTI` - none of them actually detect when RTTI is enabled. But, I did determine that enabling RTTI has a relatively small memory penalty: ```DATA:    [======    ]  58.2% (used 47684 bytes from 81920 bytes)
PROGRAM: [========  ]  76.1% (used 795216 bytes from 1044464 bytes)```
vs. ```DATA:    [=====     ]  49.0% (used 40148 bytes from 81920 bytes)
PROGRAM: [=======   ]  74.4% (used 776968 bytes from 1044464 bytes)```
So an increase in memory usage of about 26kb, or 3%, to enable RTTI.

Is this extra info all that valuable? I like seeing it, and I don't mind a little memory penalty, but if you guys think this shouldn't be done unless I can find a way to positively detect if RTTI is enabled, I won't put up a struggle. It's all been a great exercise for me, at least.

(Although this page seems to indicate that `#if _GXX_RTTI` SHOULD work! https://gcc.gnu.org/onlinedocs/cpp/Common-Predefined-Macros.html)

ba58smith: I've posted a question on the PlatformIO Community Forum about this. I will wait for a reply, and then we can decide whether to merge this change or not.

mairas: 20 KB is quite a bit in my opinion. Could you check the impact on run-time memory use, too? Might as well check systemHz, although I wouldn't assume it to affect that so much.

ba58smith: Got it working. If RTTI is NOT enabled, the output looks like this:
```
␛[0m(D) (enableAll) Enabling all required sensors and transforms
␛[0m(D) (enableAll) To see each sensor and transform listed here, add 'build_unflags = -fno-rtti' to platformio.ini
␛[0m(I) (enable) All sensors and transforms enabled
```

With RTTI enabled, it looks like this:
```
␛[0m(D) (enableAll) Enabling all required sensors and transforms
␛[0m(D) (enableAll) Enabling sensor or transform: 6Linear
␛[0m(D) (enableAll) Enabling sensor or transform: 22TemperatureInterpreter
␛[0m(D) (enableAll) Enabling sensor or transform: 16VoltageDividerR2
␛[0m(D) (enableAll) Enabling sensor or transform: 8SKOutputIfE
␛[0m(D) (enableAll) Enabling sensor or transform: 13AnalogVoltage
␛[0m(D) (enableAll) Enabling sensor or transform: 11AnalogInput
␛[0m(D) (enableAll) Enabling sensor or transform: 8SKOutputI6StringE
␛[0m(D) (enableAll) Enabling sensor or transform: 9IPAddrDev
␛[0m(D) (enableAll) Enabling sensor or transform: 8SKOutputIfE
␛[0m(D) (enableAll) Enabling sensor or transform: 6Uptime
␛[0m(D) (enableAll) Enabling sensor or transform: 8SKOutputIfE
␛[0m(D) (enableAll) Enabling sensor or transform: 7FreeMem
␛[0m(D) (enableAll) Enabling sensor or transform: 8SKOutputIfE
␛[0m(D) (enableAll) Enabling sensor or transform: 8SystemHz
␛[0m(I) (enable) All sensors and transforms enabled
```

ba58smith: @mairas - I missed your question about freemem on the device. Using the analog_input.cpp example:
- without RTTI enabled: 24,560
- with RTTI enabled: 17,168
But, since it now works with our without RTTI enabled, it can be left up to whoever's building the project.

src/system/enable.h
```diff
@@ -22,6 +23,11 @@ class Enable {
          * to add runtime initialization code to your class
          */
         virtual void enable() {}
+     
+
+        #if __GXX_RTTI
+        virtual const char* getClassName() { return typeid(*this).name(); }
```
joelkoz: One alternative: Keep enable calling this method regardless.  Put the #if __GXX_RTTI INSIDE the method, and have it return a class name if its enabled, or return a static number that you increment.  That way, at minimum, you'd see "enabling transform #1, enabling transform#2, etc."  If it locks up or crashes, you'd have an idea of how many transforms in to it it got before puking.  I am ok with leaving this enhancement for another day

ba58smith: @joelkoz, good idea - I did it. Posted it to you in Slack to look at before I commit it.

src/system/enable.h
ba58smith: Also missed the part about systemHz!
- without RTTI enabled: 99,000 +/- a few hundred.
- with RTTI enabled: 98,000 +/- a few hundred.
But again, it now works with or w/o RTTI enabled.

src/system/enable.cpp
```diff
@@ -8,10 +9,27 @@ Enable::Enable(uint8_t priority) : priority{priority} {
...
 void Enable::enableAll() {
+    debugD("Enabling all required sensors and transforms");
     while (!enableList.empty()) {
         auto& obj = *enableList.top();
+        debugD("Enabling sensor or transform: %s", obj.getClassName());
         obj.enable();
         enableList.pop();
     } // while
+    #if ! __GXX_RTTI
+    debugD("To see the name of each sensor and transform, add 'build_unflags = -fno-rtti' to platformio.ini");
```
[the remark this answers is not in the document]
ba58smith: Will do.

I was going to use a getClassName() member in each inheriting Class, but @joelkoz suggested doing it this way, for efficiency of programming. I'm guessing he didn't realize RTTI was such a resource hog.
If you'd rather, I'll toss all this out an implement it the other way, @mairas. I don't mind doing the work - this is all very educational for me. It would also solve the issue of those extra characters that get added to the actual Class name - the leading int and the sometimes trailing alpha characters. Let me know - this is your project, and it should be done the way you want.

src/system/enable.h
```diff
@@ -3,6 +3,7 @@
 
 #include <queue>
 #include <stdint.h>
+#include <typeinfo>
```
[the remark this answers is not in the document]
ba58smith: Will do.

src/system/enable.cpp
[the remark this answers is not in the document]
ba58smith: @mairas - I'm going to do it (get rid of the RTTI approach). Commit should be ready soon.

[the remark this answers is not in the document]
ba58smith: Well, easier said than done. The Sensors are easy, as each one of them has overridden `enable()`, so I added the debug() line there. But most of the Transforms don't override `enable()` - they use the `enable` in the Base Class "Enable". 
(Before I go on... I don't want to waste your time, @mairas and @joelkoz , if you don't think this is worth any more effort. PLEASE let me know if you feel like this isn't worth it, and I'll just drop it all, or maybe leave in the non_RTTI version that simply counts the sensors and transports as they're being enabled, and leave it at that. But IF feel like having this in the initialization output is worth it, I will proceed, either as follows, or however you guys suggest.)
Is there a way to add a data member to the "Enable" class called "className", and easily setting that member's value in the Constructor of each Class that inherits from "Enable"? Or something like that? Is there a way to declare a data member that MUST be overridden by every class that inherits from it? Sorry - I know this is not SensESP stuff, but rather, just basic C++ stuff.
Awaiting your guidance....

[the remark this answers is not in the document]
ba58smith: "Awaiting" is a relative term, I guess. After tossing and turning most of the night (truly), I got up this morning determined to find a way to do this easily. After much research on Constructors and Initialization Lists and Templates, etc., etc., ad nauseum, an idea hit me. What if Class Enable had a private attribute `String classDisplayName = NULL`, and a public `getClassDisplayName()` and `setClassDisplayName()`. If you want your Sensor or Transport to display its name during initialization, you use `setClassDisplayName()` after you instantiate any Sensor or Transport. `EnableAll()` is then smart enough to output the classDisplayName if it's been defined, or to output something generic if it's still NULL.
Awaiting your guidance...

joelkoz:
Sorry - forgot to hit “send” yesterday!


You don’t need to define enable() in each class. Keep the implementation like you have it: have the base class output the debug msg by calling getClassName. Then define getClassName as a “virtual” function. make the base definition of the function simply return “unknown”. Then, on each sensor or transform, re-define getClassName like:

virtual const char* getClassName() override { return “myName”; }

-jk
(sent from my iPhone)

joelkoz:
I actually like this solution, as it allows you to define an instance specific name for debugging purposes. 

You could combine the two solutions (my “virtual method” suggestion) by renaming your private variable “debugName”  Enable could be:

if (getDebugName() != null)
   display debug name
else
   display getClassName()

-jk
(sent from my iPhone)

src/system/enable.cpp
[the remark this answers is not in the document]
mairas: Hi, just got back on the hard. Your proposed approach sounds sane, and the obvious way to do it. :-) Some minor details, not affecting the concept: assuming the intention would be to set classDisplayName in class constructors, you'd want to make it protected, not private. If it's private, inherited classes won't see it. 

As for the concept itself, I think it's fine. The class name could be used to output debugging messages in e.g. SKOutput or some other often called base class, so there's some general use for it. :-)

ba58smith:
I ended up doing it differently, but I think it's simple, and it works. You can see the guts of the change in enable.h and enable.cpp. All the other changes are simply adding `this->className = <the actual name of the Class>` to the Constructor of each Class, but NOT as a parameter. (I didn't see a use case for someone changing the className from the actual Class name, so I just hard-coded it into each Constructor.) I did, however, create Enable::setClassName, and I tested it, and it works, in case someone wants to change the className in main.cpp.

If you'd rather I use `this->setClassName(<the actual name of the Class)`, instead of `this->className = <the actual name of the Class>, I can do that pretty easily. Just tell me why, so I'll not make the same mistake again.

ba58smith:
@joelkoz , I didn't implement your suggestion to do this:
```
if (getDebugName() != null)
   display debug name
else
   display getClassName()
```
because, if I understand your intent, I made that possible by using `this->setClassName` in your main.cpp, so you can temporarily change the ClassName for a particular instance of a Sensor or Transport for debugging purposes.

PR #53, opened 2019-06-23
fix: Hard-code network settings after one successful connection

### LINKED ISSUE #16: fix: doesn't reconnect to wifi if wifi goes off and comes back on - not saving the wifi config?
I thought this was happening, so I just restarted my wifi to test it, and it is.

Turned wifi on, turned D1 Mini on, SensESP connected and started sending data. Was sending data for a few hours. I turned off the wifi, and SensESP immediately detected that, and restarted. When it came up, it didn't find my wifi (it was still off), so it went into AP mode as it's supposed to. But then, after I turned the wifi on, and SensESP was seeing the wifi, it still didn't connect to it. I had to go to Unconfigured Sensor and re-select my wifi and re-enter the password, and then it was OK. Here's the serial output of the whole test, starting with me turning the wifi off.
```
␛[0m(E) (check_connection) Wifi disconnected: restarting...
␛[0m
 ets Jan  8 2013,rst cause:2, boot mode:(3,6)

load 0x4010f000, len 1384, room 16
tail 8
chksum 0x2d
csum 0x2d
v8b899c12
~ld
(I) (enable) Connecting SignalK source sensors.WemosD1miniTest.systemhz
␛[0m(I) (enable) Connecting SignalK source sensors.WemosD1miniTest.freemem
␛[0m(I) (enable) Connecting SignalK source sensors.WemosD1miniTest.uptime
␛[0m(I) (enable) Connecting SignalK source sensors.WemosD1miniTest.ipaddr
␛[0m(I) (enable) Connecting SignalK source propulsion.left.revolutions
␛[0m(I) (enable) Enabling subsystems
␛[0m(I) (enable) Subsystem: networking->setup()
␛[0m*WM: Adding parameter
*WM: hostname
*WM:
*WM: AutoConnect Try No.:
*WM: 0
*WM: Connecting as wifi client...
*WM: Using last saved values, should be faster
*WM: Connection result:
*WM: 1
*WM: SET AP STA
*WM:
*WM: Configuring access point...
*WM: Unconfigured Sensor
*WM: AP IP address:
*WM: 192.168.4.1
*WM: HTTP server started
*WM: About to scan()
*WM: About to scan()
*WM: Scan done 
// it does this about 20 times.
// I turned my router back on about here....
*WM: About to scan()
*WM: About to scan()
*WM: Scan done
(E) (setup) Failed to connect to wifi and config timed out. Restarting...
␛[0m
 ets Jan  8 2013,rst cause:2, boot mode:(3,6)

load 0x4010f000, len 1384, room 16
tail 8
chksum 0x2d
csum 0x2d
v8b899c12
~ld
(I) (enable) Connecting SignalK source sensors.WemosD1miniTest.systemhz
␛[0m(I) (enable) Connecting SignalK source sensors.WemosD1miniTest.freemem
␛[0m(I) (enable) Connecting SignalK source sensors.WemosD1miniTest.uptime
␛[0m(I) (enable) Connecting SignalK source sensors.WemosD1miniTest.ipaddr
␛[0m(I) (enable) Connecting SignalK source propulsion.left.revolutions
␛[0m(I) (enable) Enabling subsystems
␛[0m(I) (enable) Subsystem: networking->setup()
␛[0m*WM: Adding parameter
*WM: hostname
*WM:
*WM: AutoConnect Try No.:
*WM: 0
*WM: Connecting as wifi client...
*WM: Try to connect with saved credentials
*WM: Connection result:
*WM: 0
*WM: SET AP STA
*WM:
*WM: Configuring access point...
*WM: Unconfigured Sensor
*WM: AP IP address:
*WM: 192.168.4.1
*WM: HTTP server started
*WM: About to scan()
*WM: About to scan()
*WM: Scan done
*WM: DUP AP: theturtle_IoT
*WM: DUP AP: theturtle
*WM: DUP AP: fos_guest
*WM: About to scan()
*WM: About to scan()
*WM: Scan done
*WM: DUP AP: fos_guest
*WM: DUP AP: theturtle
*WM: DUP AP: theturtle_IoT
// So it's obviously seeing the network, but won't connect to it.
```
It's now more than an hour after I pasted the above in, and I'm just finishing this. It's still doing what it was doing above: scanning, finding the networks, but not connecting to them. So I just restarted the Wemos, and it's doing the same thing. I have to go into Unconfigured Sensor and tell it my wifi info again, and then it's OK.

mairas:
With this one, I'd like you to think hard whether you like my approach. Basically, I found that the bug @ba58smith reported was almost a show-stopper for me. I had some sensors behind the same power switch as the SK server itself. When I turned the power on, the Raspi couldn't get the network up fast enough and the ESPs would fall back to unconfigured mode. Every single time.

What I do is I save SSIDs and network passwords after a successful network login via Wifimanager. Then, if the values have been set, they're used unconditionally for reconnecting to the network. Basically, once Wifimanager is successfully used, the next time the device will wait forever until the same network becomes available again. The downside is there's no way to access Wifimanager again but by resetting the device. This worked for my use case but would break if, say, you wanted to use a different network for development.

Fixes #16

### DIFF
--- a/src/sensors/onewire_temperature.cpp
+++ b/src/sensors/onewire_temperature.cpp
@@ -135,7 +135,7 @@ void OneWireTemperature::update() {
 }
 
 void OneWireTemperature::read_value() {
-  output = dts->sensors->getTempC(address.data());
+  output = dts->sensors->getTempC(address.data()) + 273.15;
   this->notify();
 }
 
--- a/src/sensors/onewire_temperature.cpp
+++ b/src/sensors/onewire_temperature.cpp
@@ -135,6 +135,7 @@ void OneWireTemperature::update() {
 }
 
 void OneWireTemperature::read_value() {
+  // getTempC returns degrees Celsius but SignalK expects Kelvins
   output = dts->sensors->getTempC(address.data()) + 273.15;
   this->notify();
 }
--- a/src/net/networking.cpp
+++ b/src/net/networking.cpp
@@ -37,7 +37,8 @@ void Networking::check_connection() {
 void Networking::setup(std::function<void(bool)> connection_cb) {
   if (ap_ssid != "" && ap_password != "") {
     setup_saved_ssid(connection_cb);
-  } else {
+  }
+  if (WiFi.status() != WL_CONNECTED) {
     setup_wifi_manager(connection_cb);
   }
   app.onRepeat(1000, std::bind(&Networking::check_connection, this));
@@ -46,16 +47,20 @@ void Networking::setup(std::function<void(bool)> connection_cb) {
 void Networking::setup_saved_ssid(std::function<void(bool)> connection_cb) {
   WiFi.begin(ap_ssid, ap_password);
 
+  uint32_t timer_start = millis();
+
   rdebugI("Connecting");
-  while (WiFi.status() != WL_CONNECTED)
-  {
+  while (WiFi.status() != WL_CONNECTED &&
+         (millis() - timer_start) < 3 * 60 * 1000) {
     delay(500);
     rdebugI(".");
   }
   debugI("\n");
 
-  debugI("Connected to wifi, SSID: %s", WiFi.SSID().c_str());
-  connection_cb(true);
+  if (WiFi.status() == WL_CONNECTED) {
+    debugI("Connected to wifi, SSID: %s", WiFi.SSID().c_str());
+    connection_cb(true);
+  }
 }
 
 void Networking::setup_wifi_manager(std::function<void(bool)> connection_cb) {
--- a/src/net/networking.cpp
+++ b/src/net/networking.cpp
@@ -35,6 +35,30 @@ void Networking::check_connection() {
 }
 
 void Networking::setup(std::function<void(bool)> connection_cb) {
+  if (ap_ssid != "" && ap_password != "") {
+    setup_saved_ssid(connection_cb);
+  } else {
+    setup_wifi_manager(connection_cb);
+  }
+  app.onRepeat(1000, std::bind(&Networking::check_connection, this));
+}
+
+void Networking::setup_saved_ssid(std::function<void(bool)> connection_cb) {
+  WiFi.begin(ap_ssid, ap_password);
+
+  rdebugI("Connecting");
+  while (WiFi.status() != WL_CONNECTED)
+  {
+    delay(500);
+    rdebugI(".");
+  }
+  debugI("\n");
+
+  debugI("Connected to wifi, SSID: %s", WiFi.SSID().c_str());
+  connection_cb(true);
+}
+
+void Networking::setup_wifi_manager(std::function<void(bool)> connection_cb) {
   should_save_config = false;
 
   //set config save notify callback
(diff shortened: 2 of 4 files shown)


mairas:
One of the issues here was that the network configuration was saved in a rather implicit manner, and the ESP WiFi library isn't really that well documented. I still have no idea where the settings are actually saved and whether it would be possible to overwrite the current settings in that location.

ba58smith:
If I understand correctly, it would work like this:
1. Connect successfully to network once, and SSID/pw are saved, and used forever, unless...
2. ... you bring up the Config UI and do a "Reset Device", which resets to factory settings (which wipes out not only the wifi config, but any other config you've done for sensors and transforms).

That's not ideal, but for ME, it would be better than having to reconfig the wifi whenever my Wemos starts or restarts and the wifi network isn't ready.

Would it be possible to have something in the Config UI that erases ONLY the wifi config, so it does that, then restarts, and sees no wifi config, so it bring up Unconfigured SensESP Device? That would seem to be a good compromise. But, if that's not practical, given the lack of WiFiManager documentation, I can certainly live with your change. I don't plan on changing my wifi for any reason.

mairas:
> If I understand correctly, it would work like this:
> 
> 1. Connect successfully to network once, and SSID/pw are saved, and used forever, unless...
> 2. ... you bring up the Config UI and do a "Reset Device", which resets to factory settings (which wipes out not only the wifi config, but any other config you've done for sensors and transforms).

Yep, precisely.

> Would it be possible to have something in the Config UI that erases ONLY the wifi config, so it does that, then restarts, and sees no wifi config, so it bring up Unconfigured SensESP Device? That would seem to be a good compromise. But, if that's not practical, given the lack of WiFiManager documentation, I can certainly live with your change. I don't plan on changing my wifi for any reason.

That would be quite easy to do. But if you'd still need to remember to kick SensESP before modifying your network.

Thinking of it, it would be quite easy to extend the logic of this PR to attempt to connect to a predefined SSID for 3 minutes, then open WiFiManager (again with a 3 minute timeout), then reboot again etc. The devices would eventually always become online but still be available for reconfiguration every once in a while. How would that sound?

ba58smith:
I like the "try defined SSID for 3 minutes, then go to Unconfigured for 3 minutes, then repeat" approach.

mairas:
Implemented the best effort scheme: first try for 180s to connect to the saved network, then revert to WiFiManager with a 180s timeout, then rinse and repeat. Tested the different scenarios and would seem to work as intended.

ba58smith:
Did you mean to include the changes to onewire_temperature.cpp and wiring_helpers.cpp in this PR?

I don't see anything wrong in the networking changes, but I don't think I would recognize anything if it were wrong. After you merge it, I'll test it ASAP.

src/sensors/onewire_temperature.cpp
```diff
@@ -135,7 +135,7 @@ void OneWireTemperature::update() {
 }
 
 void OneWireTemperature::read_value() {
-  output = dts->sensors->getTempC(address.data());
+  output = dts->sensors->getTempC(address.data()) + 273.15;
```
ba58smith: I would add a comment to this about why you add 273.15. (Not everyone will know that the OneWire sensors output in Centigrage, but SK wants input in Kelvin.)

mairas: Done.

mairas:
> Did you mean to include the changes to onewire_temperature.cpp and wiring_helpers.cpp in this PR?

Sorry, yes, they were minor fixes which I forgot to mention about.

> I don't see anything wrong in the networking changes, but I don't think I would recognize anything if it were wrong. After you merge it, I'll test it ASAP.

I think the logic is a bit clunky at the moment - not elegant but works. Since @joelkoz is distracted, I'll use the opportunity and merge. :-)

PR #56, opened 2019-07-20
NEW: NoiseFilter and ChangeFiler

joelkoz:
This PR adds two new filters:

1. NoiseFilter is a passthrough filter that only outputs values if the delta between the new value and the previous emitted value is less than a specified threshold.

2. ChangeFilter is a passthrough filter that only outputs values if the delta between the new value and the previous emitted value is greater than a specified threshold.

It also includes a single fix for MovingAverage - the previous implementation started with the initial value being averaged with zero, which skewed the first several results. Now, the initial value is passed through, the second value is averaged with the first value, etc.


joelkoz:
I am still open to comments/changes if anyone wants to review.

ba58smith:
Looks good to me.

Issue #28, opened 2019-06-12
BUG: mDNS hostname is not advertised properly


mairas:
The mDNS responder no longer properly advertises the hostname on the local network. As a result, accessing the ESP device using the address, say, sensesp.local no longer works.

ba58smith:
@joelkoz fixed this with PR #58.

PR #58, opened 2019-07-21
Fix mDns discovery

joelkoz:
Fixes issue #28

### DIFF
--- a/src/net/networking.cpp
+++ b/src/net/networking.cpp
@@ -19,7 +19,7 @@ void save_config_callback() {
 
 Networking::Networking(String config_path)
     : Configurable{config_path} {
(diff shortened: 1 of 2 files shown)


joelkoz:
Indeed.  More so than the error - now we can reach our devices without knowing the IP address!

ba58smith:
I hadn't thought about that, which is, I guess, one of the main reasons to use mDNS. Do you reach the device with whatever hostname you give it when you complete the "Unconfigured SensESP Device" setup? So, in a browser, go to http://myDeviceName ?

joelkoz:
Almost - its the device name plus ".local":

http://myDeviceName <http://mydevicename/>.local

PR #59, opened 2019-07-24
Add comments for documentation

ba58smith:
Reviewing @joelkoz's recent change to MovingAverage, I realized I didn't understand how it worked. The PR adds a description for the class documentation, and explains how the transform works.

### DIFF
--- a/src/transforms/moving_average.h
+++ b/src/transforms/moving_average.h
@@ -12,28 +12,22 @@
  * MovingAverage outputs the average of the most recent n values. It also incorporates
  * a "scale" factor, in case you want to increase or decrease your final output by a
  * fixed percentage.
- * 
- * @param n is the number of samples to average
- * 
- * @param k is the "scale" - defaults to 1.0, so has no effect unless you use other than 1.0
- * 
  */
 
 // y = k * 1/n * \sum_k=1^n(x_k)
 class MovingAverage : public NumericTransform {
 
  public:
-  MovingAverage(int n, float k=1., String config_path="");
-
   /**
-   * set_input is used to input a new value to this transform. After the
-   * transformation, notify() is called to output the new value.
-   */
+   * @param n is the number of most recent values you want to average for your output
+   * 
+   * @param the moving average will be multiplied by k before it is output - make it 
+   * something other than 1. if you need to scale your output up or down by a fixed
+   * percentage
+   * 
+   * */
+  MovingAverage(int n, float k=1., String config_path="");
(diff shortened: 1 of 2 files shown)


src/transforms/moving_average.h
```diff
@@ -5,12 +5,35 @@
...
 
+/**
+ * MovingAverage is used to smooth the output of a value (signal) that has frequent
+ * variations. For example, the output of a temperature sensor may vary from 180 to 185
+ * several times over a short period, but you just want to see the average of that.
+ * MovingAverage outputs the average of the most recent n values. It also incorporates
+ * a "scale" factor, in case you want to increase or decrease your final output by a
+ * fixed percentage.
+ * 
+ * @param n is the number of samples to average
```
joelkoz: I think @param tag is only valid for comments that are on functions or the constructor.  Move these comments to just above the constructor declaration.

src/transforms/moving_average.h
```diff
@@ -5,12 +5,35 @@
...
   MovingAverage(int n, float k=1., String config_path="");
+
+  /**
+   * set_input is used to input a new value to this transform. After the
+   * transformation, notify() is called to output the new value.
+   */
   virtual void set_input(float input, uint8_t inputChannel = 0) override;
+
+  /** 
+   * For reading and writing the configuration of this transform
```
joelkoz: The way Doxygen works, this comment, which applies to the three methods below, will actually only apply to the first method declaration (get_configuration()). I would put NO comments on ANY of these methods, as they are all "virtual" and "override."  Instead, if you document each method in the base class where they are inherited from (if they are not already documented), then these overrides will "inherit" the documentation.  So, in general, we should not have any Doxygen docs for any "virtual ... override" methods except for the original declarations in the base class.  Conversely, we should try our best to ALWAYS document "virtual" methods declared in a base class so as to allow descendant classes to inherit those docs.

src/transforms/moving_average.h
ba58smith: Done.

src/transforms/moving_average.h
ba58smith: The base class is Configurable, and configurable.h has a comment for each of these methods, so I'll remove them from here. Thanks for the clarification. (I was using curveinterpolator.h as a model for doing this - it has "//" in the places where I put the "/**" comments.)

mairas-cujo:
Sorry, forgot about this!

PR #60, opened 2019-08-02
Refactor change noise filter

joelkoz:
"NoiseFilter" only handled changes in value outside a maximum range. There are actually a variety of ways to handle noise, so I am removing this filter, and folding its functionality into ChangeFilter.  Now, ChangeFilter will filter out values when the delta between the value and the previously output value is either too small OR NOW TOO LARGE.


src/transforms/change_filter.cpp
```diff
@@ -12,26 +11,45 @@ static float absf(float val) {
...
+          maxDelta{maxDelta},
+          maxSkips{maxSkips},
+          NumericTransform(config_path) {
+
+  load_configuration();
+  skips = maxSkips+1;
+}
+
+
 void ChangeFilter::set_input(float newValue, uint8_t inputChannel) {
```
ba58smith: Consider this scenario: newValue is "out of range", but because we've reached maxSkips, newValue becomes output. Now, the next maxSkips number of newValues are rejected (if they're all "in range"), or a bunch of "out of range" values are output if they're all near the first one that was allowed through.
May not be likely, and I guess it will eventually "settle down"? And I guess as long as maxSkips is set high enough, it's likely that an "in range" value will pass through and restart the whole process, most of the time. How big is maxSkips in your implementation?

mairas: I think this is a valid concern; by exhausting skips, the filter might output a random peak value well outside the normal region.

joelkoz: Yes, I've not only considered the scenario, I've encountered it.  However, what is the "solution?"  Do not support the feature at all? I figured a user would do one of two things:

1. Set the max skips to a "reasonable" number like 10 or 20.  The normal AnalogInput sends its samples 10 times a second, to using "20" for max skips means it will output numbers that remain in that range for over two seconds.  If there happens to be more than two seconds of noise on the line, yes, the value could get "stuck" there, but only for another two seconds.

2. Set the max skips to a huge number, like 9999.  That in effect disables the feature, and no number will ever be output unless the numbers are "out of range" for ten thousand consecutive samples.  Note that in this scenario, if a number is out of range for five thousand samples, and the 5,001 sample is in range, the number is output and the skip count is reset.

Is there a third scenario?

mairas:
I wonder if the existing implementation mixes two separate concerns: on the other hand, it removes random outliers, but it also requires a minimum delta to output anything. These are somewhat separate issues, in my opinion.

SignalK supports metadata including signal validity durations for different paths. If I remember correctly, the server and the clients should treat the path values as not existing if no values have been received within the validity window. Once the metadata is set for SensESP output paths, I think it would be important to set the minDelta and skip values low enough to prevent the output from timing out.

Actually, is there a concrete need for not just sending the same values out each and every time? Even if there's no change, it would probably be beneficial to receive a timestamp of the latest measurement.

src/transforms/change_filter.cpp
joelkoz: One enhancement: I could create a separate minValSkipCount and maxValSkip count, which would allow you to, say "send an unchanged value every 10, just to give make sure the consumers know the producer is even still alive", but "never send max value changes unless its over that value for 9999"

joelkoz:
> I wonder if the existing implementation mixes two separate concerns: on the other hand, it removes random outliers, but it also requires a minimum delta to output anything. These are somewhat separate issues, in my opinion.

Well, the original implementation was "ChangeFilter", which really was meant to filter out large streams of unchanged data, but at least allow a value thru every once in a while so it doesn't look like it times out, and separately "NoiseFilter" which was to filter out a consecutive value that has changed dramatically.  My thought on this PR was a) there is more than one way to filter noise, and b) the technique used by NoiseFilter was really the same as ChangeFilter, just using "max" instead of "min"

Maybe I stick with the separate implementations, but rename them MinChangeFilter and MaxChangeFilter?  Or, I could just drop this PR and stick with what is currently in the repo?

joelkoz:
> Actually, is there a concrete need for not just sending the same values out each and every time? Even if there's no change, it would probably be beneficial to receive a timestamp of the latest measurement.

And to answer this question - since AnalogInput samples 10 times per second, a properly functioning temperature or oil pressure gauge would hold the same value steady for hours of operation.  Do we really need 600 values sent every minute?  I would think once per second is more than enough if the value is identical.  That was the original thinking behind ChangeFilter in the first place

ba58smith:
I agree that we certainly don't need to sample typtical gauge data 10 times per second. Once per second seems plenty. Of course, there will be exceptions, so I suppose AnalogInput needs a "sampleInterval", either in the constructor, or as with a setSampleInterval() method. 

And I agree that there's nothing wrong with outputting the last valid "output" value every so often, so that the server sees a heartbeat. Every 5 seconds? But only to a point - after a certain period, if no valid newValue is received, it should stop sending anything, or send something that will be displayed as an obvious error (the old "999" or "000" trick). (I know - bad program design - but still effective!)

Ultimately, I think this is just masking a hardware problem with software. That sending unit shouldn't be putting out such variable data. (Or maybe it's the long wires.) I'm going to do some Googling and see if I can find anything helpful.

joelkoz:
>  so I suppose AnalogInput needs a "sampleInterval", either in the constructor, or as with a setSampleInterval() method.

Actually, I kind of like the way it is now.  AnalogInput data is knows to be unstable regardless, so I've read its best to take several samples to determine a single "read".  Here is what I currently have:

```
pAnalogInput->connectTo(new Median(5, "/gauge/smooth/samples")) ->
                connectTo(new ChangeFilter(0.0, 100.0, 15, "/gauge/smooth/change")) ->
                connectTo(new MovingAverage(5, 1.0, "/gauge/smooth/avg")) ->
                connectTo(pVoltage = new AnalogVoltage()) ->
                connectTo(new VoltageDividerR2(R1, Vin)) ->
                connectTo(new TemperatureInterpreter("/gauge/temp/curve")) ->
                connectTo(pTempInKelvin = new Linear(1.0, 0.0, "/gauge/temp/calibrate")) ->
                connectTo(new SKOutputNumber(sk_path, "/gauge/temp/sk")) ->
                connectTo(pGauge);
```
So, I find the median value of five reads.  That is passing a value every 1/2 second to the ChangeFilter. The change filter is throwing out any jump by more than 100 (remember, at this point, we are still getting values that are 0 thru 1023). If I get 15 reads out of range (about 7 1/2 seconds worth), it lets it thru.  Values that make it thru are then smoothed using a moving average. ONLY THEN do I take the value, convert it to a voltage value, and then figure out the temperature.

All of those initial transforms can be configured, so I've been playing around with the numbers.  I'd have to look at my gauges current config to see what values I'm currently using.  Its "pretty good", but I still get the occasional jumps.  

I agree that I PROBABLY need to clean up the signal in hardware, but I first need to get a little more scientific about determining what the real problem is.

mairas:
Hi, it's finally the weekend with no plans and a bit of time to devote for nerdy stuff. It's surprising how being back to work tends to interfere with your hobbies. Not sure if you guys remember that word any longer. "Work."

First, I'll get over with the most pressing issue: the PR is fine for all intents and purposes and can be merged as-is, as far as I'm concerned. I'll continue discussing the various details above but none of them are essential,

mairas:
> I agree that we certainly don't need to sample typtical gauge data 10 times per second. Once per second seems plenty. Of course, there will be exceptions, so I suppose AnalogInput needs a "sampleInterval", either in the constructor, or as with a setSampleInterval() method.

Yes, setting the sampling interval (or a sample rate, which is its inverse) would make a lot of sense. If you know your signal is clean, it's better to acquire it just as often as you need. I'd definitely go with a constructor argument.

mairas:
> ```
> pAnalogInput->connectTo(new Median(5, "/gauge/smooth/samples")) ->
>                 connectTo(new ChangeFilter(0.0, 100.0, 15, "/gauge/smooth/change")) ->
>                 connectTo(new MovingAverage(5, 1.0, "/gauge/smooth/avg")) ->
>                 connectTo(pVoltage = new AnalogVoltage()) ->
>                 connectTo(new VoltageDividerR2(R1, Vin)) ->
>                 connectTo(new TemperatureInterpreter("/gauge/temp/curve")) ->
>                 connectTo(pTempInKelvin = new Linear(1.0, 0.0, "/gauge/temp/calibrate")) ->
>                 connectTo(new SKOutputNumber(sk_path, "/gauge/temp/sk")) ->
>                 connectTo(pGauge);
> ```

This, btw, is a beautiful processing chain. :-)

The signal processing geek in me, though, would try to keep the sample rate constant (or rather, known) throughout the processing chain. As it is, after the second step you no longer know how many raw samples are used to create the output. So, assuming you already have filtered the outliers with the median filter, I'd instead first use a `MovingAverage` with a long window of, say, 50 samples to add low-pass filtering and then implement a new `DownSample` transform to only take every 50th sample to increase the sample interval from 0.1s to 5s. Then you could tell throughout processing how many samples affect the output.

> So, I find the median value of five reads. That is passing a value every 1/2 second to the ChangeFilter. The change filter is throwing out any jump by more than 100 (remember, at this point, we are still getting values that are 0 thru 1023). If I get 15 reads out of range (about 7 1/2 seconds worth), it lets it thru. Values that make it thru are then smoothed using a moving average. ONLY THEN do I take the value, convert it to a voltage value, and then figure out the temperature.

Could you show a plot or a screenshot of the raw AnalogInput output signal? I was assuming that the spikes were only one sample wide but obviously their duration must be longer if they come through the median filter. In any case, median filter is very effective in removing such issues, so just increasing its length could help a lot.

joelkoz:
> Could you show a plot or a screenshot of the raw AnalogInput output signal? I was assuming that the spikes were only one sample wide but obviously their duration must be longer if they come through the median filter. In any case, median filter is very effective in removing such issues, so just increasing its length could help a lot.

Since we can configure each filter, I've been playing with a lot of combinations.  I determined, however, that I was just doing random combinations to see what happened, and I need better data to see what is going on.  I plan on collecting some raw data to show you guys, but I haven't had an opportunity to do that yet.  I am going to go ahead an merge this in, since there is no rule that says you can't use the same filter more than once.  Thus, I COULD put TWO change filters in, one for low change values, one for high change values, and if needed, I could make the "max skips" so high as to have the effect to disable the feature.  I think it gives a lot of flexibility as currently written. Once I put some more diagnostics in, I can add some documentation to it to give advice to end users on its proper use.

PR #61, opened 2019-08-05
Reset GPS after boot

mairas:
The NS-HP-GN GPS modules didn't seem to boot reliably after power-up, so I added some code to reset them after the startup.

Also: properly set all time struct fields when decoding GPS messages.

Also also: return the GPSInput object so that the wiring can be amended afterwards.

### DIFF
--- a/src/system/nmea_parser.cpp
+++ b/src/system/nmea_parser.cpp
@@ -519,6 +519,9 @@ void PSTI032SentenceParser::parse(char* buffer, int term_offsets[], int num_term
     return;
   }
 
+  time.tm_sec = (int)second;
+  time.tm_isdst = 0;
+
   if (is_valid) {
     nmea_data->datetime.set(mktime(&time));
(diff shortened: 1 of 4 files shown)


src/system/nmea_parser.cpp
```diff
@@ -519,6 +519,9 @@ void PSTI032SentenceParser::parse(char* buffer, int term_offsets[], int num_term
     return;
   }
 
+  time.tm_sec = (int)second;
+  time.tm_isdst = 0;
```
ba58smith: I assume time.tm_isdst is the setting for Daylight Savings Time. If so, is it appropriate to always have that be 0?

joelkoz: I THINK the answer is that the time is stored in UTC, which does not have DLS

mairas: Yes, that is the case.

PR #62, opened 2019-08-14
Make read_delay in DigitalInputCounter configurable

ba58smith:


mairas:
Dammit. That was me but on my work account...

ba58smith:
@mairas - Well, at least I can usually figure out how to copy someone else's code. As long as it doesn't require too many changes!

src/sensors/digital_input.cpp
```diff
@@ -61,4 +61,34 @@ void DigitalInputCounter::enable() {
...
+  JsonObject& root = buf.createObject();
+  root["read_delay"] = read_delay;
+  root["value"] = output;
+  return root;
+  };
+
+  static const char SCHEMA[] PROGMEM = R"###({
+    "type": "object",
+    "properties": {
+        "read_delay": { "title": "Read delay", "type": "number" },
```
joelkoz: Just FYI - you can add an optional "description" field to the property definition to give a little more information if you'd like.  For example:

``
{ "title": "Read delay", "type": "number", "description": "A delay, in milliseconds, between each read of the input" },
``

ba58smith: I will do just that! Thanks.

ba58smith: Done. I will merge in a little bit. Taking the boat to the yard for a new coat of bottom paint.

PR #63, opened 2019-09-14
Configurable NMEA input stream

mairas:
NMEA Input streams are now configurable and the software serial library can be used to handle the serial receive pin.

Also removed the resetting logic because it was very specific to my own setup and shouldn't have been added to SensESP in the first place.


mairas:
The changes are relatively benign - merging by myself.

PR #67, opened 2019-10-28
Remove compiler warnings

joelkoz:
Fixed some simple compiler warnings

### DIFF
--- a/src/sensesp_app.cpp
+++ b/src/sensesp_app.cpp
@@ -110,7 +110,7 @@ void SensESPApp::enable() {
 
   // connect all transforms to the Signal K delta output
 
-  ObservableValue<String>* hostname = networking->get_hostname();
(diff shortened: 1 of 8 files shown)


src/sensesp_app.cpp
```diff
@@ -110,7 +110,7 @@ void SensESPApp::enable() {
 
   // connect all transforms to the Signal K delta output
 
-  ObservableValue<String>* hostname = networking->get_hostname();
+  // ObservableValue<String>* hostname = networking->get_hostname();
```
mairas: Is this really not needed? My mind boggles.

joelkoz:
Well, hostname is a local variable in this instance, and that local variable was not referenced anywhere in that method.  I suppose its possible that it was supposed to set a member property vs. a local variable?  I didn't write that code, however.

PR #79, opened 2020-03-11
Configpath msgs

ba58smith:
Make the debug message for `Configurable::load_configuration()` specific to the cause of the message (no config path was specified, or no file was found in SPIFFS).
I did this because I often saw one or the other of these messages in the Serial Monitor that didn't make sense, so I wanted to make sure I understood the message. Now that I've done that, I've uncovered what I think is a bug: the error message that results from `if (!SPIFFS.exists(config_path))` happens for many sensors and transports when it doesn't appear to be correct. See Issue #88 .

### DIFF
--- a/src/sensesp_app.cpp
+++ b/src/sensesp_app.cpp
@@ -21,7 +21,7 @@ RemoteDebug Debug;
 
 // FIXME: Setting up the system is too verbose and repetitive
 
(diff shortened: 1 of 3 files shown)


ba58smith:
Somehow I got this PR jumbled up with the previous one. They're both simple, so hopefully both can just be merged with no problems.

PR #81, opened 2020-03-21
Add sensor, transforms, and example for ADS1x15 4-channel ADC

ba58smith:
My most ambitious PR by far! It all works great with my ADS1115 chip, reading two different battery banks (12V and 24V). I did not test the reading of the difference between two channels, but that's simply a different function call from the Adafruit library. (Different from the one that reads from a single channel.)

ADS1x15Voltage has a parameter that represents the chip (ADS1015 or ADS1115) you're using: a uint8_t that's 0 for an ADS1015 and 1 for an ADS1115. Should I have done a typedef to give those two chips a specific name and type, instead of just 0 and 1?


ba58smith:
In ADS1x15_voltage, I did implement an enum to give the two chip types a name. And I tested channel 10, which represents the reading of the difference between channels 1 and 0. I didn't test channels 2 or 3, nor 23 (the diff btwn 2 and 3), but they work exactly the same as channels 0, 1, and 10.

This has been working on my boat for over a week now. I'm going to merge it.

PR #92, opened 2020-03-28
Add 'last value' to moving_avg config data

ba58smith:
The moving_average transport didn't show its last value when looking at its configuration data. I've found it helpful to have that when troubleshooting output, so I added it.


ba58smith:
It's been working great, and no one has had time to review it, so I'm going to merge it.

PR #96, opened 2020-04-04
Add BME280 sensor with example

ba58smith:
Per a request from someone on Slack. Issue #96 .

### DIFF
--- a/examples/bme280_example.cpp
+++ b/examples/bme280_example.cpp
@@ -26,35 +26,35 @@ ReactESP app([] () {
 
   sensesp_app = new SensESPApp(disableStandardSensors);
 
-// Create a BME280, which represents the physical sensor.
-// 0x77 is the default address. Some chips use 0x76, which is shown here.
-auto* pBME280 = new BME280(0x76);
+  // Create a BME280, which represents the physical sensor.
+  // 0x77 is the default address. Some chips use 0x76, which is shown here.
+  auto* pBME280 = new BME280(0x76);
 
-// If you want to change any of the settings that are set by Adafruit_BME280::setSampling(), do
-// that here, like this:
-// pBME280->pAdafruitBME280->setSampling(); // pass in the parameters you want
+  // If you want to change any of the settings that are set by Adafruit_BME280::setSampling(), do
+  // that here, like this:
+  // pBME280->pAdafruitBME280->setSampling(); // pass in the parameters you want
 
 
-// Define the read_delays you're going to use:
-const uint read_delay = 1000; // once per second
-const uint pressure_read_delay = 60000; // once per minute
+  // Define the read_delays you're going to use:
+  const uint read_delay = 1000; // once per second
+  const uint pressure_read_delay = 60000; // once per minute
 
-// Create a BME280value, which is used to read a specific value from the BME280, and send its output
-// to SignalK as a number (float). This one is for the temperature reading.
-auto* pBMEtemperature = new BME280value(pBME280, temperature, read_delay, "Outside/Temperature");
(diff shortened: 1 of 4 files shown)


examples/bme280_example.cpp
```diff
@@ -0,0 +1,63 @@
...
+  delay(100);
+  Debug.setSerialEnabled(true);
+  #endif
+
+  // true will disable systemHz, freemem, uptime, and ipaddress "sensors"
+  bool disableStandardSensors = false;
+
+  sensesp_app = new SensESPApp(disableStandardSensors);
+
+// Create a BME280, which represents the physical sensor.
```
mairas, line 29: From here onwards, the indentation in the file swings back and forth. Maybe you could let your editor reformat the file?

src/sensors/bme280.cpp
```diff
@@ -0,0 +1,87 @@
...
+      className = "BME280value";
+      load_configuration();
+}
+
+// BME280 outputs temp in Celsius. Need to convert to Kelvin before sending to Signal K.
+// Pressure is output in Pascals, Humidity is output in relative humidity (0 - 100%)
+void BME280value::enable() {
+  app.onRepeat(read_delay, [this](){ 
+      if (val_type == temperature) {
+          output = pBME280->pAdafruitBME280->readTemperature() + 273.15; // Kelvin is Celsius + 273.15
```
mairas, line 44: You've got this backwards: now you're moving from Kelvin to °C. I wonder is that correct? Temperatures in SK should be in Kelvin because that's the base unit.

mairas:
Looks really good except for the two comments!

src/sensors/bme280.cpp
ba58smith: I don't think it's backwards. Kelvin is Celsius + 273.15. That's what I'm doing: readTemperature() returns temp in Celsius, so I'm adding 273.15 to that to get the value for `output`.

examples/bme280_example.cpp
ba58smith: Fixed.

src/sensors/bme280.cpp
ba58smith: This is how it's done for one_wire_temp:
```
void OneWireTemperature::read_value() {
  // getTempC returns degrees Celsius but SignalK expects Kelvins
  output = dts->sensors->getTempC(address.data()) + 273.15;
  this->notify();
}
```

KEGustafsson:
If BME280 device is not present or broken then SensESP startup fails and device is halted.
`␛[0m(I) (check_status) Could not find a valid BME280 sensor. Check wiring, address, and sensor ID.
␛[0m(I) (check_status) SensorID is: 0x255
␛[0m(I) (check_status) 0xFF: is a BMP180 or BMP085, or a bad address
␛[0m(I) (check_status) 0x56-0x58 is a BMP280
␛[0m(I) (check_status) 0x60 is a BME280
␛[0m(I) (check_status) 0x61 is a BME680
␛[0m`
Log reporting stops here.

If SensorID is 0x255, should I2C initialization be skipped?

mairas:
> Is this needed?
> https://github.com/SignalK/SensESP/blob/master/src/sensors/bme280.cpp#L27

I'd say that's harmful. Would be much better to catch the issue internally and just disable reading the values for the sensor and allow the rest of the program to continue.

Recommend opening an issue, though. Discussions in already merged PRs are quite difficult to find for most people. :-)

KEGustafsson:
I'll do that and I'll copy your comment as an option to overcome the issue.

Issue #94, opened 2020-04-03
Add support for BMP280 temp/pressure sensor 


ba58smith:
After I said I'd write the BME280, someone said they wanted the BMP280, and they're so similar, I said I'd do the BMP280, too.

ba58smith:
See PR #97 .

ba58smith:
PR #97 merged.

PR #97, opened 2020-04-04
Add BMP280 sensor

ba58smith:
Per a request from somone on Slack, after I said I'd do the BME280. Issue #94 .


mairas:
Duplicate. :-)

ba58smith:
> Duplicate. :-)

No, it's not. This is the BMP280 sensor. The other one is the BME280. The former doesn't read humidity. More important, they use different libraries, or I would have put them both in the same .h/.cpp files. I asked @joelkoz about this, and he said that including both libraries, when only one of them would be needed for each instance, would make the code unnecessesarily large.

PR #107, opened 2020-04-12
Improve method for disabling std sensors, incl in examples

ba58smith:


ba58smith:
Tested thoroughly, works great.

Issue #86, opened 2020-03-23
"Unconfigured SensESP Device" SSID s/b unique


ba58smith:
If you happen to have two unconfigured SensESP devices, they both get the same the SSID ("Unconfigured SensESP Device"), and the same IP address.  It would be better if the name were unique to the device, using some part of the device's Hostname to make it unique. It would be best if each had a unique name AND a unique IP address.

This problem is made worse by the fact that some ESP's will broadcast their SSID (with the name "Unconfigured SensESP Device") even after they have successfully connected to the network. (That's another issue, of course. Issue # 87, in fact. )

ba58smith:
I've spent hours looking at this, and I can't make any progress on it. I can't sort out the confusion between the `hostname` and the `SSID`, and also, I think the SSID may be cached on the ESP or on all of my wifi-enabled devices, because even when I change the string "Unconfigured SensESP Device" to something else, it STILL broadcasts "Unconfigured SensESP Device".

mairas:
I can have a look soon(ish). Tasmota would be a good reference project for this one - the Tasmota devices advertise with an SSID like Tasmota-1234 (don't remember the exact format), where the number is unique for each sensor, probably taken from the MAC address of the wifi device.

ba58smith:
One challenge with this is that there seems to be a bug in WiFiManager in which the SSID and PW of the saved network are not erased from the flash memory on the ESP8266 when `resetSettings()` is called (which just calls `disconnect()`). This is still an open Issue with WiFiManager: https://github.com/tzapu/WiFiManager/issues/142.

ba58smith:
See PR #108.

ba58smith:
Fixed with PR #108

Issue #87, opened 2020-03-23
Some ESP's broadcast their AP when they shouldn't


ba58smith:
This may be a firmware issue with some clones of the original Wemos D1 Mini - not sure. But some ESP8266's will activate the "Unconfigured SensESP Device" AP even after successfully connecting to the network and sending deltas to Signal K. If you connect to it, and go to 192.168.4.1, you see the config web page, the same as if you go to the IP address of the ESP. Same config web page, and both are active.

EdKok:
It is not just the case for clones.
I'm using an original Wemos D1 miniPro v1.0.0 which does this too.
That is an older type, but original: https://wiki.wemos.cc/products:retired:d1_mini_pro_v1.0.0
(as I write this the wiki is down, but it really exists)
I choose this model because of its external WiFi-antenna which is great for better connection from the engine-room.
The ESP is working perfectly with four DS18B20's attached, but vulnerably 'open' to hackers. Even worse, after connecting and going to 'system'->'networking' it shows the SSID plus password for your AP in plain text!

ba58smith:
@EdKok - I don't think it's vulnerable to hackers. The only way you can see system->networking is if you're already connected to your boat's wifi. The only time anything is open is when the Wemos is trying to connect to your wifi, and it can't, so it launches the configuration portal (with the blue buttons). And when that comes up, there is nothing in the network SSID and password fields.

ba58smith:
@EdKok - I do agree with you that it's not just clones that do it. Although I never saw it before with my genuine Wemos's, I'm seeing it now. I'm still poking around on Google for a solution, but it really is harmless, I think. Just a bit annoying.

EdKok:
> @EdKok - I don't think it's vulnerable to hackers. The only way you can see system->networking is if you're already connected to your boat's wifi. The only time anything is open is when the Wemos is trying to connect to your wifi, and it can't, so it launches the configuration portal (with the blue buttons). And when that comes up, there is nothing in the network SSID and password fields.

I tested this today by connecting my Windows10 pc to the SSID ‘unconfigered SensESP device’ while it was serving data to signalk. After opening a browser to 192.168.4.1 it showed me all info in clear text.

ba58smith:
@EdKok - yep, you're right. I just did the same thing. Well, the good news is that someone would have to be pretty close to you to connect to the ESP's wifi. But it's still a vulnerability, and it gives me the perfect reason to open an issue on the ESPAsyncWifiManager github page.

[...]

PR #108, opened 2020-04-13
Unique SSID for each ESP

ba58smith:
See Issue #86.

### DIFF
--- a/examples/bme280_example.cpp
+++ b/examples/bme280_example.cpp
@@ -21,10 +21,9 @@ ReactESP app([] () {
   Debug.setSerialEnabled(true);
   #endif
 
-  // true will disable systemHz, freemem, uptime, and ipaddress "sensors"
-  bool disableStandardSensors = false;
-
-  sensesp_app = new SensESPApp(disableStandardSensors);
+  // Create the SensESPApp with whatever "standard sensors" you want: noStdSensors, allStdSensors, or uptimeOnly.
+  // The default is allStdSensors.
+  sensesp_app = new SensESPApp(uptimeOnly);
 
   // Create a BME280, which represents the physical sensor.
   // 0x77 is the default address. Some chips use 0x76, which is shown here.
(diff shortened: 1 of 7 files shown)


ba58smith:
DO NOT MERGE THIS YET. Although it does what I intended it to do, there appears to be a bug way down deep in Wifi Manager that doesn't erase the SSID and pw of the configuration UI when `resetSettings()` is called. If it's not erasing those, then it's probably also not erasing any custom parameters, which is what this PR is all about. https://github.com/tzapu/WiFiManager/issues/142

mairas:
Looks good to me!

ba58smith:
Matti,

It works, but there's still something not quite right about it. Sometimes,
when the ESP8266 is connected to the SK server and sending data, it's still
broadcasting the configuration SSID. For example, right now, in my list of
available wifi networks, I see "Configure MEtemps". MEtemps is the name of
one of my ESP8266's. I expect to see that SSID in my list of available
networks ONLY when the ESP can't connect to my boat's wifi, and it starts
up the AP mode so that the initial configuration can be done (of wifi SSID
and password, and the custom hostname for the Device). When I connect to
that SSID, and go to 192.168.4.1, I get the configuration UI that you and
JKoz wrote (where you configure the sensors and transports), not the config
UI for setting up the wifi and custom hostname. I hope that makes sense.

I'm wondering if somehow the mode in WifiManager is not being set
correctly, but I can't figure it out. Do you remember how this works well
enough to have any thoughts on it? If so, I'd love to hear them. Otherwise,
I think I'll start digging into the different WifiManager modes, and see if
this has something to do with the wrong mode being active.

Thanks,

@butch

Brian "Butch" Smith
M/V Smartini
www.smartini.life

ba58smith:
OK, this PR works. Each of my four ESP's is now broadcasting a custom SSID when my onboard wifi is turned off (the wife that all four of the ESP's are supposed to connect to). 
If you ever have what appears to be an outdated SSID being broadcast for a particular ESP, try the PlatformIO option "Erase Flash", and then upload the new firmware.
Issue #87 is still open, and is made even more obvious by this PR, as you'll now see a separate SSID being broadcast by all of your ESP's any time your wifi is not available.

PR #109, opened 2020-04-14
Add SKValueListener and DigitalOutput

JohnySeven:
Added support for listening SignalK paths using SignalK subscription over WS. 
Added few classes to support subscribing to data on server. 

I've created class that represents value listerner called **SKValueListerner<T>**, that is **ValueProducer<T>**, so using what is already there I can connect value in SignalK to for example **DigitalOutput** (new class added). 

This is sample code for connecting navigation.lights value to D1 digital output for switching relay:

```
  auto * listener = new SKValueListener<bool>("navigation.lights", 500);
  listener->attach([]()
  {
    debugI("Navigation lights status has been updated!");
  });
  listener->connectTo(new DigitalOutput(5));
```

I've tested this code using D1 mini with relay HAT and SignalK Node Server and it works just fine.

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,8 +84,7 @@ Configuration can be updated with HTTP PUT requests:
 - [x] Authentication token support
 - [x] Make the project a library
 - [x] Web configuration UI
-- [ ] Control device support. For now, all devices are read-only, and controls such as leds, relays, or
-PWM output are not supported.
+- [x] Control device support. Support for SKValueListener is working with WS subscription. Relay sample is working, LEDs should work.
 
 SensESP Class Diagram
 ----------------------------
--- a/examples/relay_control.cpp
+++ b/examples/relay_control.cpp
@@ -1,8 +1,6 @@
 #include <Arduino.h>
 
 #include "sensesp_app.h"
-#include "sensors/analog_input.h"
-#include "transforms/linear.h"
 #include "signalk/signalk_output.h"
 #include "signalk/signalk_listener.h"
 #include "signalk/signalk_value_listener.h"
--- a/examples/relay_control.cpp
+++ b/examples/relay_control.cpp
@@ -1,0 +1,52 @@
+#include <Arduino.h>
+
+#include "sensesp_app.h"
+#include "sensors/analog_input.h"
+#include "transforms/linear.h"
+#include "signalk/signalk_output.h"
+#include "signalk/signalk_listener.h"
+#include "signalk/signalk_value_listener.h"
+#include "sensors/digital_output.h"
+#include "transforms/threshold.h"
+
+
+// Following sample shows how to listen for value on Signal K server.
+// This sample demonstrates how to make automatic navigation light switching device depending on light conditions.
+
+// SensESP builds upon the ReactESP framework. Every ReactESP application
+// defines an "app" object vs defining a "main()" method.
+ReactESP app([] () {
+
+  // Some initialization boilerplate when in debug mode...
+  #ifndef SERIAL_DEBUG_DISABLED
+  Serial.begin(115200);
+
+  // A small arbitrary delay is required to let the
+  // serial port catch up
+  delay(100);
+  Debug.setSerialEnabled(true);
+  #endif
+
+
+  // Create the global SensESPApp() object.
+  sensesp_app = new SensESPApp();
+  // The "SignalK path" identifies subscription of data from the SignalK network.
+  const char* sk_path = "environment.outside.illuminance";
+  // The "Configuration path" is combined with "/config" to formulate a URL
+  // used by the RESTful API for retrieving or setting configuration data.
+  // It is ALSO used to specify a path to the SPIFFS file system
+  // where configuration data is saved on the MCU board.  It should
+  // ALWAYS start with a forward slash if specified.  If left blank,
+  // that indicates this sensor or transform does not have any
+  // configuration to save.
+  // Note that if you want to be able to change the sk_path at runtime,
+  // you will need to specify a config_path.
+  const char* config_path = "/threshold/lights";
+  // Wire up the output of the float value on server "environment.outside.illuminance" to the NumericThreshold,
+  // and then output the transformed float to boolean to DigitalOutput
+  auto * listener = new SKOutputNumber(sk_path);
+  listener->connectTo(new NumericThreshold(0.0f, 100.0f, true, config_path), 0)
+          ->connectTo(new DigitalOutput(5));
+  // Start the SensESP application running
+  sensesp_app->enable();
+});
--- a/src/net/ws_client.cpp
+++ b/src/net/ws_client.cpp
@@ -107,15 +107,15 @@ void WSClient::subscribe_listeners() {
     for (size_t i = 0; i < listeners.size(); i++) {
       auto* listener = listeners.at(i);
       String sk_path = listener->get_sk_path();
-      int period = listener->getPeriod();
+      int listenDelay = listener->getListenDelay();
 
       JsonObject& subscribePath = subscribe.createNestedObject();
 
       subscribePath["path"] = sk_path;
-      subscribePath["period"] = period;
+      subscribePath["period"] = listenDelay;
 
-      debugI("Adding %s subscription with period %d\n", sk_path.c_str(),
-             period);
+      debugI("Adding %s subscription with listen delay %d\n", sk_path.c_str(),
+             listenDelay);
     }
 
     String messageJson;
--- a/src/net/ws_client.cpp
+++ b/src/net/ws_client.cpp
@@ -5,9 +5,9 @@
 #include <ArduinoJson.h>
 #include <ESP8266HTTPClient.h>
 #ifdef ESP8266
-  #include <ESP8266mDNS.h>        // Include the mDNS library
+#include <ESP8266mDNS.h>  // Include the mDNS library
 #elif defined(ESP32)
-  #include <ESPmDNS.h>
+#include <ESPmDNS.h>
 #endif
 
 #include <ESP8266TrueRandom.h>
@@ -16,10 +16,12 @@
 
 #include "sensesp_app.h"
 
+#include "signalk/signalk_listener.h"
+
 WSClient* ws_client;
 
-void webSocketClientEvent(WStype_t type, uint8_t * payload, size_t length) {
-  switch(type) {
+void webSocketClientEvent(WStype_t type, uint8_t* payload, size_t length) {
+  switch (type) {
     case WStype_DISCONNECTED:
       ws_client->on_disconnected();
       break;
@@ -35,12 +37,13 @@ void webSocketClientEvent(WStype_t type, uint8_t * payload, size_t length) {
     default:
       // Do nothing for other types
       break;
-   }
+  }
 }
 
 WSClient::WSClient(String config_path, SKDelta* sk_delta,
                    std::function<void(bool)> connected_cb,
-                   void_cb_func delta_cb) : Configurable{config_path} {
+                   void_cb_func delta_cb)
+    : Configurable{config_path} {
   this->sk_delta = sk_delta;
   this->connected_cb = connected_cb;
   this->delta_cb = delta_cb;
@@ -52,26 +55,26 @@ WSClient::WSClient(String config_path, SKDelta* sk_delta,
 }
 
 void WSClient::enable() {
-  app.onDelay(0, [this](){ this->connect(); });
-  app.onRepeat(20, [this](){ this->loop(); });
-  app.onRepeat(100, [this](){ this->send_delta(); });
-  app.onRepeat(10000, [this](){ this->connect_loop(); });
+  app.onDelay(0, [this]() { this->connect(); });
+  app.onRepeat(20, [this]() { this->loop(); });
+  app.onRepeat(100, [this]() { this->send_delta(); });
+  app.onRepeat(10000, [this]() { this->connect_loop(); });
 }
 
 void WSClient::connect_loop() {
-  if (this->connection_state==disconnected) {
+  if (this->connection_state == disconnected) {
     this->connect();
   }
 }
 
 void WSClient::on_disconnected() {
   if (this->connection_state == connecting && server_detected) {
-     // Going from connecting directly to disconnect when we
-     // know we have found and talked to the server usually means
-     // the authentication token is bad.
-     debugW("Bad access token detected. Setting token to null.");
-     auth_token = NULL_AUTH_TOKEN;
-     save_configuration();
+    // Going from connecting directly to disconnect when we
+    // know we have found and talked to the server usually means
+    // the authentication token is bad.
+    debugW("Bad access token detected. Setting token to null.");
+    auth_token = NULL_AUTH_TOKEN;
+    save_configuration();
   }
   this->connection_state = disconnected;
   server_detected = false;
@@ -84,20 +87,86 @@ void WSClient::on_error() {
   this->connected_cb(false);
 }
 
-void WSClient::on_connected(uint8_t * payload) {
+void WSClient::on_connected(uint8_t* payload) {
   this->connection_state = connected;
   debugI("Websocket client connected to URL: %s\n", payload);
   this->connected_cb(true);
+  debugI("Subscribing to SignalK listeners...");
+  this->subscribe_listeners();
 }
 
-void WSClient::on_receive_delta(uint8_t * payload) {
+void WSClient::subscribe_listeners() {
+  const std::vector<SKListener*>& listeners = SKListener::get_listeners();
+
+  if (listeners.size() > 0) {
+    DynamicJsonBuffer jsonBuffer;
+    JsonObject& subscription = jsonBuffer.createObject();
+    subscription["context"] = "vessels.self";
+    JsonArray& subscribe = subscription.createNestedArray("subscribe");
+
+    for (size_t i = 0; i < listeners.size(); i++) {
+      auto* listener = listeners.at(i);
+      String sk_path = listener->get_sk_path();
+      int period = listener->getPeriod();
+
+      JsonObject& subscribePath = subscribe.createNestedObject();
+
+      subscribePath["path"] = sk_path;
+      subscribePath["period"] = period;
+
+      debugI("Adding %s subscription with period %d\n", sk_path.c_str(),
+             period);
+    }
+
+    String messageJson;
+
+    subscription.printTo(messageJson);
+    debugI("Subscription JSON message:\n %s", messageJson.c_str());
+    this->client.sendTXT(messageJson);
+  }
+}
+
+void WSClient::on_receive_delta(uint8_t* payload) {
+  #ifdef SIGNALK_PRINT_RCV_DELTA
   debugD("Websocket payload received: %s", (char*)payload);
+  #endif
+  
+  DynamicJsonBuffer jsonBuffer;
+  JsonObject& message = jsonBuffer.parseObject(String((char*)payload));
+
+  if (message.success()) {
+    JsonArray& updates = message["updates"];
+
+    for (size_t i = 0; i < updates.size(); i++) {
+      JsonObject& update = updates[i];
+
+      JsonArray& values = update["values"];
+
+      for (size_t vi = 0; vi < values.size(); vi++) {
+        JsonObject& value = values[vi];
+        
+        const char* path = value["path"];
+        //debugD("Got update of value %s\n", path);
+
+        const std::vector<SKListener*>& listeners = SKListener::get_listeners();
+
+        for (size_t i = 0; i < listeners.size(); i++)
+        {
+          SKListener* listener = listeners[i];
+          if(listener->get_sk_path().equals(path))
+          {
+            listener->parseValue(value);
+          }
+        }        
+      }
+    }
+  }
 }
 
 bool WSClient::get_mdns_service(String& server_address, uint16_t& server_port) {
   // get IP address using an mDNS query
   int n = MDNS.queryService("signalk-ws", "tcp");
-  if (n==0) {
+  if (n == 0) {
     // no service found
     return false;
   } else {
@@ -109,7 +178,7 @@ bool WSClient::get_mdns_service(String& server_address, uint16_t& server_port) {
 }
 
 void WSClient::connect() {
-  if (connection_state!=disconnected) {
+  if (connection_state != disconnected) {
     return;
   }
   debugD("Initiating connection");
@@ -149,13 +218,14 @@ void WSClient::connect() {
   this->test_token(server_address, server_port);
 }
 
-void WSClient::test_token(const String server_address, const uint16_t server_port) {
-
+void WSClient::test_token(const String server_address,
+                          const uint16_t server_port) {
   // FIXME: implement async HTTP client!
   WiFiClient client;
   HTTPClient http;
 
-  String url = String("http://") + server_address + ":" + server_port + "/signalk/v1/api/";
+  String url = String("http://") + server_address + ":" + server_port +
+               "/signalk/v1/api/";
   debugD("Testing token with url %s", url.c_str());
   http.begin(client, url);
   String full_token = String("JWT ") + auth_token;
@@ -169,8 +239,7 @@ void WSClient::test_token(const String server_address, const uint16_t server_por
       debugD("Returned payload (length %d) is: ", payload.length());
       debugD("%s", payload.c_str());
       debugD("End of payload output");
-    }
-    else {
+    } else {
       debugD("Returned payload is empty");
     }
     if (httpCode == 200) {
@@ -189,7 +258,8 @@ void WSClient::test_token(const String server_address, const uint16_t server_por
   }
 }
 
-void WSClient::send_access_request(const String server_address, const uint16_t server_port) {
+void WSClient::send_access_request(const String server_address,
+                                   const uint16_t server_port) {
   debugD("Preparing a new access request");
   if (client_id == "") {
     // generate a client ID
@@ -203,16 +273,15 @@ void WSClient::send_access_request(const String server_address, const uint16_t s
   DynamicJsonBuffer buf;
   JsonObject& req = buf.createObject();
   req["clientId"] = client_id;
-  req["description"] =
-    String("SensESP device: ") + sensesp_app->get_hostname();
+  req["description"] = String("SensESP device: ") + sensesp_app->get_hostname();
   String json_req = "";
   req.printTo(json_req);
 
   WiFiClient client;
   HTTPClient http;
 
-  String url = String("http://") + server_address + ":" + server_port
-      + "/signalk/v1/access/requests";
+  String url = String("http://") + server_address + ":" + server_port +
+               "/signalk/v1/access/requests";
   http.begin(client, url);
   http.addHeader("Content-Type", "application/json");
   int httpCode = http.POST(json_req);
@@ -243,12 +312,14 @@ void WSClient::send_access_request(const String server_address, const uint16_t s
   save_configuration();
 
   debugD("Polling %s in 5 seconds", polling_href.c_str());
-  app.onDelay(5000, [this, server_address, server_port](){
+  app.onDelay(5000, [this, server_address, server_port]() {
     this->poll_access_request(server_address, server_port, this->polling_href);
   });
 }
 
-void WSClient::poll_access_request(const String server_address, const uint16_t server_port, const String href) {
+void WSClient::poll_access_request(const String server_address,
+                                   const uint16_t server_port,
+                                   const String href) {
   debugD("Polling SK Server for authentication token");
 
   WiFiClient client;
@@ -265,7 +336,9 @@ void WSClient::poll_access_request(const String server_address, const uint16_t s
     String state = resp["state"];
     debugD("%s", state.c_str());
     if (state == "PENDING") {
-      app.onDelay(5000, [this, server_address, server_port, href](){ this->poll_access_request(server_address, server_port, href); });
+      app.onDelay(5000, [this, server_address, server_port, href]() {
+        this->poll_access_request(server_address, server_port, href);
+      });
       return;
     } else if (state == "COMPLETED") {
       JsonObject& access_req = resp["accessRequest"];
@@ -282,13 +355,15 @@ void WSClient::poll_access_request(const String server_address, const uint16_t s
         String token = access_req["token"];
         auth_token = token;
         save_configuration();
-        app.onDelay(0, [this, server_address, server_port](){ this->connect_ws(server_address, server_port); });
+        app.onDelay(0, [this, server_address, server_port]() {
+          this->connect_ws(server_address, server_port);
+        });
         return;
       }
     }
   } else {
     http.end();
-    if (httpCode==500) {
+    if (httpCode == 500) {
       // this is probably the server barfing due to
       // us polling a non-existing request. Just
       // delete the polling href.
@@ -314,16 +389,12 @@ void WSClient::connect_ws(const String host, const uint16_t port) {
   this->client.setAuthorization(full_token.c_str());
 }
 
-void WSClient::loop() {
-  this->client.loop();
-}
+void WSClient::loop() { this->client.loop(); }
 
-bool WSClient::is_connected() {
-  return connection_state==connected;
-}
+bool WSClient::is_connected() { return connection_state == connected; }
 
 void WSClient::restart() {
-  if (connection_state==connected) {
+  if (connection_state == connected) {
     this->client.disconnect();
     connection_state = disconnected;
   }
@@ -333,7 +404,7 @@ void WSClient::send_delta() {
   String output;
   if (sk_delta->data_available()) {
     sk_delta->get_delta(output);
-    if (connection_state==connected) {
+    if (connection_state == connected) {
       this->client.sendTXT(output);
       this->delta_cb();
     }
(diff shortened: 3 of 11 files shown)


ba58smith:
This all looks good to me, but I am NOT the one who needs to review this for real substance. (I'm the new guy - this is above my pay grade.) @mairas and/or @joelkoz need to do that. I have only a few minor comments.

How do you specify the value that the listener is listening for? For example, if I want to listen to `environment.outdoor.illuminance` (the ambient light outside as read by an analog light sensor), and I want to turn on my anchor light if the light level is lower than XX, how do I specify XX?

The "period" in all other sensors is called "read_delay". To be consistent, can you change "period" to  "read_delay", or maybe "listen_delay" is more appropriate?

Please provide a default value for this member (whatever you change `period` to be). Maybe 500ms? Maybe even 1000ms? Or less, if you can think of a use case that would need to know about a change more often than that.

Please create a value_listener.cpp example, for `/examples'. I suggest you copy and modify one of the simple examples that is already there, maybe `analog_input.cpp`. Change it from "indoor" to "outdoor", and use the Listener to turn on the anchor light. That should be easy.

README.md
```diff
@@ -84,8 +84,7 @@ Configuration can be updated with HTTP PUT requests:
 - [x] Authentication token support
 - [x] Make the project a library
 - [x] Web configuration UI
-- [ ] Control device support. For now, all devices are read-only, and controls such as leds, relays, or
-PWM output are not supported.
+- [x] Control device support. Support for SKValueListener is working with WS subscription. Relay sample is working, LEDs should work.
```
mairas, line 87: This must've felt good! :-D

mairas:
Brilliant! Thanks a lot for this! I like the implementation, it's clean and easy to follow, and it adds a major piece of missing functionality.

One request, one comment:
- The sample code in this PR description will be pretty much lost after the PR is merged. Since this PR provides quite significant new functionality, would you mind creating a proper example in the examples subdirectory?
- Getting code formatting fixes is great as well, but for PR reviews it would be slightly more convenient if they were in separate commits. No need to start splitting the commits in this PR, of course.

ba58smith:
Maybe add another optional parameter to DigitalOutput::DigitalOutput() that is the value you're looking for (such as "thresholdValue"), and then in DigitalOutput::set_input(), compare newValue to thresholdValue to determine whether or not send something to the pin. Of course, that means the input needs to be able to be a bool, an int, or a float.

JohnySeven:
> This all looks good to me, but I am NOT the one who needs to review this for real substance. (I'm the new guy - this is above my pay grade.) @mairas and/or @joelkoz need to do that. I have only a few minor comments.
> 
> How do you specify the value that the listener is listening for? For example, if I want to listen to `environment.outdoor.illuminance` (the ambient light outside as read by an analog light sensor), and I want to turn on my anchor light if the light level is lower than XX, how do I specify XX?

My design for this is that I've BH1750 sensor, that is sending value to Signal K "environment.outside.illuminance" as you mentioned. On RPI side in Node-Red I've done subscription for changes of 'environment.outside.illuminance' and there is little logic that will change value of 'navigation.lights' to 'true' or 'false' when illuminance is less than 100 lux.

But in SigSens there could be some kind of transformation that will accept some threshold and it could return true if value is less than threshold or false otherwise.
> 
> The "period" in all other sensors is called "read_delay". To be consistent, can you change "period" to "read_delay", or maybe "listen_delay" is more appropriate?

Sure, I'll change it.
> 
> Please provide a default value for this member (whatever you change `period` to be). Maybe 500ms? Maybe even 1000ms? Or less, if you can think of a use case that would need to know about a change more often than that.

I agree, sometimes you need system to react quickly :-)

> 
> Please create a value_listener.cpp example, for `/examples`. I suggest you copy and modify one of the simple examples that is already there, maybe `analog_input.cpp`. Change it from "indoor" to "outdoor", and use the Listener to turn on the anchor light. That should be easy

Good point :-)

Thanks for your comments, I appreciate them.

README.md
JohnySeven: You bet :-D

JohnySeven:
> Brilliant! Thanks a lot for this! I like the implementation, it's clean and easy to follow, and it adds a major piece of missing functionality.

No need to, I'm happy to contribute with my little stuff, I've done :-)

> 
> One request, one comment:
> 
> * The sample code in this PR description will be pretty much lost after the PR is merged. Since this 
PR provides quite significant new functionality, would you mind creating a proper example in the examples subdirectory?

No problem, I'll update samples with my testing sample called relay_control.cpp if it's appropriate one?

> * Getting code formatting fixes is great as well, but for PR reviews it would be slightly more convenient if they were in separate commits. No need to start splitting the commits in this PR, of course.

Yeah, I'm sorry for that I often use auto formatting in VS Code when writing my own code. I'll do better next time :-)

JohnySeven:
> Maybe add another optional parameter to DigitalOutput::DigitalOutput() that is the value you're looking for (such as "thresholdValue"), and then in DigitalOutput::set_input(), compare newValue to thresholdValue to determine whether or not send something to the pin. Of course, that means the input needs to be able to be a bool, an int, or a float.

I would like more to use transformations for that, some kind of threshold transformation that could do exactly that you've described. It could allow chaining multiple values and producing more flexible outputs. Just idea. What do you think?

Thanks for you comment, I appreciate it!

ba58smith:
> > Maybe add another optional parameter to DigitalOutput::DigitalOutput() that is the value you're looking for (such as "thresholdValue"), and then in DigitalOutput::set_input(), compare newValue to thresholdValue to determine whether or not send something to the pin. Of course, that means the input needs to be able to be a bool, an int, or a float.
> 
> I would like more to use transformations for that, some kind of threshold transformation that could do exactly that you've described. It could allow chaining multiple values and producing more flexible outputs. Just idea. What do you think?
> 
> Thanks for you comment, I appreciate it!

My only problem with DigitalOutput as it currently exists is that it can be used ONLY with boolean input. I think, with a little modification, it could be used with a boolean, an int, or a float (like the ValueConsumer template class), and then ::set_input( ) could handle all of those instances. The end result would still always be the same - the output would be HIGH or LOW.

However - since the boolean case is the one you're interested in, and it's working as-is, I suggest we merge it in this way, and then maybe I'll look at modifying it later to handle floats and ints - OR we can do that with a separate Transport, as you've suggested.

ba58smith:
@JohnySeven - after you've modified the PR for the changes (and the example) requested above, I will merge it. Thanks! This will be very useful to me on my own boat!

JohnySeven:
@ba58smith I need to test this, but this is what I've done so far regarding your comments about being able to handle float, int values:

```cpp
#include <Arduino.h>

#include "sensesp_app.h"
#include "sensors/analog_input.h"
#include "transforms/linear.h"
#include "signalk/signalk_output.h"
#include "signalk/signalk_listener.h"
#include "signalk/signalk_value_listener.h"
#include "sensors/digital_output.h"
#include "transforms/threshold.h"

// SensESP builds upon the ReactESP framework. Every ReactESP application
// defines an "app" object vs defining a "main()" method.
ReactESP app([] () {

  // Some initialization boilerplate when in debug mode...
  #ifndef SERIAL_DEBUG_DISABLED
  Serial.begin(115200);

  // A small arbitrary delay is required to let the
  // serial port catch up
  delay(100);
  Debug.setSerialEnabled(true);
  #endif


  // Create the global SensESPApp() object.
  sensesp_app = new SensESPApp();
  // The "SignalK path" identifies subscription of data from the SignalK network.
  const char* sk_path = "sensors.outdoor.illumination";
  // The "Configuration path" is combined with "/config" to formulate a URL
  // used by the RESTful API for retrieving or setting configuration data.
  // It is ALSO used to specify a path to the SPIFFS file system
  // where configuration data is saved on the MCU board.  It should
  // ALWAYS start with a forward slash if specified.  If left blank,
  // that indicates this sensor or transform does not have any
  // configuration to save.
  // Note that if you want to be able to change the sk_path at runtime,
  // you will need to specify a config_path.
  const char* config_path = "/threshold/lights";
  // Wire up the output of the float value on server "sensors.outdoor.illumination" to the transform,
  // and then output the results on the SignalK network...
  auto * listener = new SKValueListener<float>(sk_path);
  auto * threshold = new NumericThreshold(0.0f, 100.0f, true, false, config_path);

  listener->connectTo(threshold, 0)
          ->connectTo(new DigitalOutput(5));
  // Start the SensESP application running
  sensesp_app->enable();
});
```
I've created class that transform float/int to boolean value. What do you think about this sample?

ba58smith:
@JohnySeven - without seeing your code for `Threshold` and `NumericThreshold` I can't really comment, but if it has to work the way you've done it in the example, that's different from how other examples work. To make it like all the other examples (all of them written and modified since `connectTo` was implemented, at least), would look like this:
```
auto * listener = new SKValueListener<float>(sk_path);

listener->connectTo(new NumericThreshold(0.0f, 100.0f, true, false, config_path)
            ->connectTo(new DigitalOutput(5));

  // Start the SensESP application running
  sensesp_app->enable();
```
In other words, you would not create the NumericThreshold with all the parameters, outside of the string of `connectTo`'s. If you can make it work like I've shown here, then I don't have a real objection to it. And in fact, if I think of `DigitalOutput` as being like `SKOutput`, I guess it's consistent - the final output (to a digital pin, or to the SK Server) is handled by a special transport that does nothing else.

ba58smith:
One little thing: `SKValueListener::SkValueListener()` should have `Enable::className = "SKValueListener";` in the body of the constructor.

ba58smith:
Now.... before you get too far... how about making SKValueListener be the class that knows how to handle booleans, ints, and floats? Just like SKOutput does, with SKOutputBool, SKOutputNumber, and SKOutputInt? Just a thought.

JohnySeven:
> @JohnySeven - without seeing your code for `Threshold` and `NumericThreshold` I can't really comment, but if it has to work the way you've done it in the example, that's different from how other examples work. To make it like all the other examples (all of them written and modified since `connectTo` was implemented, at least), would look like this:
> 
> ```
> auto * listener = new SKValueListener<float>(sk_path);
> 
> listener->connectTo(new NumericThreshold(0.0f, 100.0f, true, false, config_path)
>             ->connectTo(new DigitalOutput(5));
> 
>   // Start the SensESP application running
>   sensesp_app->enable();
> ```
> 
> In other words, you would not create the NumericThreshold with all the parameters, outside of the string of `connectTo`'s. If you can make it work like I've shown here, then I don't have a real objection to it. And in fact, if I think of `DigitalOutput` as being like `SKOutput`, I guess it's consistent - the final output (to a digital pin, or to the SK Server) is handled by a special transport that does nothing else.

I see and agree, below is updated sample that follows your syntax notes. I've done some testing today and the code works just fine. I have ESP8266 that is pushing lux from light sensor, in daylight relay on my desk is open, when I cover light sensor it gets closed after about 1-2 seconds. So it works!! :-) One ESP is changing state of other ESP via Signal K. One think I need to fix is configuration in web page - something is wrong with my scheme. I'll commit my changes ASAP.

```cpp
#include <Arduino.h>

#include "sensesp_app.h"
#include "sensors/analog_input.h"
#include "transforms/linear.h"
#include "signalk/signalk_output.h"
#include "signalk/signalk_listener.h"
#include "signalk/signalk_value_listener.h"
#include "sensors/digital_output.h"
#include "transforms/threshold.h"

// SensESP builds upon the ReactESP framework. Every ReactESP application
// defines an "app" object vs defining a "main()" method.
ReactESP app([] () {

  // Some initialization boilerplate when in debug mode...
  #ifndef SERIAL_DEBUG_DISABLED
  Serial.begin(115200);

  // A small arbitrary delay is required to let the
  // serial port catch up
  delay(100);
  Debug.setSerialEnabled(true);
  #endif


  // Create the global SensESPApp() object.
  sensesp_app = new SensESPApp();
  // The "SignalK path" identifies subscription of data from the SignalK network.
  const char* sk_path = "environment.outside.illuminance";
  // The "Configuration path" is combined with "/config" to formulate a URL
  // used by the RESTful API for retrieving or setting configuration data.
  // It is ALSO used to specify a path to the SPIFFS file system
  // where configuration data is saved on the MCU board.  It should
  // ALWAYS start with a forward slash if specified.  If left blank,
  // that indicates this sensor or transform does not have any
  // configuration to save.
  // Note that if you want to be able to change the sk_path at runtime,
  // you will need to specify a config_path.
  const char* config_path = "/threshold/lights";
  // Wire up the output of the float value on server "sensors.outdoor.illumination" to the transform,
  // and then output the results on the SignalK network...
  auto * listener = new SKValueListener<float>(sk_path);
  listener->connectTo(new NumericThreshold(0.0f, 100.0f, true, false, config_path), 0)
          ->connectTo(new DigitalOutput(5));
  // Start the SensESP application running
  sensesp_app->enable();
});
```

ba58smith:
@JohnySeven - if your configurable items aren't showing up the way you expect, read the Wiki and scroll down to the section on Configuration Paths. https://github.com/SignalK/SensESP/wiki/SensESP-Wiki

But looking at your example, I don't see anything wrong. If you can't sort it out, post your threshold.cpp file here and I'll have a look at it.

JohnySeven:
I've pushed all my changes up. Now you can see all of them at once. I've managed to fix configuration schema for **NumericThreshold** and **IntegerThreshold**. Done some testing it works for me. I've incorporated all sugested changes and I'm ready to finish this :-) Please let me know if you have any comments.

Thanks!

JohnySeven:
Wemos D1 with relay

This is my testing device, in case anyone would like to follow :-)

examples/relay_control.cpp
```diff
@@ -0,0 +1,52 @@
+#include <Arduino.h>
+
+#include "sensesp_app.h"
+#include "sensors/analog_input.h"
+#include "transforms/linear.h"
```
ba58smith, line 5: Since you're not using a Linear transport, do you need this `#include`?

src/sensors/digital_output.cpp
```diff
@@ -0,0 +1,13 @@
+#include "digital_output.h"
+#include "Arduino.h"
+
+DigitalOutput::DigitalOutput(int pin)
+{
+    this->pinNumber = pin;
```
ba58smith, line 6: Since there's no abiguity about which pinNumber you're referring to, I don't think you need `this->`. Just `pinNumber = pin` is sufficient. It's OK if you include `this->`, of course. Just not necessary. (This was a comment @mairas made on some of my code when I first started working on this, so I thought I'd pass it along.)

src/sensors/digital_output.cpp
```diff
@@ -0,0 +1,13 @@
...
+    pinMode(pin, OUTPUT);
+}
+
+void DigitalOutput::set_input(bool newValue, uint8_t inputChannel)
+{
+    digitalWrite(this->pinNumber, newValue);
```
ba58smith, line 12: Same comment as above.

src/signalk/signalk_value_listener.h
```diff
@@ -0,0 +1,32 @@
+#ifndef _sk_value_listener_H
+#define _sk_value_listener_H
+
+#include "signalk_listener.h"
+#include "system/observable.h"
+#include "system/valueproducer.h"
+#include <ArduinoJson.h>
+///////////////////
+// SKListener is object that listens for specific value in Signal K, period is interval between updates in ms
```
ba58smith, line 9: You changed `period` to `listen_Delay` in the code. (Thank you!) You should change it in the comment, too.

src/net/ws_client.cpp
```diff
@@ -84,20 +87,86 @@ void WSClient::on_error() {
...
+
+    for (size_t i = 0; i < listeners.size(); i++) {
+      auto* listener = listeners.at(i);
+      String sk_path = listener->get_sk_path();
+      int listenDelay = listener->getListenDelay();
+
+      JsonObject& subscribePath = subscribe.createNestedObject();
+
+      subscribePath["path"] = sk_path;
+      subscribePath["period"] = listenDelay;
```
ba58smith, line 115: For clarity, should `"period"` be changed to `"listen_delay"`? `period` isn't used anywhere else in your code now.

src/signalk/signalk_listener.cpp
```diff
@@ -0,0 +1,8 @@
+#include "signalk_listener.h"
+
+std::vector<SKListener*> SKListener::listeners;
+
+SKListener::SKListener(String sk_path, int listen_Delay) : sk_path{sk_path} {
+  listeners.push_back(this);
+  this->listen_delay = listen_Delay;
```
ba58smith, line 7: Same comment as above about `this->listen_delay`.

src/signalk/signalk_listener.h
```diff
@@ -0,0 +1,59 @@
...
+         * this particular subscription to value
+         */
+        SKListener(String sk_path, int listen_delay);
+
+        /**
+         * Returns the current SignalK path.  An empty string
+         * is returned if this particular source is not configured
+         * or intended to return actual data.
+         */
+        String& get_sk_path() {
```
ba58smith, line 33: Will this return an empty string if it's not configured? I don't see where `sk_path` is set to an empty string anywhere as the default value. Should line 52 below be `String sk_path = ""`?

src/signalk/signalk_listener.h
```diff
@@ -0,0 +1,59 @@
...
+ */
+class SKListener : virtual public Observable {
+
+    public:
+        /**
+         * The constructor
+         * @param sk_path The SignalK path that identifies
+         * this particular subscription to value
+         */
+        SKListener(String sk_path, int listen_delay);
```
ba58smith, line 26: Should you add to the comment a description of the listen_delay parameter? It may seem obvious, but maybe not?

src/signalk/signalk_value_listener.h
```diff
@@ -0,0 +1,32 @@
...
+///////////////////
+template <class T>
+class SKValueListener : public SKListener, public ValueProducer<T> {
+ public:
+  SKValueListener(String sk_path, int listen_Delay = 1000) : SKListener(sk_path, listen_Delay)
```
ba58smith, line 14: Should the default value for sk_path = ""?

src/signalk/signalk_value_listener.h
```diff
@@ -0,0 +1,32 @@
...
+  SKValueListener(String sk_path, int listen_Delay = 1000) : SKListener(sk_path, listen_Delay) 
+  {
+       
+  }
+
+  void parseValue(JsonObject& json) override
+  {
+       this->output = (T)json["value"];
+       this->notify();
+  }
```
ba58smith, line 23: Same comment as above about `this->` for output and notify(). I don't think it's necessary, is it? (But maybe it is in this case?)

examples/relay_control.cpp
JohnySeven: You're right, not necessary. Removed.

src/net/ws_client.cpp
JohnySeven: If you checkout Signal K WS specification, there is expected **period** in subscriptions, more information and samples are here: https://signalk.org/specification/1.4.0/doc/subscription_protocol.html below "To subscribe to the required criteria send a suitable subscribe message:"

src/sensors/digital_output.cpp
JohnySeven: Fixed, thanks :-)

src/sensors/digital_output.cpp
JohnySeven: Fixed, thanks :-)

src/signalk/signalk_listener.cpp
JohnySeven: Nice, thanks :-)

src/signalk/signalk_listener.h
JohnySeven: Done

src/signalk/signalk_listener.h
JohnySeven: Empty string would be issue, because it defines value we are subscribing to in SignalK - also this class is just "abstract" base class. From my point of you it isn't useful to have predefined values in here. Predefined values are in inherited class **SKValueListener**. 

What I need to think about is when user uses empty string in **sk_path** parameter! Currently I think we should not allow that.

src/signalk/signalk_value_listener.h
JohnySeven: Done, thanks 👍

src/signalk/signalk_value_listener.h
JohnySeven: check my comment regarding **get_sk_path**, pls.

src/signalk/signalk_value_listener.h
JohnySeven: Removed `this->` and I agree.

JohnySeven:
@ba58smith I've commited most of the suggested changes - one think I need to brainstorm is regarding case when argument `sk_path` in `SKListener(String sk_path, int listen_delay)` is null or empty string. As it is now it will cause subscription failure. So from my point of view we should throw exception or something, what do you think?

src/transforms/threshold.h
```diff
@@ -0,0 +1,50 @@
...
+ public:
+  ThresholdTransform(C minValue, C maxValue, P inRange, P outRange,
+                     String config_path="") : Transform<C, P>(config_path) {
+    this->minValue = minValue;
+    this->maxValue = maxValue;
+    this->inRange = inRange;
+    this->outRange = outRange;
+    Enable::className = "ThresholdTransform";
+    this->load_configuration();
+  };
```
ba58smith, line 15: In keeping with all the other sensor and transform constructor declarations, why not make this one like this, and move it to the .cpp file?
```
ThresholdTransform(C minValue, C maxValue, P inRange, P outRange,
                     String config_path="") : Transform<C, P>(config_path), minValue{minValue},
                     maxValue{maxValue}, inRange{inRange}, outRange{outRange}  {
    className = "ThresholdTransform";
    load_configuration();
  };
```

src/transforms/threshold.cpp
```diff
@@ -0,0 +1,106 @@
+#include "threshold.h"
+
+template <class C, class P>
+void ThresholdTransform<C,P>::set_input(C input, uint8_t inputChannel) {
+  
+  if(input >= this->minValue && input <= this->maxValue)
```
ba58smith, line 6: Lots of unnecessary `this->` in this file.

src/transforms/threshold.h
```diff
@@ -0,0 +1,50 @@
+#include "transform.h"
+#ifndef _threshold_h
+#define _threshold_h
+template <typename C, typename P>
+class ThresholdTransform : public Transform<C, P> {
+ public:
+  ThresholdTransform(C minValue, C maxValue, P inRange, P outRange,
```
ba58smith, line 7: Since SensESP is a library, the end user should be able to know how to use it based only on what's in the .h file. So please add whatever comments are necessary for the classes and methods in here so that an end user will know how to use them w/o looking at the .cpp file.

ba58smith, line 7: Since inRange will typically be set to `true` (because you're looking for a value within a range, not outside of a range), you could give inRange a default value = true. That way, if the end user doesn't need to provide a config_path, they also won't need to provide a value for inRange, so their constructor will be very simple: `new NumericThreshold(50, 100);`. I know, it's not critical, but it's a nice little touch, I think. (As someone who has been the end user way more often than the library developer, I certainly appreciate things like that.)

src/transforms/threshold.cpp
```diff
@@ -0,0 +1,106 @@
...
+
+static const char NUMERIC_SCHEMA[] PROGMEM = R"({
+    "type": "object",
+    "properties": {
+        "min": { "title": "Minimal value", "type": "number" },
+        "max": { "title": "Maximal value", "type": "number" },
+        "in-range": { "title": "In range value", "type": "boolean" },
+        "value": { "title": "Last value", "type": "boolean" }
+    }
+  })";
```
ba58smith, line 43: This is a great place to help the end user know how to use this - by providing a description. Here's an example from another transform:
`"min_angle": { "title": "Minimum angle value", "description": "If you have output between -pi and pi, use -3.14159265, otherwise use 0.", "type": "number" },`.

"value" represents the most recent value of `output`, and it's supposed to be read-only. See any other sensor or transform. 
`"value": { "title": "Last value", "type" : "number", "readOnly": true }`

src/transforms/threshold.cpp
```diff
@@ -0,0 +1,106 @@
...
+JsonObject& NumericThreshold::get_configuration(JsonBuffer& buf)
+{
+    JsonObject& root = buf.createObject();
+    root["min"] = this->minValue;
+    root["max"] = this->maxValue;
+    root["in-range"] = this->inRange;
+    root["value"] = this->output;
+    return root;
+}
+
```
ba58smith, line 34: For consistency with the rest of SensESP, please change your `in-range` to `in_range`.

src/transforms/threshold.cpp
```diff
@@ -0,0 +1,106 @@
...
+      this->output = this->outRange;
+
+      debugI("Threshold is out of range.");
+  }
+
+  debugI("Threshold updated.");
+  
+  this->notify();
+}
+
```
ba58smith, line 23: I think you mean "Value is in range" (or out of range). No reason for the "Threshold updated" - no other sensor or transport has that. For that matter, no other sensor or transport has ANY debug messages when `output` is set. I don't suppose it hurts anything, but it's really not necessary, especially after you've determined that it works the way you intend.

src/transforms/threshold.cpp
```diff
@@ -0,0 +1,106 @@
...
+  static const char INTEGER_SCHEMA[] PROGMEM = R"({
+      "type": "object",
+      "properties": {
+          "min": { "title": "Minimal value", "type": "number" },
+          "max": { "title": "Maximal value", "type": "number" },
+          "in-range": { "title": "In range value", "type": "boolean" },
+          "value": { "title": "Last value", "type": "boolean" }
+      }
+    })";
+
```
ba58smith, line 86: Same comments as those for NumericThreshold.

src/net/ws_client.cpp
ba58smith: Ah! I didn't know that!

src/signalk/signalk_listener.h
ba58smith: I agree - an empty sk_path doesn't make sense. I made my comment here based on the comment in the code, saying that an empty string is returned if this source isn't configured. Which it would, but that would not be OK. Maybe add a debug message to the constructor of SKValueListener if no sk_path is provided?

ba58smith:
> @ba58smith I've commited most of the suggested changes - one think I need to brainstorm is regarding case when argument `sk_path` in `SKListener(String sk_path, int listen_delay)` is null or empty string. As it is now it will cause subscription failure. So from my point of view we should throw exception or something, what do you think?

I agree. `main.cpp` shouldn't even compile if no sk_path is provided (which will happen as long as there is no default value for sk_path in the constructor). I don't think you can make it not compile if the sk_path is invalid (something like "/environment/outside/illuminance" instead of "environment.outside.illuminance"), and it would be a lot of work to make sure it's valid. I think it's probably OK to make it very clear in the comments to the constructor in the .h file that a valid sk_path must be provided, and it must be in the form "<regExp>.<regExp>.<optionalRegExp>.<etc>". Also, make sure the comment in the example makes it clear what they have to provide. Since most people will be looking at examples to make their own main.cpp, as long as the example spells out what they have to do, I think it'll be OK.

src/transforms/threshold.h
JohnySeven: I had problem with linker, moving it to declarations fixed it somehow.

ba58smith: @JohnySeven - Ah - I think I know what the problem is. When I added the ads1x15.h/.cpp, it was the first time I used a class template. It would not link, because the linker doesn't know what the possible instances of the template class will be. So your choices are either to do what you did (move the implementation into the .h file), or do what I did in ads1x15.cpp:
```
// define all possible instances of an ADS1x15
template class ADS1x15<Adafruit_ADS1015>;
template class ADS1x15<Adafruit_ADS1115>;
```
You'll see that also farther down in that same file, and you'll see it also at the end of /src/system/valueconsumer.h and /src/system/valueproducer.h. And you did it in signalk_value_listener.h. Obviously, in thoses cases, those typedefs are in the .h file, and so is the implementation of the constructors.

Just leave it as it is. It's working, and that'w what's most important.

ba58smith:
@JohnySeven - I hope you're going to get back on this. It's a great enhancement!

JohnySeven:
@ba58smith Yeah, been little busy last few days, but I'm up to finish it :-) I don't leave things not done.

src/signalk/signalk_listener.h
JohnySeven: Good, added following message "SKValueListener: User has provided empty sk_path. The subscription will not work as expected!" if string is empty.

src/transforms/threshold.cpp
JohnySeven: outRange removed

src/transforms/threshold.h
JohnySeven: Comments added, I hope it's enough :-)

JohnySeven: Done it in NumericThreshold and IntegerThreshold.

src/transforms/threshold.cpp
JohnySeven: Done 👍

src/transforms/threshold.cpp
JohnySeven: Done

src/transforms/threshold.cpp
JohnySeven: Removed debug messages - used them when debugging. You're right, now it's useless. Thanks

src/transforms/threshold.cpp
JohnySeven: Done

JohnySeven:
I've committed all my changes. @ba58smith what do you think now? 🙂

ba58smith:
I should be able to review all the new commits, and will re-review everything, within the next 24 hours. Thanks!

src/signalk/signalk_value_listener.h
ba58smith: I don't know if `DEBUG_E` is the same as `debugE`, but all throughout SensESP, the latter form is used. (debugE, debugI, etc.)

src/signalk/signalk_value_listener.h
ba58smith: It looks like you added back the `this->`: (`this->output`).

src/transforms/threshold.cpp
ba58smith: It looks like you added back `this->` to `output` and `notify`. I thought you were going to remove them all.

examples/relay_control.cpp
```diff
@@ -0,0 +1,50 @@
...
+  // ALWAYS start with a forward slash if specified.  If left blank,
+  // that indicates this sensor or transform does not have any
+  // configuration to save.
+  // Note that if you want to be able to change the sk_path at runtime,
+  // you will need to specify a config_path.
+  const char* config_path = "/threshold/lights";
+  // Wire up the output of the float value on server "environment.outside.illuminance" to the NumericThreshold,
+  // and then output the transformed float to boolean to DigitalOutput
+  auto * listener = new SKOutputNumber(sk_path);
+  listener->connectTo(new NumericThreshold(0.0f, 100.0f, true, config_path), 0)
```
ba58smith, line 46: Shouldn't this be `auto * listener = new SKListener(sk_path);`?

src/sensors/digital_output.cpp
```diff
@@ -0,0 +1,13 @@
+#include "digital_output.h"
+#include "Arduino.h"
+
+DigitalOutput::DigitalOutput(int pin)
+{
+    pinNumber = pin;
+    pinMode(pin, OUTPUT);
+}
+
```
ba58smith, line 9: I believe the constructor should call its base class constructor. Also, the assignment of pinNumber can be done a little cleaner. (This comment is based on so many other constructors I've looked at throughout SensESP, such as the ones in digital_input.cpp, for reference.)
```
DigitalOutput::DigitalOutput(int pin) : BooleanConsumer(), pinNumber{pin}
{
    pinMode(pin, OUTPUT);
}

src/transforms/threshold.cpp
ba58smith: @JohnySeven, I don't know what happened, but now a lot of those unnecessary `this->` are back in your files, and so is `outRange`. Please re-do the fixes you did to remove them the first time. (Probably a git thing - it's really easy to get commits mixed up!)

src/transforms/threshold.cpp
ba58smith: "value" is still not "readOnly".

src/transforms/threshold.cpp
ba58smith: "value" is still not readOnly.

ba58smith:
I had to UnResolve some of the items, because it looks like your changes were un-done. (Probably a git thing.) Please look at all Unresolved comments. Sorry!

examples/relay_control.cpp
ba58smith: Actually, it should be SKNumericListener (was SKFloatListener, but I'm changing it).

src/signalk/signalk_value_listener.h
ba58smith: Oh, I guess it has to be there.

ba58smith:
I'm going to merge this, as it works the way it is. But I'm going to make some "cleanup" changes, in a new PR.

JohnySeven:
Hi Brian,
thanks for finishing this up. I'm little busy right now with other stuff.

I have few other enhancements in my mind. I hope to add more PR in future.

Thanks for cooperation and for you great work for community!
Jan

PR #110, opened 2020-04-15
Add Adafruit INA219 Library

plerenius:
The Adafruit INA219 Library was missing from the dependency list causing the examples to compile with an error. This fixed it for me. I just started learning about SensESP, so I'm not sure it's the only place to edit...


ba58smith:
@plerenius , thanks for this. In my installation, I have to add new libraries to the platformio.ini that's in my project folder in order for it to work. I'm not sure why, but because I did that, and it worked, I forgot to add it to the SensESP's platformio.ini. I'll merge this in the next time I merge PR's.

PR #113, opened 2020-04-26
Fix missing library dependencies.

gzahl:
I tried building the ina219 example, but some library dependencies were missing. That is fixed by this pull request.
Btw. thanks for this great library!


ba58smith:
@gzahl - Well, apparently that IS how you add a library to a PlatformIO project, if it's a library. I just learned this today. My apologies. I will reopen this, and fix it, along with several other libraries that need the same treatment.

PR #115, opened 2020-04-28
add support for ESP32

loopsmark:
Added support to the esp32 platform and the board az-delivery-devkit-v4

PR #120, opened 2020-05-04
add transforms for counter and timer with an example

loopsmark:


examples/bilge_counter_and_timer.cpp
```diff
@@ -0,0 +1,101 @@
...
+  // run-time configuration.
+  const char* counter_engine_config_path = "/bilges/engine_counter";
+  const char* timer_engine_config_path = "/bilges/engine_timer";
+
+  const char* counter_cabin_config_path = "/bilges/cabin_counter";
+  const char* timer_cabin_config_path = "/bilges/cabin_timer";
+    
+  const char* counter_midship_config_path = "/bilges/midship_counter";
+  const char* timer_midship_config_path = "/bilges/midship_counter";
+
```
ba58smith, line 56: You've used "/bilges/midship_counter" twice (instead of "/bilges/midship_timer").

This arrangement of config paths will result in a single "main menu item" called "bilges", with 6 sub-items. If that's what you want, great.
bilges > engine_counter
          > engine_timer
          > cabin_counter
          > cabin_timer
         > midship_counter
         > midship_timer

If you'd rather have the menu look like this:
bilges > engine > counter
                            timer
bilges > cabin > counter
                          timer
bilges > midship > counter
                              timer

replace your underscores with with forward slashes, like "/bilges/engine/counter".

The behavior will be about the same, because after you save a configuration change, it goes back to the same menu. Actually, your arrangement will save you a click each time.

examples/bilge_counter_and_timer.cpp
```diff
@@ -0,0 +1,101 @@
...
+  uint8_t cabinPin = 5;
+  uint8_t midshipPin = 13;
+
+  // Constant for the Timer and Counter for when it will reset in milliseconds
+  const unsigned long oneDay = 1440 * 60 * 1000UL;
+
+  auto* pEngineInput = new DigitalInputValue(enginePin, INPUT_PULLDOWN, RISING);  
+  auto* pCabinInput = new DigitalInputValue(cabinPin, INPUT_PULLUP, RISING);  
+  auto* pMidshipInput = new DigitalInputValue(midshipPin, INPUT_PULLUP, RISING);  
+
```
ba58smith, line 71: I don't think INPUT_PULLDOWN works on some of the ESP's. This won't hurt to have it in here, but you might need to have a physical 10K pulldown resistor in your circuit for it to work properly.

ba58smith, line 71: DigitalInputValue currently outputs a boolean - it's intended to output the current value of the pin - true or false (1 or 0). I recently tried to use it for my bilge pump indicator (is it running or not?), and it didn't work. I changed it to inherit from IntegerProducer instead of BooleanProducer, and changed `::enable` to just this:
```
void DigitalInputValue::enable() {
  app.onRepeat(read_delay, [this](){
    output = digitalRead(pin);
        notify();
  });
```
It works great (and I will make a PR for it soon) for my bilge pump status indicator. But will it work in your situation, as it is now? (Outputting only booleans.)

src/transforms/counter.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+  static const char SCHEMA[] PROGMEM = R"###({
+    "type": "object",
+    "properties": {
+        "resetPeriod": { "title": "Reset perid", "type": "number", "description": "The time, in milliseconds, between each reset cycle" },
+        "defaultState": { "title": "Default state", "type" : "boolean", "description": "Default state of input" },
+        "count": { "title": "Count", "type" : "number" }
+    }
+  })###";
+
```
ba58smith, line 48: Please add `"value": { "title": "Last value", "type" : "number", "readOnly": true }` after "count" in the "properties" section. (All configurables have this - its very helpful.)

src/transforms/counter.cpp
```diff
@@ -0,0 +1,65 @@
...
+  if (defaultState == input && true == last_value) {
+    count++; 
+    last_value = false;
+  } else {
+      last_value = true;
+  }
+  output = count;
+  notify();
+}
+
```
ba58smith, line 30: I worked through this logic for a bilge pump, whose defaultState should be false. I don't think it works, or maybe it does, but it doesn't count the first time it comes on. Please re-check the logic - it might be OK, and I'm just not understanding it correctly.

examples/bilge_counter_and_timer.cpp
ba58smith: Also - if you use `RISING`, the interrupt will fire only when the pin goes from LOW to HIGH. But I think your code in `Counter::set_input` assumes you're being told about ALL changes (LOW to HIGH, and HIGH to LOW). If that's what you want, then you need to change `RISING` to `CHANGE`. However... I wonder if you can simplify your logic in `Counter::set_input` by using the value of `defaultState` to determine whether you should use `RISING` or `FALLING`. Here's what I mean:
Let's say I'm monitoring my bilge pump, and I want to know how many times it turns on. The `defaultState` would be `false` (OFF), and the interrupt type would be `RISING`, because I want to count the times it turns on. (I don't need to count the times it turns off.) Or if you wanted to monitor something whose `defaultState` is `true` (ON), and you want to know how many times it turns off, the interrupt type would be `FALLING`. Right? Or maybe it's not that simple?

src/transforms/timer.cpp
```diff
@@ -0,0 +1,78 @@
...
+
+  static const char SCHEMA[] PROGMEM = R"###({
+    "type": "object",
+    "properties": {
+        "resetPeriod": { "title": "Reset perid", "type": "number", "description": "The time, in milliseconds, between each reset cycle" },
+        "defaultState": { "title": "Default state", "type" : "boolean", "description": "Default state of input" },
+        "time": { "title": "Time", "type" : "number" }
+    }
+  })###";
+
```
ba58smith, line 61: Please add `"value": { "title": "Last value", "type" : "number", "readOnly": true }` as one of the "properties" (as all other Configurables have it.)

loopsmark:
It's hard for me to continue this development without having support for esp32 in the master branch. Do you know when PR #118 will be merged?

ba58smith:
> It's hard for me to continue this development without having support for esp32 in the master branch. Do you know when PR #118 will be merged?

Sorry - I've been at sea for 72 hours, just now seeing this.  I should be able to provide a better update later today.

ba58smith:
@loopsmark - Since PR #118 is so close to your PR #115, maybe you can help troubleshoot the problem he's having. Try it on your ESP32, running his example. The more people who try it, the more likely we are to find a solution quickly. Thanks!

ba58smith:
@loopsmark - I see that your ESP32 PR was very, very close to being ready. Rather than wait for PR #118, you can reopen #115, and address the few remaining comments I had on it, and then we can merge that one. I don't care which one we go with, as long as all the discrepancies between the two have been addressed by one of them.

ba58smith:
@loopsmark - PR#118 has been merged (ESP32 support). Hopefully you can now get back on this one. Sorry for the delay!

examples/bilge_counter_and_timer.cpp
```diff
@@ -0,0 +1,101 @@
...
+  // Some initialization boilerplate when in debug mode...
+  #ifndef SERIAL_DEBUG_DISABLED
+  Serial.begin(115200);
+
+  // A small arbitrary delay is required to let the
+  // serial port catch up
+  delay(100);
+  Debug.setSerialEnabled(true);
+  #endif
```
tedenda, line 22: This debug code has been simplified to this:
```
// Start serial debug if not disabled
#ifndef SERIAL_DEBUG_DISABLED
  SetupSerialDebug(115200);
#endif
```

tedenda (commented):
Only a smal change

tedenda:
@loopsmark
A lot has been fixed / added in SensESP since May when you opened this PR.
What is your situation now, do you have some time to spare to complete this PR?
Please let me know if you need some help with testing it out before it's merged!

src/transforms/counter.h
```diff
@@ -0,0 +1,28 @@
...
+  virtual void set_input(bool input, uint8_t inputChannel = 0) override;
+  virtual JsonObject& get_configuration(JsonBuffer& buf) override;
+  virtual bool set_configuration(const JsonObject& config) override;
+  virtual String get_config_schema() override;
+
+ private:
+  uint8_t count;
+  bool last_value;
+  static long unsigned int lastSampleTime;
+  unsigned long resetPeriod;
```
ba58smith, line 24: There is now a LambdaConsumer that can be used to call an arbitrary function. Although it could be used for any function, it was originally created to call a `reset()` method in the Integrator Transform (which is a form of counter). So before this PR is merged, the "reset after a period of time" function of Counter should probably be changed to use that method for doing a reset instead, or at least "enhanced" to add a `reset()` method that could be called by LambdaConsumer, so that the reset can occur when a physical button is pushed.

mairas:
I'll close this PR because there seems to be no activity on it. Please make a new one if you pick it up at some point.

Issue #114, opened 2020-04-27
DigitalInputValue doesn't work


ba58smith:
I connected a bilge pump wire to the D1 pin (stepped down to 2.5v) and confirmed that when the pump is on, the pin has 2.5v and when the pump is off the pin has 0v. I added the debug messages to show if the interrupt is being triggered on any CHANGE, and it is, and the value of `output` is what it's supposed to be: 0 or 1. But no delta ever showed up in the serial monitor, even though `uptime` did. So the `app.onTick` wasn't working. It looks correct, but it's not ever reporting the value of `output` as a delta.

Another problem (I think) is that Instrument Panel doesn't know what to do with a Boolean value - it doesn't have a dedicated gauge, so it just shows the words "false" or "true" when you select Digital Gauge, and doesn't show any difference (the needle never moves) if you select Analog Gauge.

I changed the base class from BooleanProducer to IntegerProducer, and then it would output a 0 or 1 as expected... almost. It would often leave a 1 showing in Instrument Panel after the pump was off. (Yes, I have a pulldown resistor, and as noted above, the value on the pin is 0v when the pump is off.) So I threw out the whole `app.onInterrupt` and `app.onTick` approach (commented out below) and changed it to `app.onRepeat` and report the value every 1000ms, and now it works as expected.

A side benefit of this approach is that in Grafana, there is data to be graphed every second, so it's easy to see the history of the pump's on/off cycle, which you don't get if it only ever reports a change in the state.

I don't want to make this change until someone who understands what might be the problem with `app.onTick` has a look at the problem. I'll just use this simpler approach in my boat to test it for the time being.

```
void DigitalInputValue::enable() {
 app.onRepeat(1000, [this](){
    output = digitalRead(pin);
        notify();
 //app.onInterrupt(
  //  pin, interrupt_type,
  //  [this](){
  //    debugI("Interrupt triggered");
  //    output = digitalRead(pin);
  //    debugI("output is now %d", output);
  //    triggered = true;
  //    //notify();
  //  });
// });
  //app.onTick(
  //  [this](){
  //    if (triggered) {
  //      triggered = false;
  //      notify();
  //    }
  //  }
  //);  
}
```

ba58smith:
See PR #125 .

PR #125, opened 2020-05-13
Fix digital input value

ba58smith:
Changed it from a BooleanProducer to an IntegerProducer and removed the interrupt to watch for changes. Now it simply reports the value of the digital pin every `read_delay` ms.


ba58smith:
Addresses Issue #114 .

PR #128, opened 2020-05-25
Show IP addr of Device in Serial Monitor

ba58smith:
Right after connecting to wifi, show the IP address. Needed to configure the ESP.


ba58smith:
Resolves Issue #128.

PR #132, opened 2020-06-04
Add example for BMP280 Sensor

free-x:
Hi,

I have build succesfully my first ESP32 with BMP280.
Here is example, how to use this sensor with SensESP

Fair wind
free-x

### DIFF
--- a/examples/bmp280_example.cpp
+++ b/examples/bmp280_example.cpp
@@ -30,10 +30,9 @@ ReactESP app([] () {
   // 0x77 is the default address. Some chips use 0x76, which is shown here.
   auto* pBMP280 = new BMP280(0x76);
 
-  // If you want to change any of the settings that are set by Adafruit_BME280::setSampling(), do
+  // If you want to change any of the settings that are set by Adafruit_BMP280::setSampling(), do
   // that here, like this:
   // pBMP280->pAdafruitBMP280->setSampling(); // pass in the parameters you want
-
 
   // Define the read_delays you're going to use:
   const uint read_delay = 1000; // once per second
--- a/examples/bmp280_example.cpp
+++ b/examples/bmp280_example.cpp
@@ -1,0 +1,57 @@
+// BMP280_example.cpp
+
+#include <Arduino.h>
+
+//#define SERIAL_DEBUG_DISABLED
+
+#define USE_LIB_WEBSOCKET true
+
+#include "sensesp_app.h"
+#include "signalk/signalk_output.h"
+#include "sensors/bmp280.h"
+
+ReactESP app([] () {
+  #ifndef SERIAL_DEBUG_DISABLED
+  Serial.begin(115200);
+
+  // A small delay and one debugI() are required so that
+  // the serial output displays everything
+  delay(100);
+  Debug.setSerialEnabled(true);
+  #endif
+  delay(100);
+  debugI("Serial debug enabled");
+
+  // Create the SensESPApp with whatever "standard sensors" you want: noStdSensors, allStdSensors, or uptimeOnly.
+  // The default is allStdSensors.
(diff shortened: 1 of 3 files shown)


examples/bmp280_example.cpp
```diff
@@ -0,0 +1,57 @@
...
+
+  // Create a BMP280, which represents the physical sensor.
+  // 0x77 is the default address. Some chips use 0x76, which is shown here.
+  auto* pBMP280 = new BMP280(0x76);
+
+  // If you want to change any of the settings that are set by Adafruit_BME280::setSampling(), do
+  // that here, like this:
+  // pBMP280->pAdafruitBMP280->setSampling(); // pass in the parameters you want
+
+
```
ba58smith, line 37: Please remove the three lines above (plus all the extra blank lines), as the example is for a BMP280, and those lines are for the BME280.

examples/bmp280_example.cpp
```diff
@@ -0,0 +1,57 @@
...
+  // Define the read_delays you're going to use:
+  const uint read_delay = 1000; // once per second
+  const uint pressure_read_delay = 60000; // once per minute
+
+  // Create a BMP280value, which is used to read a specific value from the BMP280, and send its output
+  // to SignalK as a number (float). This one is for the temperature reading.
+  auto* pBMPtemperature = new BMP280value(pBMP280, temperature, read_delay, "Outside/Temperature");
+
```
ba58smith, line 45: Configuration paths must begin with a forward slash - so "Outside/Temperature" needs to be "/Outside/Temperature". If you don't begin with a forward slash, you won't be able to configure this item at runtime.

examples/bmp280_example.cpp
```diff
@@ -0,0 +1,57 @@
...
+  // to SignalK as a number (float). This one is for the temperature reading.
+  auto* pBMPtemperature = new BMP280value(pBMP280, temperature, read_delay, "Outside/Temperature");
+      
+      pBMPtemperature->connectTo(new SKOutputNumber("environment.outside.temperature"));
+
+
+  // Do the same for the barometric pressure value. Its read_delay is longer, since barometric pressure can't
+  // change all that quickly. It could be much longer for that reason.
+  auto* pBMPpressure = new BMP280value(pBMP280, pressure,  pressure_read_delay, "Outside/Pressure");
+
```
ba58smith, line 52: Same comment as above: "/Outside/Pressure".

examples/bmp280_example.cpp
free-x: done

examples/bmp280_example.cpp
free-x: done

examples/bmp280_example.cpp
free-x: corrected

examples/bmp280_example.cpp
free-x: done

examples/bmp280_example.cpp
free-x: done

free-x:
have added some leading slashes in BME280/SHT31 examples

examples/bmp280_example.cpp
ba58smith: I don't think you did correct this. The BMP280 doesn't have ::setSampling() as a method, so you need to REMOVE the lines that refer to it.

free-x: https://github.com/adafruit/Adafruit_BMP280_Library/blob/a49542c69d134c180c45622d1c2da7d030fbbe88/Adafruit_BMP280.h#L203

am I wrong?

ba58smith: No, you're not wrong - I am. Sorry!

PR #133, opened 2020-06-07
Add PCB Files

doubledutch2:
Added the files used to create / order PCB's with sample images and other helpful stuff.


ba58smith:
@doubledutch2 - Please add a README file that explains, in as much detail as you can, what this PCB is for, and how to use it. Since these files are all in their own directory, this new README won't conflict with the main one.
Thanks for adding this! I really want to make some PCB's of my existing cobbled-together projects, and you've given me some great ideas.

doubledutch2:
Great point - will do.

doubledutch2:
Added readme file as suggested

doubledutch2:
@ba58smith Added readme file as suggested. Is there anything else I need to do?

pcb/README.md
```diff
@@ -0,0 +1,86 @@
...
+
+  DallasTemperatureSensors* dts = new DallasTemperatureSensors(D1);
+
+Dip Switch:
+- D1 - 5V to Wemos Chip On/Off. Using this allows you measure adjust the voltage of your buck converter. When doing so - note that the Capacitor (will need charging / de-charging for every adjustment made)
+- D2 - Supply 3.3v to VCC on above 3 headers
+- D3 - Supply 5v to VCC on above header
+- D4 - LED on/off
+- D5-D8: "Open/Close" J5-> J2,J3,J4 connections. The through here is that if you have a sensors where VCC, GND, D1 (SCL) and D2 (SCA) don't line up to this order, you close the switch and modify the order by (and wiring) J5 to J3 in a different order
+
```
ba58smith, line 36: I think you mean "the THOUGHT here"...
This is still not exactly clear to me how you use D5 - D8. Maybe you could give an example of how you would set each of these for two different I2C sensors: one whose pins are in this order:
VCC - GND - SCL - SCA
and one whose pins are in some other order.

ba58smith:
@doubledutch2 - Other than the specific change I suggested above, just read the whole document very carefully - there are some typos, I think. (Or have someone else read it - it's often hard to find your own typos.) Nothing that will prevent publication - just little things.

doubledutch2:
@ba58smith - Made suggested changes. Does the SW5-8 example make sense?

tkurki:
Would some screenshots or pics make sense in this context?

doubledutch2:
SensorPinOut

This is what I mean - it's not so complicated. I'm just not great at explaining it it appears :-)

doubledutch2:
@tkurki  / @ba58smith: I've added that image to this folder and update the readme file. Anything else?

tkurki:
Lgtm

PR #138, opened 2020-07-21
Add new transform: Tee class

BjarneBitscrambler:
I've drafted a **new transform, called Tee**. Tee is a numeric passthrough transform that always passes the value unchanged. It also optionally applies a linear transform of the form y = mx + b, and then passes that y value to an external function. This external function can be used, for example, to output y in the form of a frequency-modulated signal on a GPIO pin. The Tee class is so called because it **inserts a tee, or tap, into the chain of value Producers/Consumers**. It won't change the value passing through the chain, but allows one to use the value for some local purpose.

I wrote this because our boat has an analog tachometer that takes a square-wave input and displays a corresponding RPM. I wanted to use this tachometer in addition to publishing the engine speed on the SignalK network. So I connected an ESP32 input to the Alternator's AC output, converting the reading to frequency using the SensESP Tachometer example. Inserting the Tee producer/consumer into the signal chain and pointing it to a function that outputs a square-wave signal whose frequency corresponds to the frequency value, allows the ESP32 to control the analog tachometer display in addition to sending the RPM value to SignalK subscribers.

Another possible use for a tap, other than mirroring a signal out on a GPIO pin, might include directing the values to a logging function - i.e. one that is local to the ESP32, not on the external SignalK network.

There may be easier ways of implementing this solution - I'm not making any claims about the elegance of my code as I am more of a hardware engineer than software. Hopefully some others will find it a useful addition to the library of SignalK functions. I've tested it on the ESP32-WROVER-KIT, and on the ESP32-DEVKIT-C devices. Feel free to critique/improve/correct this submission.

### DIFF
--- a/src/transforms/tee.cpp
+++ b/src/transforms/tee.cpp
@@ -1,0 +1,58 @@
+#include "tee.h"
+
+Tee::Tee( uint16_t isTeeActive, float coefficient_tap, float offset_tap, void (*pFunc_tap)(float x), String config_path ) : 
+          NumericTransform(config_path), 
+          isTeeActive{isTeeActive}, coefficient_tap{coefficient_tap}, offset_tap{offset_tap}  {
+  className = "Tee";
+  load_configuration();
+  tee_to = pFunc_tap;
+}
+
+void Tee::set_input(float input, uint8_t inputChannel) {
+
+        output = input; //pass input value unchanged to next in chain
+         if( isTeeActive )
+        {   //route the data to external function, which can e.g. send it to GPIO, or log it, or...
+            (*tee_to)( (float)input * coefficient_tap + offset_tap); //apply y = mx + b linear equation to input
+        }
+        notify();  
+}
+
+JsonObject& Tee::get_configuration(JsonBuffer& buf) {
+  JsonObject& root = buf.createObject();
+  root["coefficient_tap"] = coefficient_tap;
+  root["offset_tap"] = offset_tap;
+  root["isTeeActive"] = isTeeActive;
+  debugI( "Tee get_configuration: coefficient_tap=%f offset_tap=%f isTeeActive=%d", 
+          coefficient_tap, offset_tap, isTeeActive );
+  return root;
+}
+
+static const char SCHEMA[] PROGMEM = R"###({
+    "type": "object",
+    "properties": {
+       "coefficient_tap": { "title": "Coefficient m in y=mx + b", "description": "converts Alternator signal (Hz) into Tachometer Signal (Hz)", "type": "number" },
+       "offset_tap": { "title": "Offset b in y=mx + b", "type": "number" },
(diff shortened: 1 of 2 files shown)


mairas:
Hi! Thanks for the PR. I'm trying to understand it -- since you can have multiple consumers listening to one producer, wouldn't this be functionally identical to just having two consumers listen to the same producer?

ba58smith:
To illustrate @mairas's point, here is how I use my digital tachometer as two separate Signal K instruments: one that's smoothed over 15 samples (which takes some time to stabilize), and one that has only 2 samples, for a very quick (but not as stable) output:
```
auto* pDigitalIn = new DigitalInputCounter(D5, INPUT_PULLUP, RISING, digital_input_read_delay, "/Tach/read delay");

  pDigitalIn->connectTo(new Frequency(freq_multiplier, config_path_multiplier))  // connect the output of pSensor to the input of Frequency()
        ->connectTo(new MovingAverage(15, 1.0, config_path_moving_avg))        
        ->connectTo(new SKOutputNumber("propulsion.main.revolutions"));   // connect the output of Frequency() to a SignalK Output as a number

  // Create a second instance (of a consumer) to display the immediate output of the tachometer
  pDigitalIn->connectTo(new Frequency(0.25)) // connect the output of pSensor to the input of Frequency()
        ->connectTo(new MovingAverage(2, 1.0))     
        ->connectTo(new SKOutputNumber("propulsion.immed.revolutions"));
```

BjarneBitscrambler:
I think I see what you are suggesting: why didn't I just define a consumer that accomplished what I wanted (generate a square wave on a GPIO), and include that in the chain, like this:
```c
pSensor_tach->connectTo(new Frequency( multiplier_tach, config_path_rpm_calibrate))  // connect output of pSensor to Frequency()
            ->connectTo(new GenSquareWave( ... ))   // connect Frequency() to signal generator for analog tachometer
            ->connectTo(new SKOutputNumber( sk_path_rpm, config_path_rpm_skpath ));   // output as a SignalK number
```
Good question :thinking:  I guess I was stuck in thinking that the Producer/Consumer chain wasn't intended to perform any actions other than numeric manipulation of the value being passed along. It seemed to me that a transform that performed an external action would be a different beast. However, I see now that that is an unnecessary limitation...  

I suppose the only argument in favour of a 'tee' might be that one doesn't want to delay passing the value along in the Producer/Consumer chain, and so interspersing a CPU-intensive operation in the middle would be undesirable.  However, that doesn't apply in my analog tach application, since updating the output freq has proven to be pretty straightforward.

As I said - I'm pretty new to this :grin: so if you can't see a need for this then I'll defer to your advice and we can close this PR.  Thanks for your thoughts!

BjarneBitscrambler:
Thanks, too, @ba58smith, for your example!

ba58smith:
If this solves your tach issue, go ahead and close this PR.

Thanks for participating in the development of SensESP! I hope to see more PR's from you in the future.

PR #140, opened 2020-07-29
fix: change to WIFI_STA mode when connected to known WiFi

KEGustafsson:
This should fix issue #87. When ESP is connecting to known WiFi SSID, then mode is changed to WIFI_STA instead of leaving it on WIFI_AP_STA.

### DIFF
--- a/src/net/networking.cpp
+++ b/src/net/networking.cpp
@@ -45,6 +45,7 @@ void Networking::setup(std::function<void(bool)> connection_cb) {
 }
 
 void Networking::setup_saved_ssid(std::function<void(bool)> connection_cb) {
+  WiFi.mode(WIFI_STA);
   WiFi.begin(ap_ssid.c_str(), ap_password.c_str());
 
   uint32_t timer_start = millis();
@@ -61,7 +62,6 @@ void Networking::setup_saved_ssid(std::function<void(bool)> connection_cb) {
     debugI("Connected to wifi, SSID: %s", WiFi.SSID().c_str());
     connection_cb(true);
   }
-  WiFi.mode(WIFI_STA);
 }
 
 void Networking::setup_wifi_manager(std::function<void(bool)> connection_cb) {
--- a/src/net/networking.cpp
+++ b/src/net/networking.cpp
@@ -61,6 +61,7 @@ void Networking::setup_saved_ssid(std::function<void(bool)> connection_cb) {
     debugI("Connected to wifi, SSID: %s", WiFi.SSID().c_str());
     connection_cb(true);
   }
+  WiFi.mode(WIFI_STA);
 }
 
 void Networking::setup_wifi_manager(std::function<void(bool)> connection_cb) {
--- a/src/net/networking.cpp
+++ b/src/net/networking.cpp
@@ -45,7 +45,6 @@ void Networking::setup(std::function<void(bool)> connection_cb) {
 }
 
 void Networking::setup_saved_ssid(std::function<void(bool)> connection_cb) {
-  WiFi.mode(WIFI_STA);
   WiFi.begin(ap_ssid.c_str(), ap_password.c_str());
 
   uint32_t timer_start = millis();
@@ -61,6 +60,7 @@ void Networking::setup_saved_ssid(std::function<void(bool)> connection_cb) {
   if (WiFi.status() == WL_CONNECTED) {
     debugI("Connected to wifi, SSID: %s", WiFi.SSID().c_str());
     connection_cb(true);
+    WiFi.mode(WIFI_STA);
   }
 }
 


ba58smith:
@KEGustafsson - that seems so simple! I spent hours looking for a fix for this from a WifiManager perspective, and none of the forum threads I found suggested it. I'm going to wait until @mairas has a look at it, and let him merge it. Unless he sees a problem with it, of course. Thanks!

KEGustafsson:
"ap sta ESPAsyncWiFiManager" search resulted a link to [https://github.com/alanswx/ESPAsyncWiFiManager/issues/52](url).
From there I saw this function and started to search SensESP code. I didn't find any and then started to figure out where this could be put for testing and I came up this location.

Missing of this command is explaining very well why AP+STA is ON after captive portal. It is never turned off.

ba58smith:
One thing I just thought about - have you tested what happens after you are connected, and then the wifi goes down? And then comes back up? It probably works, but will you test that, please, if you haven't already?

KEGustafsson:
Issue_2
Blue line is ESP AP, yellow strongest co-channel.

Test procedure:
1) Known WiFi AP goes down (not visible in the graph) and ESP AP comes up.
2) Captive portal accessible from 192.168.4.1.
3) Known WiFi AP up again and ESP connects to that and turn off it's own AP.
4) ESP is accessible to it's configurations via known WiFi AP.

Result: Working as expected.

JohnySeven:
Nice fix! One question I have, why do you place the mode configuration after you are already connected Wifi? Any reason for that? Thanks!

KEGustafsson:
Good question.

If we running this function, void Networking::setup_saved_ssid(), then we are sure that AP where we are trying to connect is known, up & running and we have credentials available?

Then WiFi.mode change could be the first that we should do and ESP is already in STA mode before trying to contact AP. Now it is short period in AP_STA before changing to STA at the end.

I'm not too familiar with the code and I went thru all most all codes to figure out how this is working. Then this place was quite convenient to test the function. Most probably place is not the optimal, as I started to get second thoughts based on you comment.

KEGustafsson:
Tested WiFi.mode change before connecting to AP and that is working as well OK. If PR is acceptable then I would propose this change too.

ba58smith:
@KEGustafsson - are you saying you should MOVE `WiFi.mode(WIFI_STA);` to a different spot, or that it should be where you put it, and also in another location? You can explain it here, or if you're confident in the change, go ahead and make it, and commit it, so that the PR shows it.

KEGustafsson:
I was missing one critical use case from testing. Known AP password is changed and ESP cannot connect to it. Then ESP own AP is still kept on and Captive portal is available for end user. If connection to known AP is successful then ESP own AP is tuned off.

Either Captive portal or connection to known AP are now two possible options. No lock out possible.

ba58smith:
That makes sense to me - change to STA mode only after we know we don't need the AP anymore.

ba58smith:
I will merge this after the new sensesp_app_options PR is merged. Both PR's affect networking.cpp.

mairas:
Looks good to me too! (Sorry for not going through the PRs earlier.)

Not essential, but maybe you could squash the individual commits into a single one before merging?

ba58smith:
@mairas - When I merge this, I will use the GitHub option "Squash and merge". Hopefully that will do what you're requesting. (But I'm going to wait until the SensespAppOptions PR is merged before I merge this one, as they both affect networking.cpp, and I don't want to introduce a merge problem for that PR.)

mairas:
Yeah, that will have the same effect in this case. That's something to be used sparingly, though - often the individual PRs contain very useful information.

As this ended up being a one-liner and the other PR has been dragging on, I'd have just gone forward and merged this. But it's no biggie.

ba58smith:
@mairas - do you mean the individual commits have useful information? I've always used Squash and Merge because, in my case, the existence of multiple commits is a result of me having to fix something from the first commit.

Do you typically just Merge, rather than Squash and Merge?

mairas:
> @mairas - do you mean the individual commits have useful information? I've always used Squash and Merge because, in my case, the existence of multiple commits is a result of me having to fix something from the first commit.
> 
> Do you typically just Merge, rather than Squash and Merge?

Yes, individual commits absolutely contain useful information. Personally, I'd never use Squash and Merge (except in corner cases such as this PR, in which the end result is a one-liner and there's little benefit in having the submitter make the change). Instead, it'd be better to request the submitter to clean up the commits. In professional software development, you are often expected to modify the git commit chain to form a cohesive, progressing story. Like, not changing something in one commit and then reverting the change in the next one. In such cases, the commit history should be locally modified by just removing those commits. Additionally, commits should be atomic, only changing one functionality at a time. I sometimes spend non-trivial amounts of time modifying the commit history before making a PR.

mairas:
Having said all that, in this case I'll just squash and merge. :-)

PR #142, opened 2020-08-05
SensESPAppOptions - allows configuration of SensESPApp application

JohnySeven:
This PR was highly discussed and tested with @mairas and @ba58smith. Thank you guys for your feedback and help!

This PR will allow SensESP lib user configure following features:

- Wifi - preconfiguration of connection settings from main.cpp, or using Wireless manager
- Networking - allows user to enable / disable mDNS, use static server configuration from main.cpp, allows to preconfigure device hostname
- Allows configuring LED blink intervals and / or pin number to use

In future it will allow any developer of the library to configure more parameters of SensESP from main.cpp. This PR adds more flexibility to SensESP library and defines the way how behaviour for SensESP library is changed by end user.

This PR is backwards compatible with previous code, so no breaking changes are introduced for end user if the end user doesn't change or rely on internals SensESP structures.

### DIFF
--- a/examples/relay_control.cpp
+++ b/examples/relay_control.cpp
@@ -8,11 +8,13 @@
 #include "transforms/threshold.h"
 
 
-// Following sample shows how to listen for value on Signal K server.
-// This sample demonstrates how to make automatic navigation light switching device depending on light conditions.
+// This example shows how to listen to a specific SK Path (on the SK Server) for a value within a specified range.
+// If the value is in the range (or outside, if that's how you've set it up), the output will be "true".
+// You can then connect to a DigitalOutput, to turn on an LED, or activate a relay to sound an alarm, or whatever.
+// This example demonstrates how to turn on a navigation light when it gets dark outside.
 
 // SensESP builds upon the ReactESP framework. Every ReactESP application
-// defines an "app" object vs defining a "main()" method.
+// defines an "app" object (vs. defining a "main()" method).
 ReactESP app([] () {
 
   // Some initialization boilerplate when in debug mode...
@@ -28,8 +30,10 @@ ReactESP app([] () {
 
   // Create the global SensESPApp() object.
   sensesp_app = new SensESPApp();
-  // The "SignalK path" identifies subscription of data from the SignalK network.
+  
+  // Define the SK Path you want to listen to
   const char* sk_path = "environment.outside.illuminance";
+
   // The "Configuration path" is combined with "/config" to formulate a URL
   // used by the RESTful API for retrieving or setting configuration data.
   // It is ALSO used to specify a path to the SPIFFS file system
@@ -40,11 +44,13 @@ ReactESP app([] () {
   // Note that if you want to be able to change the sk_path at runtime,
   // you will need to specify a config_path.
   const char* config_path = "/threshold/lights";
+  
   // Wire up the output of the float value on server "environment.outside.illuminance" to the NumericThreshold,
   // and then output the transformed float to boolean to DigitalOutput
-  auto * listener = new SKOutputNumber(sk_path);
-  listener->connectTo(new NumericThreshold(0.0f, 100.0f, true, config_path), 0)
-          ->connectTo(new DigitalOutput(5));
+  auto * pListener = new SKNumericListener(sk_path);
+  pListener->connectTo(new NumericThreshold(0.0f, 100.0f, true, config_path), 0)
+           ->connectTo(new DigitalOutput(5));
+  
   // Start the SensESP application running
   sensesp_app->enable();
 });
--- a/examples/relay_control.cpp
+++ b/examples/relay_control.cpp
@@ -1,6 +1,7 @@
 #include <Arduino.h>
 
 #include "sensesp_app.h"
+#include "sensesp_app_options.h"
 #include "signalk/signalk_output.h"
 #include "signalk/signalk_listener.h"
 #include "signalk/signalk_value_listener.h"
@@ -27,9 +28,11 @@ ReactESP app([] () {
   Debug.setSerialEnabled(true);
   #endif
 
-
   // Create the global SensESPApp() object.
-  sensesp_app = new SensESPApp();
+  sensesp_app = new SensESPApp(new SensESPAppOptions()
+                ->useWifi("ssdi", "password")
+                ->useServer("10.10.10.1", 3000)
+                ->useStandardSensors(allStdSensors));
   
   // Define the SK Path you want to listen to
   const char* sk_path = "environment.outside.illuminance";
--- a/examples/relay_control.cpp
+++ b/examples/relay_control.cpp
@@ -22,11 +22,13 @@ ReactESP app([] () {
   #ifndef SERIAL_DEBUG_DISABLED
   Serial.begin(115200);
 
-  // A small arbitrary delay is required to let the
-  // serial port catch up
+  // A small delay and one debugI() are required so that
+  // the serial output displays everything
   delay(100);
   Debug.setSerialEnabled(true);
   #endif
+  delay(100);
+  debugI("Serial debug enabled");
 
   // Create the global SensESPApp() object.
   sensesp_app = new SensESPApp(new SensESPAppOptions()
@@ -51,7 +53,7 @@ ReactESP app([] () {
   // Wire up the output of the float value on server "environment.outside.illuminance" to the NumericThreshold,
   // and then output the transformed float to boolean to DigitalOutput
   auto * pListener = new SKNumericListener(sk_path);
-  pListener->connectTo(new NumericThreshold(0.0f, 100.0f, true, config_path), 0)
+  pListener->connectTo(new NumericThreshold(0.0f, 100.0f, true, config_path))
            ->connectTo(new DigitalOutput(5));
   
   // Start the SensESP application running
--- a/examples/relay_control.cpp
+++ b/examples/relay_control.cpp
@@ -32,7 +32,7 @@ ReactESP app([] () {
 
   // Create the global SensESPApp() object.
   sensesp_app = new SensESPApp(new SensESPAppOptions()
-                ->useWifi("ssdi", "password")
+                ->useWifi("yourSSID", "yourPassword")
                 ->useServer("10.10.10.1", 3000)
                 ->useStandardSensors(all)
                 ->setLEDPin(10)
--- a/examples/relay_control.cpp
+++ b/examples/relay_control.cpp
@@ -31,9 +31,9 @@ ReactESP app([] () {
   debugI("Serial debug enabled");
 
   // Create the global SensESPApp() object.
-  sensesp_app = new SensESPApp([] (SensESPAppOptions*o)
+  sensesp_app = new SensESPApp([] (SensESPAppOptions*options)
   {
-         o->setServerOptions("10.10.10.1", 3000)
+         options->setServerOptions("10.10.10.1", 3000)
           ->setWifiOptions("yourSSID", "yourPassword")
           ->setHostName("relay")
           ->setStandardSensors();
--- a/examples/relay_control.cpp
+++ b/examples/relay_control.cpp
@@ -34,7 +34,9 @@ ReactESP app([] () {
   sensesp_app = new SensESPApp(new SensESPAppOptions()
                 ->useWifi("ssdi", "password")
                 ->useServer("10.10.10.1", 3000)
-                ->useStandardSensors(allStdSensors));
+                ->useStandardSensors(all)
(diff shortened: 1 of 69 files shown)


src/net/ws_client.h
```diff
@@ -54,6 +56,7 @@ class WSClient : public Configurable {
   std::function<void(bool)> connected_cb;
   void_cb_func delta_cb;
   bool get_mdns_service(String &server_address, uint16_t& server_port);
+  void setConfigurationFromOptions();
 };
```
ba58smith, line 61: Same comment as above for http.h, about not passing the entire SensESPAppOptions object to the constructor.

src/net/ws_client.cpp
```diff
@@ -184,22 +189,28 @@ void WSClient::connect() {
...
+       debugE("No SignalK server found in network when using mDNS service!");
     }
-  } else {
-    server_address = this->server_address;
-    server_port = this->server_port;
+    else
+    {
+      debugI("SignalK has been found at address %s:%d by mDNS.", server_address.c_str(), server_port);
+    }    
   }
```
ba58smith, line 204: "SignalK server has been found..."? (Add the word "server"?)

src/net/ws_client.cpp
```diff
@@ -427,26 +446,68 @@ static const char SCHEMA[] PROGMEM = R"({
...
+
+void WSClient::setConfigurationFromOptions()
+{
+  if(options->isServerSet())
+  {
+      debugI("Using preconfigured server address %s and port %d from main.cpp", options->getServerAddress().c_str(), options->getServerPort());
+      this->server_address = options->getServerAddress();
+      this->server_port = options->getServerPort();
+  }
+}
```
ba58smith, line 485: You don't need "preconfigured" in the above debug. The fact that it's coming from main.cpp means that it's preconfigured there. (I'm very particular about comments. :-) )

src/sensesp_app_options.cpp
```diff
@@ -0,0 +1,79 @@
...
+    mDNSEnabled = true;
+    wifiSet = false;
+    serverSet = false;
+#ifdef LED_BUILTIN
+    ledPin = LED_BUILTIN;
+#endif
+#ifdef LED
+    ledPin = LED;
+#endif
+
```
ba58smith, line 24: Does this handle the fact that ESP32 doesn't have LED_BUILTIN? I'm not familiar with "LED" as a defined LED pin. Where would that come from?

src/net/http.cpp
```diff
@@ -221,6 +222,45 @@ void HTTPServer::handle_device_restart(AsyncWebServerRequest* request) {
...
+  {
+    response->println("SignalK adddress will be found using mDNS.");
+  }
+  else
+  {
+    response->println("SignalK adddress is set in Web UI");
+  }  
+
+  request->send(response);
 }
```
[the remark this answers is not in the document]
JohnySeven: This is for /info page, it's not new. It prints all the options configuration into page.

src/net/http.h
```diff
@@ -19,6 +20,7 @@ class HTTPServer {
   AsyncWebServer* server;
   std::function<void()> reset_device;
   void handle_config_list(AsyncWebServerRequest* request);
+  SensESPAppOptions*options;
 };
```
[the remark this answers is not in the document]
JohnySeven: Yes, but in this case it seems to be appropriate, because I'm using a lot parameters from options in /info page. And it's expected to be growing in future as more parameters are added. So if you agree let's leave it as it is in this case.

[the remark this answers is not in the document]
ba58smith: Makes sense.

src/net/http.cpp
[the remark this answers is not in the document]
ba58smith: I don't remember seeing it before. Thanks for the explanation.

examples/relay_control.cpp
```diff
@@ -29,9 +30,14 @@ ReactESP app([] () {
...
-
   // Create the global SensESPApp() object.
-  sensesp_app = new SensESPApp();
+  sensesp_app = new SensESPApp([] (SensESPAppOptions*o)
+  {
+         o->setServerOptions("10.10.10.1", 3000)
+          ->setWifiOptions("yourSSID", "yourPassword")
+          ->setHostName("relay")
+          ->setStandardSensors();
+  });
```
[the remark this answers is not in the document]
JohnySeven: Done in latest commit

src/net/ws_client.cpp
JohnySeven: Done

src/sensesp_app_options.cpp
JohnySeven: Reworked, if LED_BUILTIN isn't defined by default LED blinking is disabled. But If user defines LED_BUILTIN or uses options->useLED(...) it will use this config.

platformio.ini
```diff
@@ -31,6 +31,7 @@ platform = espressif32
 board = esp32dev
 framework = arduino
 monitor_speed = 115200
+upload_speed = 460800
```
mairas, line 34: I'd rather not change the upload speed from the default for everyone - I've sometimes had issues with high upload speeds.

src/net/http.cpp
```diff
@@ -221,6 +222,45 @@ void HTTPServer::handle_device_restart(AsyncWebServerRequest* request) {
...
+
+  response->println();
+  response->printf("MAC: %s", WiFi.macAddress().c_str());
+  response->println();
+  response->printf("WiFi signal: %d", WiFi.RSSI());
+  response->println();
+
+  if(options->isWifiSet())
+  {
+    response->println("Wifi configuration is set in main.cpp");
```
mairas, line 238: Henceforth: instead of saying "set in main.cpp", which might or might not be true (someone might have a more complex initialization process), could you rather say "hard-coded" or "set in compile-time options" or something?

src/net/networking.cpp
```diff
@@ -124,33 +139,62 @@ static const char SCHEMA[] PROGMEM = R"({
...
+  }
 }
 
 JsonObject& Networking::get_configuration(JsonBuffer& buf) {
   JsonObject& root = buf.createObject();
+
   root["hostname"] = this->hostname->get();
   root["ap_ssid"] = this->ap_ssid;
-  root["ap_password"] = this->ap_password;
+  root["ap_password"] = this->ap_password;
```
mairas, line 164: Nitpick: trailing whitespace added

src/net/http.cpp
ba58smith: @mairas - I believe the way it's coded now, "set in main.cpp" means exactly that. Can you think of a way that someone could initialize those settings other than by hard-coding them in main.cpp? If someone does a future PR that changes it, then I would agree the language should change. But for right now, I think it's totally accurate, and totally clear. "Hard-coded" could come from an h./.cpp file. "Set in compile-time options" is really vague.

src/sensesp_app_options.h
```diff
@@ -0,0 +1,57 @@
...
+        bool isWifiSet() { return wifiSet; }
+        bool isServerSet() { return serverSet; }
+        LedIntervals_T getLEDIntervals() { return ledIntervals; }
+        String getHostname() { return hostName; }
+        bool isHostNameSet() { return isHostNameChanged; }
+        
+    private:
+        String ssid;
+        String password;
+        bool wifiSet;
```
mairas, line 44: Instead of managing the `wifiSet`, `serverSet` and similar state variables, I'd much rather just get that information from the string itself: if the string is empty, it's not set. That functionality can of course be wrapped to a better named function to self-document the logic, if desired.

Maintaining and distributing two alternate representations of same state (the string itself & some boolean based on the string) is a prime source for future bugs, IMHO.

platformio.ini
JohnySeven: Agreed, will change it.

src/net/http.cpp
mairas: Also, the logic here makes strong assumptions on how different values are set. For code maintainability, I'd prefer just reporting the value and not trying to deduce how or where it was set.

src/net/ws_client.h
JohnySeven: Will change that. Thanks for spotting this.

src/net/http.cpp
ba58smith: Reporting how and where it's set could save someone hours of troubleshooting, if, for example, it's set in main.cpp from the example they copied and modified, but forgot to change that part. I'm speaking as a newbie programmer, who often makes that kind of mistake. Just my opinion.

mairas: @ba58smith someone might simply want to backport SensESP back to Arduino IDE, prefer to use a different name for `main.cpp` in their project, or have a more complicated initialization pattern where different functions are offloaded to separate files. My problem with making these kinds of assumptions is that they increase the future maintenance burden. By not following the assumptions you suddenly have wrong information printed out, and that's not good. The same goes with assuming how some variable got set: the minute someone further improves or changes how the initialization works, these kinds of hard-coded assumptions break and the developer gets no warning. Such traps make future work on the code more difficult than necessary.

mairas:
One more design consideration: if you used the builder pattern as discussed on Slack, you could use the `SensESPApp` class to store and access the configuration options. There is a global pointer to the `app` object, so you wouldn't have to pass the options object around. Accessing the different options could be limited/enabled by defining `HTTPServer` a friend class of `SensESPApp`, if so desired.

src/net/http.cpp
ba58smith: @mairas, I don't disagree. But NOT having information in the Serial Monitor about how these variables are set makes it harder for the end user to troubleshoot problems.
It's your project, and your choice: easier for the end user, or easier for the future developer?

mairas: @ba58smith I'd argue erroneous information is even worse than no information at all. To be clear: I don't insist on not having the conditional reporting of how the values were set (although that might break sooner than later, too), but reporting something as being set in `main.cpp` is both exposing too much implementation details and potentially erroneous as that message won't get automatically updated whenever the implementation changes.

ba58smith: How about saying "...set with SensespAppOptions"? That could be done in main.cpp or sensesp.ino.

mairas: Sure, something like that would be more accurate.

JohnySeven: Are you guys done? 🙂 If so I'll change it to "set with SensespAppOptions", ok?

src/sensesp_app_options.h
JohnySeven: Sure, but I'll need to leave those values in each class, it's useful for locking updates via REST API.

JohnySeven:
@mairas that was my initial approach, but problem with this is the app global pointer is set by user in main.cpp and I can't be sure when the user will do it and I didn't want to change this as it would be breaking change. Another option is to create library wide global pointer to app object that will be set in SensESPApp ctor, then I'll be sure when it's set to current app.

mairas:
> @mairas that was my initial approach, but problem with this is the app global pointer is set by user in main.cpp and I can't be sure when the user will do it and I didn't want to change this as it would be breaking change. Another option is to create library wide global pointer to app object that will be set in SensESPApp ctor, then I'll be sure when it's set to current app.

I don't think you need to use that global pointer during initialization. The only need I see for that is for the `/info` URL handler in `http.cpp`. That needs to access the current state of different objects, and for that the global `sensesp_app` pointer makes most sense. For initialization purposes, it's better to provide any required data as constructor arguments.

src/net/ws_client.h
JohnySeven: Done with latest comit

platformio.ini
JohnySeven: Removed in latest commit.

src/net/http.cpp
JohnySeven: Changed in latest commit.

src/sensesp_app_options.h
JohnySeven: Done in latest commit

JohnySeven:
New commit addresses all requested changes, I've formatted all changed files as @mairas requested. Just let me know if you spot anything else prior merging the changes. Thanks!

ba58smith:
Thoroughly tested with ESP8266. One issue found with configuring hostname in main.cpp, but not configuring ssid and password in main.cpp (or vice versa), but after disccusion with @JohnySeven, we decided to fix that in a new issue after this PR is merged.

Compiles with ESP32, but the known problem with mDNS is preventing further testing at the moment. Will go ahead and merge, then address that issue, then continue testing.

PR #146, opened 2020-08-17
Added connection check to ::loop() function

bwssytems:
Added ws connection check to loop function and reordered connection check in send delta.

This makes the ReactApp to SensApp behavior not guru on you. IT also allows access to the web configuration while it is trying as well.

### DIFF
--- a/src/net/ws_client.cpp
+++ b/src/net/ws_client.cpp
@@ -400,7 +400,11 @@ void WSClient::connect_ws(const String host, const uint16_t port) {
   this->client.setAuthorization(full_token.c_str());
 }
 
-void WSClient::loop() { this->client.loop(); }
+void WSClient::loop() {
+  if (this->connection_state == disconnected) {
+    this->client.loop();
+  }
+}
 
 bool WSClient::is_connected() { return connection_state == connected; }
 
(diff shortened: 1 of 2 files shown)


bwssytems:
Don't merge this yet, I have a stuck situation...

bwssytems:
Have updated to work much more smoothly

mairas:
Looks good to me! How did you test the change?

bwssytems:
Testing was based on this:

Scenario 1 - No mDNS running on SignalK server
1) erase ESP32 flash completely
2) upload firmware
3) connect to wifi
4) watch output wait for mDNS

Scenario 2 - mDNS running - Deny Auth
1) erase ESP32 flash completely
2) upload firmware
3) connect to wifi
4) mDNS connects
5) Auth goes into Pending
6) Auth denied
7) SensESP goes into reconnect and reauth

Scenario 3 - mDNS running - Accept Auth
1) erase ESP32 flash completely
2) upload firmware
3) connect to wifi
4) mDNS connects
5) Auth goes into Pending
6) Auth Approved
7) SensESP gets Auth and connects to WS service and subscribes

I still see a SW CPU Reset right after the send_access_request runs at the point of setting up app.delay on poll_access_request. Not sure where that is coming from, but it is not a guru panic and then the code just continues on after reboot properly

mairas:
Thanks! I made a quick test with an ESP8266 device and it still works fine. Merging.

Issue #144, opened 2020-08-16
Can't set hostname in main.cpp w/o also setting ssid and pw


ba58smith:
The new SensespAppOptions feature that allows various options to be set in main.cpp has a problem if you try to set the SSID and password in main.cpp, but DON'T set the hostname in main.cpp. This is most likely due to the fact that WifiManager saves the hostname as a custom parameter, so when you don't use WifiManager to set SSID and pw, you don't get access to hostname, because WifiManager isn't involved.
I think an acceptable solution would be to bundle these three options: either set all three in main.cpp, or set none of them in main.cpp. But that's up for discussion.

mairas:
Could you verify that this is still the case after my related modifications?

ba58smith:
@mairas, it actually seems much worse now. With this in main.cpp:
```
  // Create the global SensESPApp() object.
  sensesp_app = builder.set_standard_sensors(UPTIME)
              //->set_hostname("d1minitest")
              //->set_sk_server("10.10.10.1", 3000)
              ->set_wifi("contemporary", "heatheranderic")
              ->get_app();
```
the program uploads and runs, and as soon as it gets to `setup_discovery( )`, it crashes hard. Below is everything from "Linking" to the crash.

Linking .pio/build/d1_mini/firmware.elf
Building .pio/build/d1_mini/firmware.bin
Retrieving maximum program size .pio/build/d1_mini/firmware.elf
Checking size .pio/build/d1_mini/firmware.elf
Advanced Memory Usage is available via "PlatformIO Home > Project Inspect"
RAM:   [=====     ]  50.4% (used 41324 bytes from 81920 bytes)
Flash: [========  ]  76.2% (used 795856 bytes from 1044464 bytes)
Creating BIN file ".pio/build/d1_mini/firmware.bin" using "/home/butch/.platformio/packages/framework-arduinoespressif8266/bootloaders/eboot/eboot.elf" and ".pio/build/d1_mini/firmware.elf"
Configuring upload protocol...
AVAILABLE: espota, esptool
CURRENT: upload_protocol = esptool
Looking for upload port...

Warning! Your `/etc/udev/rules.d/99-platformio-udev.rules` are outdated. Please update or reinstall them.
 Mode details: https://docs.platformio.org/en/latest/faq.html#platformio-udev-rules

Auto-detected: /dev/ttyUSB0
Uploading .pio/build/d1_mini/firmware.bin
esptool.py v2.8
Serial port /dev/ttyUSB0
[...]

PR #147, opened 2020-08-17
#144 Can't set hostname in main.cpp w/o also setting ssid and pw - fix

JohnySeven:
This PR contains fix for issue #144 with hostname. I've tested the PR on ESP32 and it works as expected.


JohnySeven:
Something is wrong with my git, please ignore.

PR #148, opened 2020-08-18
Replace SensESPAppOptions with a builder pattern implementation

mairas:
The SensESPAppOptions PR slipped in before my architectural concerns were addressed. This PR simplifies the architecture while retaining the introduced features. Now, a standard builder pattern is used to simplify SensESPApp class instantiation. Although, to be frank, I'm not convinced even that would be necessary; just defining the desired arguments should be enough.


JohnySeven:
I'm sorry I've missed this PR. But good job @mairas 👍

PR #155, opened 2020-09-07
Default hostname

ba58smith:
If hostname is blank when mDNS tries to run, it crashes. I made it "SensESP", which also makes the initial configuration access point's SSID be "Configure SensESP" instead of just "Configure ".


mairas:
Looks good to me! Merging.

PR #157, opened 2020-09-08
Fix problem with hostname in Issue #144

ba58smith:
Hostname can now be set and edited independently of wifi credentials. That is, either or both can be set or not set with SensESPAppBuilder in main.cpp, and still behave properly in the Config UI.


mairas:
Great! Thanks! I'll merge.

PR #161, opened 2020-09-09
Add all Adafruit library dependencies to library.json

ba58smith:
For some reason that no one is sure of, a recent update to PlatformIO has required that all of these libraries be explicitly included in `library.json`.

### DIFF
--- a/library.json
+++ b/library.json
@@ -92,6 +92,21 @@
       "name": "Adafruit AHRS"
     },
     {
+      "name": "Adafruit Unified Sensor"
+    },
+    {
+      "name": "Adafruit BusIO"
+    },
+    {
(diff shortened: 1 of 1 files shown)


ba58smith:
This may not be necessary. Don't merge at this time.

ba58smith:
@mairas - What are your thoughts on this? I think you believe we need to do this, but also don't know if it's still necessary. It certainly is for me and a few others, but maybe we don't have something set up correctly?

mairas:
We don't really need this. I was earlier misinterpreting the library definition files; `library.properties` does have dependencies declared as well. The reason for these suddenly popping up was a (likely) regression in PlatformIO's library dependency finding logic. I'll close this PR.

PR #165, opened 2020-09-12
Add voltageDividerR1 class

RBerliner:
This class calculates the value of the resistor R1 for a voltage divider (R1, R2) where the voltage across R2 is measured .

This is in contrast to the voltageDividerR2 class where the measured voltage across R1 is used to infer R2.


src/transforms/voltagedividerR1.h
```diff
@@ -0,0 +1,37 @@
+#ifndef _voltage_divider_H
+#define _voltage_divider_H
+
```
ba58smith, line 3: voltagedividerR2.h already defines _voltage_divider_H. (Although it should have been _voltage_divider2_H, since the point is to give this particular file - the file in which this little bit of code occurs - a "label". 

Yours should be _voltage_divider1_H, to distinguish it from _voltage_divider_H. If they have the same name, then both files can't be used in the same project.

ba58smith:
I might have just added this new class to the existing voltage_dividerR2 files, and dropped the "R2" from the name. Then both voltage divider classes would be in the same file. However, that would have been a breaking change for any main.cpp that used the R2 version. But that wouldn't be so bad - there is probably only one person's main.cpp that would have been broken, and if they were ever recompiling that file, it would have been an easy and obvious fix for them to make - just change `#include voltage_dividerR2.h` to `#include voltage_divider.h`. Just a thought for the future. (Or for now, if you WANT to.)

Other than that, the only thing I see is my comment above.

RBerliner:
I screwed up with git but removed the eTape example so that we can proceed as the PR started out. I think I did what you wanted and it appears to me to be a cleaner approach.

RBerliner:
Do I need to do more at my end to complete the PR? 
When this is approved, I will PR the eTape liquid level sensor code.

ba58smith:
Ron, looks great! One more thing: please modify the temperature_sender example to change `#include "transforms/voltagedividerR2.h"`. Then I'll merge it.
Look up "git stash" - it's a way to store work in progress so you won't lose it after you submit a PR and then update all your code. Basically, you "stash" the changed files, then create the PR you want with the other files you're working on. When the PR is merged, and you update your working files, that will overwrite the stuff you were working on for the other PR. You restore that work by "un-stashing" what you stashed. Works really well.

RBerliner:
Updated the temperature_sender example to use the Builder class and the new voltageDivider.h include file.

mairas:
This PR has 30 commits, of which the first 24 are in no way relevant to the PR. Please remove them. Also, please don't include merge commits in the PR.

Since the number of unrelated commits is so large, probably the easiest way to remove them is to clone a fresh copy of the repo and reapplying the new changes. Alternatively, you can squash them away by interactive rebasing (i.e. history editing), `git rebase -i` on the command line or (probably easier) by using some graphical user interface tool.

To avoid creating bogus merge commits, use rebasing instead of merging when pulling new changes (`git pull --rebase` on command line, the respective option in a GUI).

ba58smith:
@mairas - Do you have a step-by-step written process for using Git? (I can NEVER remember the exact command lines to use, so I have them written down, and follow them religiously.) I think what you do must be different from the rest of us, because we all have ended up with commits from other work that shouldn't be in our PR's.

RBerliner:
I completely understand the criticism from Marais. In my defense, I am working alone and this is the first collaborative project I have been seriously involved with. I did not know about git pull --rebase but I have looked it up and I see how it suppresses the update merge commits.

I have been looking at the raft of commits in the PR trying to see if there was something there that I needed to not lose. I did a comparison of the local repo branch to the upstream (SignalK) branch:

git difftool master..upstream/master

It says that the only differences are the files really related to the upgrade of the voltageDivider. 

I looked at one of the commits in the PR: Changes to README_MAX31856TC that was commited to my repo in June. The is a note from gihub about this commit "This commit does not belong to any branch on this repository". The changes in that commit are indeed present in the SignalK/SensESP repo. How did they get there and if from an earlier PR, why did they end up in the list of commits in this PR? Clearly I must have done something wrong with git somewhere along the way.

I will delete my github SensESP repo, re-fork the SignalK/SensESP repo clone it to my local machine and make the changes related to the voltageDivider upgrade.

This will make a much cleaner PR

RBerliner:
I have renamed my SensESP repo to SensESP_Old, forked a new repo SensESP, cloned the new repo to my local machine, added in the voltagedivider modifications, commited and pushed. It may be that I should not have followed this route as the list of commits for the PR is unchanged. 

If this will not result in a clean commit, then close the PR without merging  and I will resubmit it.

ba58smith:
@RBerliner - because of the many unrelated commits, I'm going to delete this PR, so you can try to create a nice clean one. I don't know how to tell you to do that, unfortunately.

Issue #151, opened 2020-08-26
BME280, SensESP fails to start if sensor not present


KEGustafsson:
If BME280 device is not present then SensESP startup fails and device is halted.
␛[0m(I) (check_status) Could not find a valid BME280 sensor. Check wiring, address, and sensor ID. ␛[0m(I) (check_status) SensorID is: 0x255 ␛[0m(I) (check_status) 0xFF: is a BMP180 or BMP085, or a bad address ␛[0m(I) (check_status) 0x56-0x58 is a BMP280 ␛[0m(I) (check_status) 0x60 is a BME280 ␛[0m(I) (check_status) 0x61 is a BME680 ␛[0m
Log reporting stops here.

Program stops here.
https://github.com/SignalK/SensESP/blob/master/src/sensors/bme280.cpp#L27

Would be much better to catch the issue internally and just disable reading the values for the sensor and allow the rest of the program to continue.

BjarneBitscrambler:
This behaviour will also happen using the orientation sensor I've been working on ( #145 ).  I'll have a look at my code and see about fixing it - I agree that a system that potentially has dozens of sensors should degrade gracefully if any one of the sensors fails.

ba58smith:
@BjarneBitscrambler - I've fixed this behavior for the BME280 sensor with PR #158. I'll leave it up to you to create an issue for your magnetometer sensor, or just fix it in another PR.

PR #166, opened 2020-09-12
Failed orientation sensor communication no longer hangs/reboots system

BjarneBitscrambler:
### Summary
Issue #151 dealt with a failed BME280 sensor causing the system to fail to start. A similar situation exists with the NXP FXOS8700 / FXAS21002 combination orientation sensor: if I2C communications with the sensor fails, the system hangs. It is preferred that the SensESP program continues to run, to avoid halting data from possible other attached sensors.

### Changes Made
Only two files have changed: `sensor_NXP_FXOS8700_FXAS21002.h` and `sensor_NXP_FXOS8700_FXAS21002.cpp`
* `SensorNXP_FXOS8700_FXAS21002::connect()` changed so failure to init calibration helper (which retrieves cal values stored in EEPROM) does not halt program, but rather allows it to continue with warnings displayed on serial console.
* `SensorNXP_FXOS8700_FXAS21002::initSensors() now checks to see whether communications with the sensor have previously been working. If so, it merely 'pings' the sensors to confirm that comms still work. If comms have previously been reported as failed, or if they have never been initialized, then initSensors() tries to initialize them. In either case, it returns true if comms are currently working, and false if not.
* `SensorNXP_FXOS8700_FXAS21002::gatherOrientationDataOnce()` always calls `initSensors()` before attempting to retrieve data from the sensors, and skips the data retrieval if comms have failed.

### New Program Flow
Each time data collection is attempted from the orientation sensor, comms are checked. A failure is flagged, which causes future data retrievals to be skipped until the failure condition is corrected. Data retrieval then continues normally until the next comms failure.

### Testing Performed
* Disconnecting the orientation sensor's I2C SCL line or VCC power before the system is powered up does not hang or reboot the system. Instead, unchanging orientation values are conveyed in the SignalK traffic, and the serial console displays a message that the orientation data are invalid. Other SignalK parameters (e.g. uptime) continue to update at the expected rate.
* Attaching the orientation sensor any time afterwards causes the orientation data to start refreshing at the expected rate, and the serial console warning ceases.
* Disconnecting the sensor any time during normal operation freezes the present orientation data and starts the console warnings.
* Checking the SignalK sensesp.freemem during repeated disconnects/reconnects does not seem to show any memory leak: it is a little difficult to be absolutely certain as the freemem count bounces around a bit, but a dozen simulated failures over several minutes didn't show any unexpected behaviour.

### Future Possible Work
Does SignalK define a way to flag invalid data? Currently the orientation data freeze at whatever value they had just prior to the sensor comms failure, and this might not be noticed for a while.  I imagine a SignalK alert could be sent out...

### DIFF
--- a/src/sensors/sensor_nxp_fxos8700_fxas21002.cpp
+++ b/src/sensors/sensor_nxp_fxos8700_fxas21002.cpp
@@ -1,77 +1,79 @@
 #include "sensor_nxp_fxos8700_fxas21002.h"
 
 #include <RemoteDebug.h>
+
 #include "sensesp.h"
-#include <Adafruit_FXAS21002C.h>
-#include <Adafruit_FXOS8700.h>
 
 //#define AHRS_DEBUG_OUTPUT  //output diagnostics from sensor read and filter
 #define MIN_PRINT_INTERVAL_MS \
   (200)  // when result printing is requested, it won't happen
          // any more frequently than this,
          // to avoid overloading the serial channel
-
-// can the following be moved inside class? or into main.cpp?
-Adafruit_FXOS8700 fxos = Adafruit_FXOS8700(0x8700A, 0x8700B);  
-Adafruit_FXAS21002C fxas = Adafruit_FXAS21002C(0x0021002C);
+#define FXAS_CR1_ACTIVE_BITMASK (0b00000010) //Active bit in gyro CTRL_REG1
+#define FXAS_CR1_READY_BITMASK (0b00000001) //Ready bit in gyro CTRL_REG1
+#define FXAS_CR1_ODR_BITMASK (0b00011100) //Output Data Rate in gyro CTRL_REG1
+#define FXAS_CR1_ODR_BITS_LOC (2) //ODR bits are in positions 4:2
+#define N2K_INVALID_FLOAT (-1e-9) //NMEA2000 value for unavailable parameters
 
 Adafruit_NXPSensorFusion filter;  // when placed inside class, results from this
                                   // filter return Heading=0.0 always. Other two
                                   // filters (see below) work OK inside class
 
-SensorNXP_FXOS8700_FXAS21002::SensorNXP_FXOS8700_FXAS21002() {}
+//  Constructor creates an accelerometer/magnetometer object (fxos_)
+//  and a gyroscope object (fxas_)
+SensorNXP_FXOS8700_FXAS21002::SensorNXP_FXOS8700_FXAS21002()
+    : fxos_(0x8700A, 0x8700B), fxas_(0x0021002C) {}
 
 //  Connect to FXOS8700 & FXAS21002 sensor combination using I2C.
 //  To use default Arduino I2C pins, pass pin_i2c_sda and pin_i2c_scl = -1
 //  Load calibration values if available (readings won't be valid without
 //    calibration).
-//  TODO: set ranges/sensitivity. For now use ICs' default values.
-bool SensorNXP_FXOS8700_FXAS21002::connect(uint8_t pin_i2c_sda,
-                                           uint8_t pin_i2c_scl) {
+//  Ranges/sensitivity use ICs' default values, then setupSensors() adjusts.
+bool SensorNXP_FXOS8700_FXAS21002::connect(int pin_i2c_sda,
+                                           int pin_i2c_scl) {
   debugI("NXP 9 Degrees-of-Freedom Sensor with Adafruit AHRS");
-  if (!cal.begin()) {
-      debugE("Failed to initialize calibration helper");
-    while (1) yield();
-  }
-  if (!cal.loadCalibration()) {
+  if (!cal_.begin()) {
+    debugE("Failed to initialize calibration helper");
+    is_calibrated_ = false; 
+  }else { 
+    if (!cal_.loadCalibration()) {
       debugI("No calibration loaded/found...will start with defaults");
-    isCalibrated = false;
-  } else {
+    is_calibrated_ = false;
+    } else {
       debugI("Loaded existing calibration:");
-    cal.printSavedCalibration();
-    isCalibrated = true;
-  }
+    cal_.printSavedCalibration();
+    is_calibrated_ = true;
+    }
+  }//end of loading calibration
 
-  Wire.begin(pin_i2c_sda, pin_i2c_scl);
-  Wire.setClock(400000);  // 400KHz I2C
-
-  if (!initSensors()) {
+  Wire.begin(pin_i2c_sda, pin_i2c_scl, 400000 );
+  if (!startSensors()) {
       debugE("Failed to find sensors");
     return false;
   }
 
-  setupSensors();       // TODO - can set ranges, etc
-  printSensorDetails();
+  setupSensors();       //set ranges, etc
 
   return true;
 }  // end connect()
 
 //  Prints details of FXOS8700 & FXAS21002 sensor combination
 void SensorNXP_FXOS8700_FXAS21002::printSensorDetails(void) {
-  accelerometer->printSensorDetails();
-  gyroscope->printSensorDetails();
-  magnetometer->printSensorDetails();
+  accelerometer_->printSensorDetails();
+  gyroscope_->printSensorDetails();
+  magnetometer_->printSensorDetails();
 }  // end printSensorDetails()
 
 void SensorNXP_FXOS8700_FXAS21002::initFilter(int sampling_interval_ms) {
-  roll = 0.0;
-  pitch = 0.0;
-  heading = 0.0;
-  gx = 0.0;
-  gy = 0.0;
-  gz = 0.0;
-  filter.begin(1000.0 / sampling_interval_ms);
-  timestamp = millis();
+  roll_ = 0.0;
+  pitch_ = 0.0;
+  heading_ = 0.0;
+  gx_ = 0.0;
+  gy_ = 0.0;
+  gz_ = 0.0;
+  sampling_rate_hz_ = 1000.0 / sampling_interval_ms;
+  filter.begin(sampling_rate_hz_);
+  timestamp_ = millis();
 }  // end initFilter()
 
 /* The following get___() methods return the specified parameter
@@ -81,43 +83,43 @@ is aligned with the X-axis pointing to the Bow;
 the Y-axis pointing to Port; and the Z-axis pointing up. If your
 sensor is mounted differently, changes to the terminology/axes
 will need to be made.
-Per SignalK convention, units are SI:  radians for headings;
+Per Signal K convention, units are SI:  radians for headings;
 radians / second for turn rates; and meters / second^2 for acceleration.
 */
 float SensorNXP_FXOS8700_FXAS21002::getHeadingRadians(void) {
-  return (360.0 - heading) / SENSORS_RADS_TO_DPS;
+  return (360.0 - heading_) / SENSORS_RADS_TO_DPS;
 }  // end getHeadingRadians()
 
 float SensorNXP_FXOS8700_FXAS21002::getPitchRadians(void) {
-  return (-pitch) / SENSORS_RADS_TO_DPS;
+  return (-pitch_) / SENSORS_RADS_TO_DPS;
 }  // end getPitchRadians()
 
 float SensorNXP_FXOS8700_FXAS21002::getRollRadians(void) {
-  return roll / SENSORS_RADS_TO_DPS;
+  return roll_ / SENSORS_RADS_TO_DPS;
 }  // end getRollRadians()
 
 float SensorNXP_FXOS8700_FXAS21002::getAccelerationX(void) {
-  return accel_event.acceleration.x;
+  return accel_event_.acceleration.x;
 }  // end getAccelerationX()
 
 float SensorNXP_FXOS8700_FXAS21002::getAccelerationY(void) {
-  return accel_event.acceleration.y;
+  return accel_event_.acceleration.y;
 }  // end getAccelerationY()
 
 float SensorNXP_FXOS8700_FXAS21002::getAccelerationZ(void) {
-  return accel_event.acceleration.z;
+  return accel_event_.acceleration.z;
 }  // end getAccelerationZ()
 
 float SensorNXP_FXOS8700_FXAS21002::getRateOfTurn(void) {
-  return (-gz) / SENSORS_RADS_TO_DPS;
+  return (-gz_) / SENSORS_RADS_TO_DPS;
 }  // end getRateOfTurn()
 
 float SensorNXP_FXOS8700_FXAS21002::getRateOfPitch(void) {
-  return (-gy) / SENSORS_RADS_TO_DPS;
+  return (-gy_) / SENSORS_RADS_TO_DPS;
 }  // end getRateOfPitch()
 
 float SensorNXP_FXOS8700_FXAS21002::getRateOfRoll(void) {
-  return gx / SENSORS_RADS_TO_DPS;
+  return gx_ / SENSORS_RADS_TO_DPS;
 }  // end getRateOfRoll()
 
 // fetches data from sensors, applies filter function, and assigns orientation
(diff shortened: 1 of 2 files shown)


BjarneBitscrambler:
@ba58smith - since you brought up the issue, and fixed it in the BME280 code, would you mind reviewing my orientation sensor fixes for the same problem?  Thanks a bunch!

src/sensors/sensor_NXP_FXOS8700_FXAS21002.cpp
```diff
@@ -30,27 +30,23 @@ bool SensorNXP_FXOS8700_FXAS21002::connect(uint8_t pin_i2c_sda,
                                            uint8_t pin_i2c_scl) {
   debugI("NXP 9 Degrees-of-Freedom Sensor with Adafruit AHRS");
   if (!cal.begin()) {
-      debugE("Failed to initialize calibration helper");
-    while (1) yield();
-  }
-  if (!cal.loadCalibration()) {
+    debugE("Failed to initialize calibration helper");
+    isCalibrated = false; //TODO - check what happens if cal.begin fails
+  }else {
```
ba58smith, line 35: Do you still need to do this TODO? If so, please do it. If not, please remove the comment.

ba58smith:
It appears that your code is doing what your EXCELLENT description says. Since you've tested it, and it works as expected - please resolve the one TODO comment above, and I'll merge it. Nice work. I need to start describing my PR's as well as you do.

src/sensors/sensor_NXP_FXOS8700_FXAS21002.cpp
BjarneBitscrambler: Thanks for noticing that one...  I followed the breadcrumbs and confirmed that if loading the calibration values fails for any reason, then the default values that will be used have no effect on the raw sensor values (i.e. the offsets are zero and the transformation matrix is an identity matrix).  Revised .cpp file is committed.

BjarneBitscrambler:
@ba58smith - back to you. The TODO is DONE, and there are no adverse effects.

Thanks for your kind words re: documentation. I've noticed over the years that my memory is short enough that when I return to something a few months later I start cussing my past self if I haven't written stuff down :-)  That, plus my work in the field of medical device design, motivates me to try for thoroughness...

mairas (changes requested):
I don't think this file is related to the PR topic?

mairas:
Please remove the first three unrelated commits from the PR.

You can squash the extra away by interactive rebasing (i.e. history editing), git rebase -i on the command line or (probably easier) by using some graphical user interface tool. If that proves difficult, you can clone a fresh copy of the repo and reapply the new changes. 

To avoid creating bogus merge commits, use rebasing instead of merging when pulling new changes (git pull --rebase on command line, the respective option in a GUI).

BjarneBitscrambler:
@mairas - sorry, I've made a hash of this. I've tried rebasing using both command-line git and Github Desktop, but I can't seem to squash all the commits into one on my remote repository. Rebasing locally seems to work, but then when I try pushing that to my Github repository it complains that my current branch is behind its remote counterpart.

The closest I could come was to revert the unwanted commits, but while that gets rid of the unwanted changes, it also adds yet another commit to the history.

When you suggest "...clone a fresh copy of the repo and reapply the new changes" could you describe in more detail how I go about that?  Is that the same as closing this PR and starting a new one?

ba58smith:
@BjarneBitscrambler - I think what he means is for you to completely delete your local copy of the SensESP files - the ones that are your Git files. Don't just delete the files, delete the folder. Also, delete your .pio folder. under your PlatformIO SensESP project(s).

Now, in your Git "root" directory (the directory in which all your Git projects live - for me it's Documents/GitHubProjects), do `git clone https://github.com/YOURGITHUBDIRECTORY/SensESP.git`, which will not only give you the current version of the SensESP source files, it will also create a new, clean local git repo, that doesn't know anything about anything you've ever done with SensESP. (This may not be exactly right - Google "cloning my own github fork".)

Then you need to create your git remotes - typically called "origin" and "upstream", which point to your fork of SensESP on GitHub, and to the main SensESP repo on GitHub. (The "origin" remote might be created by the "clone" command - I don't remember. Google how to set up "upstream" and "origin" remotes.)

At this point, everything should be clean, and in sync with the main SensESP repo, and you can start making commits that shouldn't know anything about any previous commits. I think. Something like that, anyway.

mairas:
@BjarneBitscrambler Brian's advice is spot on if you want to start clean; however, if you managed to locally rebase, that's great! Now, when you try to push to the remote branch, yes, Git will complain because you'll be rewriting history. But when you're doing that on your own branch, that's perfectly fine! You just have to "force-push", that is, `git push -f etc etc`. That will update the PR and sync it with your local history.

However, be warned though; I made a PR to remove the magnetometer code. That is temporary only; there were dependency issues due to conflicting library versions, aggravated by the coinciding PlatformIO version update. Once the dependency issues have been resolved, your code must be taken back in, of course. But while this is under progress, this PR should remain on hold.

BjarneBitscrambler:
OK - thanks! The advice from @mairas  to force the push was successful, but I appreciate the explanation of the how to do a clean restart from @ba58smith. 

The extraneous commits have now been 'forgotten' and the two sensor files seem to be in the desired state. 

There is one thing I am **still concerned about**: looking at the comment history for this PR I see a commit by Brian listed between Matti's change request and my reply. It's labelled  **Correct erroneous Debug message (#162) (a07dcb8)**  Any idea how that crept into the history for this PR?  Was it caused when I merged upstream SensESP into my local repository? It's causing this PR to list 3 changed files, rather than two, on account of `src/net/networking.cpp` being affected by that commit.

BjarneBitscrambler:
@mairas - no worries about holding off on the orientation sensor until after the library dependencies are sorted out.

src/sensors/sensor_nxp_fxos8700_fxas21002.cpp
```diff
@@ -289,14 +313,64 @@ bool SensorNXP_FXOS8700_FXAS21002::initSensors() {
...
+  if (Wire.endTransmission(false) != 0) {
+    return 0;
+  }
+  Wire.requestFrom(address, (byte)1);
+  return Wire.read();
+
+}//end readByte()
+
+
 // Receives calibration values sent by external calibration utility
```
ba58smith, line 374: There seems to be a LOT going on in this commit - not only the primary purpose of the PR (to stop the endless loop if communication with the sensor fails), but also to enhance the calibration process, and maybe some other things. In the future, if you can break things like this into separate commits, that will make it a much easier to review.

I don't see anything obviously wrong, but it's a very technical sensor, so I can't really comment on how it's going to work. I have to trust that you have tested it and have gotten accurate results.

BjarneBitscrambler: Hi Brian - yes, it's gotten a bit complicated due to other changes I was working on while the orientation sensor files were 'paused'.  But I think I have a handle on it, and am doing testing today to confirm it all works.  I expect to be done in a couple days and then be ready for a review.  I'll also perform a rebase to simplify the commits.

src/sensors/sensor_nxp_fxos8700_fxas21002.cpp
```diff
@@ -1,77 +1,79 @@
...
 
 //#define AHRS_DEBUG_OUTPUT  //output diagnostics from sensor read and filter
 #define MIN_PRINT_INTERVAL_MS \
   (200)  // when result printing is requested, it won't happen
          // any more frequently than this,
          // to avoid overloading the serial channel
-
-// can the following be moved inside class? or into main.cpp?
-Adafruit_FXOS8700 fxos = Adafruit_FXOS8700(0x8700A, 0x8700B);  
-Adafruit_FXAS21002C fxas = Adafruit_FXAS21002C(0x0021002C);
```
BjarneBitscrambler, line 16: Moved the fxos and fxas objects inside class (see .h file) and instantiated in `SensorNXP_FXOS8700_FXAS21002` constructor.

src/sensors/sensor_nxp_fxos8700_fxas21002.cpp
```diff
@@ -1,77 +1,79 @@
...
 //  Connect to FXOS8700 & FXAS21002 sensor combination using I2C.
 //  To use default Arduino I2C pins, pass pin_i2c_sda and pin_i2c_scl = -1
 //  Load calibration values if available (readings won't be valid without
 //    calibration).
-//  TODO: set ranges/sensitivity. For now use ICs' default values.
-bool SensorNXP_FXOS8700_FXAS21002::connect(uint8_t pin_i2c_sda,
-                                           uint8_t pin_i2c_scl) {
+//  Ranges/sensitivity use ICs' default values, then setupSensors() adjusts.
+bool SensorNXP_FXOS8700_FXAS21002::connect(int pin_i2c_sda,
+                                           int pin_i2c_scl) {
```
BjarneBitscrambler, line 33: removed TODO, as `connect()` calls `setupSensors()` which now does adjust some sensor parameters.

src/sensors/sensor_nxp_fxos8700_fxas21002.cpp
```diff
@@ -1,77 +1,79 @@
...
+  if (!cal_.begin()) {
+    debugE("Failed to initialize calibration helper");
+    is_calibrated_ = false; 
+  }else { 
+    if (!cal_.loadCalibration()) {
       debugI("No calibration loaded/found...will start with defaults");
-    isCalibrated = false;
-  } else {
+    is_calibrated_ = false;
+    } else {
```
BjarneBitscrambler, line 42: Instead of entering infinite loop if calibration setup fails, we set flag saying that data is uncalibrated. Orientation readings will be made, but the debug console will warn that they are uncalibrated.

src/sensors/sensor_nxp_fxos8700_fxas21002.cpp
```diff
@@ -1,77 +1,79 @@
...
-  Wire.begin(pin_i2c_sda, pin_i2c_scl);
-  Wire.setClock(400000);  // 400KHz I2C
+    cal_.printSavedCalibration();
+    is_calibrated_ = true;
+    }
+  }//end of loading calibration
 
-  if (!initSensors()) {
+  Wire.begin(pin_i2c_sda, pin_i2c_scl, 400000 );
+  if (!startSensors()) {
```
BjarneBitscrambler, line 50: Combined calls to setup I2C and set I2C speed into single call.

src/sensors/sensor_nxp_fxos8700_fxas21002.cpp
```diff
@@ -1,77 +1,79 @@
...
       debugE("Failed to find sensors");
     return false;
   }
 
-  setupSensors();       // TODO - can set ranges, etc
-  printSensorDetails();
+  setupSensors();       //set ranges, etc
```
BjarneBitscrambler, line 55: printSensorDetails() removed, as it duplicates what was already done in main.cpp

src/sensors/sensor_nxp_fxos8700_fxas21002.cpp
```diff
@@ -1,77 +1,79 @@
...
 }  // end connect()
 
 //  Prints details of FXOS8700 & FXAS21002 sensor combination
 void SensorNXP_FXOS8700_FXAS21002::printSensorDetails(void) {
-  accelerometer->printSensorDetails();
-  gyroscope->printSensorDetails();
-  magnetometer->printSensorDetails();
+  accelerometer_->printSensorDetails();
+  gyroscope_->printSensorDetails();
+  magnetometer_->printSensorDetails();
```
BjarneBitscrambler, line 64: Renamed class variables to closer match Google coding style.  Also seen in following lines of code (without a comment).

src/sensors/sensor_nxp_fxos8700_fxas21002.cpp
```diff
@@ -81,107 +83,129 @@ is aligned with the X-axis pointing to the Bow;
...
+    gy_ = N2K_INVALID_FLOAT;
+    gz_ = N2K_INVALID_FLOAT;
+    roll_ = N2K_INVALID_FLOAT;
+    pitch_ = N2K_INVALID_FLOAT;
+    heading_ = N2K_INVALID_FLOAT;
+    accel_event_.acceleration.x = N2K_INVALID_FLOAT;
+    accel_event_.acceleration.y = N2K_INVALID_FLOAT;
+    accel_event_.acceleration.z = N2K_INVALID_FLOAT;
+    return;
+  }
```
BjarneBitscrambler, line 146: each time we call for an updated sensor value, we call startSensors(), which confirms that the sensor is still responding to an I2C ping.  If not, then it attempts to re-initialize the connection. If no connection is made, the orientation variables are set to special values that NMEA2000 recognizes as "N/A" or "Undefined". This reduces chance that a failed sensor will not be recognized (as opposed to the displays just freezing at the most recent valid value).

src/sensors/sensor_nxp_fxos8700_fxas21002.cpp
```diff
@@ -254,49 +278,139 @@ void SensorNXP_FXOS8700_FXAS21002::gatherCalibrationDataOnce(
...
+      accelerometer_ = fxos_.getAccelerometerSensor();
+      magnetometer_ = fxos_.getMagnetometerSensor();
+      gyroscope_ = &fxas_;
+      setupSensors();
+    }
+  }else { //already connected, we believe
+    //ping each sensor to ensure it's still working/attached
+    if( !pingFXAS21002() || !pingFXAS8700() ) {
+      is_connected_ = false;  //ping unsuccessful: assume no connection
+    }
```
BjarneBitscrambler, line 325: Changed name of initSensors() to startSensors() (to better reflect function).  Made the routine more intelligent:  if not connected, it calls the sensor begin() methods as before. If already connected (based on results from previous call to fetch orientation data) it pings the sensors to confirm that they are still responding. This way is_connected_ always reflects the most current sensor status, and they can be unplugged/replugged without crashing SensESP.  When _is_connected_ is false, the orientation values provided to SensESP are the NMEA2000 values for not-available.

src/sensors/sensor_nxp_fxos8700_fxas21002.cpp
```diff
@@ -254,49 +278,139 @@ void SensorNXP_FXOS8700_FXAS21002::gatherCalibrationDataOnce(
...
-  gyroscope = &fxas;
 
-  return true;
-}  // end initSensors()
+  return is_connected_;
+}  // end startSensors()
 
 void SensorNXP_FXOS8700_FXAS21002::setupSensors(void) {
-  //TODO: we could set the g range for accelerometer here, for example
+  //Sets ranges for sensors, when defaults are not optimal.
```
BjarneBitscrambler, line 332: setupSensors() now sets the gyro data rate to correspond with the sensor's update rate, which means the internal low-pass filter is appropriate for the sampling rate. The default power-on setting is OK for 100 Hz sampling, but can lead to digital aliasing at lower sampling rates.

src/sensors/sensor_nxp_fxos8700_fxas21002.cpp
```diff
@@ -254,49 +278,139 @@ void SensorNXP_FXOS8700_FXAS21002::gatherCalibrationDataOnce(
...
+  byte ctl_reg1 = readByte(FXAS21002C_ADDRESS,GYRO_REGISTER_CTRL_REG1); //get reg value
+  ctl_reg1 &= ~(FXAS_CR1_ACTIVE_BITMASK | FXAS_CR1_READY_BITMASK); //zero out the Active & Ready bits
+  writeByte(FXAS21002C_ADDRESS,GYRO_REGISTER_CTRL_REG1, ctl_reg1); //put gyro in standby 
+  ctl_reg1 = (ctl_reg1 & (~(FXAS_CR1_ODR_BITMASK | FXAS_CR1_ACTIVE_BITMASK))) | odr_bits | FXAS_CR1_ACTIVE_BITMASK;
+  writeByte(FXAS21002C_ADDRESS,GYRO_REGISTER_CTRL_REG1, ctl_reg1 ); //back to Active mode
+
 }  // end setupSensors()
 
+//checks I2C bus to see whether FXAS21002 is connected and responding
+bool SensorNXP_FXOS8700_FXAS21002::pingFXAS21002( void ) {
```
BjarneBitscrambler, line 372: Added new functions to ping the sensors (these weren't available in the AdaFruit libraries unfortunately), so we can quickly check whether the sensor is responding before each reading of the data.

src/sensors/sensor_nxp_fxos8700_fxas21002.h
```diff
@@ -28,20 +28,22 @@
...
 
 // Relies on Serial object existing for sending data/diagnostics to serial port
 
 #include <stdint.h>
 
 #include <Adafruit_AHRS.h>
-#include "Adafruit_Sensor_Calibration.h"
+#include <Adafruit_FXAS21002C.h>
+#include <Adafruit_FXOS8700.h>
+#include <Adafruit_Sensor_Calibration.h>
```
BjarneBitscrambler, line 40: Moved these two defines into .h from .cpp as they are needed for the Adafruit_FXOS8700 and Adafruit_FXAS21002C objects (which used to be globally variables created in the .cpp file)

src/sensors/sensor_nxp_fxos8700_fxas21002.h
```diff
@@ -65,24 +67,34 @@ class SensorNXP_FXOS8700_FXAS21002 {
...
-  Adafruit_Sensor_Calibration_EEPROM cal;  // use EEPROM on ESP32 module
-  byte caldata[68];  // buffer to receive magnetic calibration data
-  byte calcount = 0;
-  bool isCalibrated = false;
-  sensors_event_t mag_event, gyro_event, accel_event;
-  uint32_t timestamp;
-  uint32_t last_print_time = 0;
-  float roll, pitch, heading;  // rotation values calculated by filter function,
+  Adafruit_FXOS8700 fxos_;    // the combined magnetometer + accelerometer
+  Adafruit_FXAS21002C fxas_;  // the gyroscope
```
BjarneBitscrambler, line 71: Moved these here from the .cpp file, where they were global variables.

BjarneBitscrambler:
OK @ba58smith - I've implemented the changes needed to keep SensESP running during a sensor outage. Testing as described in the first comment of this PR has been done, and it functions OK on a ESP-Wrover board.  It also builds OK for the d1_mini and esp32dev targets, but I don't have those boards to test on.

In addition to the changes needed for accounting for sensor failures, I also made 2 additional types of changes:
- variable name changes to more closely adhere to our coding style guide. I didn't do *all* the variables, as I didn't want to affect any external files.
- added content to setupSensors(), which used to be an empty method. It now adjusts the internal sampling rate of the gyro to match whatever the SensESP reporting rate is set to.

I've added comments to this PR inline with the code changes to hopefully help explain the changes. Let me know of any concerns you have!  Cheers.

tedenda:
@BjarneBitscrambler , I got my NXP FX87 sensor in the mail yesterday, I can set it up with a Wemos D1-mini and see that it works. Can do that during weekend.

ba58smith:
It all looks good to me, but again, it's so technical, I can't comment on how well it will work. I think it would be good to wait until you test it with the D1 Mini before merging, so after you test, update here, please. Then, given the recent difficulties we've been having with all sorts of newly merged changes - I'll leave it up to @mairas to do the merge.

tedenda:
@BjarneBitscrambler I have now pulled your code and are running it on a Wemos D1-mini on my desk with success. I see that my heading is drifting a lot at a constant rate, have done the mag calibration. But I don't know if that is something that is related to this PR at all. Maybe we can connect on Slack and discuss what I can test?
Drifting heading

BjarneBitscrambler:
@tedenda : That's great you have it running on your d1_mini!  @mairas can now merge this PR, at his convenience.

The drifting heading is something I've seen before, and it's not related to this or any existing PR/Issue.  It's happened twice in the last 2 weeks - the rest of the time the heading is quite steady when the sensor is stationary. I haven't tracked down what's causing it yet - my current theory is that there is a zero offset in the gyro data that makes the filter algorithm think there's motion when there isn't. There may also be a scaling issue, whereby the units of the gyro data aren't matching what the filter expects: I believe this because in my testing I also see overshooting, where a sudden motion causes the reported heading to go past the new resting position and then take several seconds to recover.  Hopefully with two people now using the sensor we can figure it out quickly :-)   I'm not on Slack, but if you want to discuss this over email I think that would be productive and great. You can reach me at bhansen.victoria@gmail.com

tedenda:
@BjarneBitscrambler Email sent, now you have my email too!

DanielG86:
I received my FXOS8700 sensor as well last week and build it yesterday. I have it connected to a Lolin D32 and compiling went as a charm, apart from some of my beginner's mistakes. Calibration also very easy, and the sensor works fine.
The only thing I've noticed is that there is some overshoot in readings, when the sensor makes significant moves.
The above changes are not included in the code yet, as far as I can see. Furthermore, I had to change the SDA and SCL PIN numbers for my board, but that's easy.
Thanks for the great work sofar!

tedenda:
Good that you got it up and running, we are also one more user that can test an improve code for this sensor.
@BjarneBitscrambler is working on documentation and test setups for this sensor, please visit him on GitHub.

tedenda:
@DanielG86 Please checkout @BjarneBitscrambler code (this PR) and run that on your board and report back here, if it works fine we can merge this PR and move on with attitude and other things regarding this sensor.

DanielG86:
I will take a look tomorrow


Verstuurd vanaf mijn iPad

DanielG86:
Just tested it on my board and it all seems fine. Works the same as before. Very stable outputs, except some overshoot on sudden moves, before the data settles on the correct number

tedenda:
@DanielG86 Thanks for testing this PR!
@mairas You can merge this PR when you have the time!

DanielG86:
After initial excellent operation, I did some further testing, which showed me some unexpected behavior: 
1) The Lolin device stops supplying deltas to the server after approx 45 minutes. Exact timing not sure, but I've seen this now a couple of times. The Lolin is still there on the network, and I can access the web-page with configuration options, but the server doesn't receive updates anymore. A "Restart device" brings the updates back.
2) The LED on the Lolin is flashing the first time after some down-time, but after restarts not anymore. There is no correlation to whether the device sends updates to the server. In other words: the LED can be off while updates are still sent. Not really an issue, but this way I can't check whether the device sends updates based on the LED signal
3) The sensor sends saw-tooth signals like @tedenda saw, albeit with a period of approx 3 minutes.This could be due to calibration, because I re-arranged the breadboard and device. Will check whether re-calibration will do the trick.

DanielG86:
Update: sawtooth behavior depends on proximity of USB cable. And recalibration does the trick, so forget about item 3

tedenda:
@DanielG86 I should mount my sensor away from the desk and recalibrate and see if that fixes my sawtooth heading. There is probably some interference that causes it.

Take a look on @BjarneBitscrambler guthub, there is a lot to read there about this sensor and maybe you can help with something.

DanielG86:
Thanks. 
I’ve looked at his wiki pages, and that looks really interesting. I posted an ‘issue’, on his site, offering my support.

Regarding sawtooth, there was a real clear relation with the usb wire I accidentally had really close to the sensor. Quite a big impact of a relatively small current, I would say, and this gives some concern as to when I would take it from the desk to the boat. 

Verstuurd vanaf mijn iPad

BjarneBitscrambler:
@mairas - do you have time to merge this PR?  I'd hate to see it get stale again and overtaken by other changes to SensESP...

Just to summarize: this modification has been tested on the ESP32-WROVER, d1_mini, and LOLIN D32 boards.

mairas:
Sorry, I had failed to follow the PR closely enough! Will merge.

I've got an important-ish question, related to Issue #239, though: after this PR, are there any global variables left in the magnetometer code? They are always linked in regardless of their use, and any SensESP program was using 3 KiB more memory after the initial magnetometer code got merged. I saw that you moved two of the globals inside a class, which probably helps a lot, but it's essential that there are none.

PR #167, opened 2020-09-13
Revert magnetometer and library dependency declarations

mairas:
These changes fix the build on PlatformIO 4.3.4.

On PlatformIO 5.0.1, a local build still works, but a separate project with a dependency to SensESP still fails. For some reason the library dependency finder fails to register "Adafruit Unified Sensor", a dependency of Adafruit BME280 and BMP280. This, I believe, is a PIO bug but I before filing a report I need to have this PR merged.


ba58smith:
Merge away!

PR #171, opened 2020-09-17
Add the VoltageDividerR1 class and update the temperature_sender example

RBerliner:
Added the VoltageDividerR1 class and placed VoltageDividerR1 and VoltageDividerR2 in the same file VoltageDivider.cpp.
Replaced voltagedividerR1.h with voltagedivider.h

Modified the temperature sender to use voltagedivider.h and the Builder class


RBerliner:
In this case, I don not believe using the PRBranch was necessary. Both PRBranch and master have the same commit history.

ba58smith:
Regardless, we're going to start requiring a separate branche for each PR from this point forward. Everything I read about using Git in a community environment like this says it's important.

PR #174, opened 2020-09-28
Sample platformio.ini

mairas:
Provide a sample `platformio.ini` file in `examples`. Also reformat and standardize variable naming for all examples.

### DIFF
--- a/examples/OneWire_temperatures.cpp
+++ b/examples/OneWire_temperatures.cpp
@@ -7,18 +7,18 @@
 #include "sensesp_app.h"
 #include "sensors/onewire_temperature.h"
 #include "signalk/signalk_output.h"
-#include "wiring_helpers.h"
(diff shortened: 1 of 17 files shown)


JohnySeven:
@mairas maybe it would be nice to have complete copy&paste example as kick start? Add mention in the getting started for downloading .zip file with simple example - maybe just sending default sensors?

grahamkendall82:
@mairas..I appreciate this greatly. I was suffering from information overload on what to do with the libraries but it all makes sense now. Thank you

mairas:
> @mairas maybe it would be nice to have complete copy&paste example as kick start? Add mention in the getting started for downloading .zip file with simple example - maybe just sending default sensors?

That's a good idea! Maybe you can make a PR? :-)

mairas:
Merging.

PR #180, opened 2020-10-01
Example platformio.ini environments

mairas:
I had accidentally enabled more than one environment in the example platformio.ini file, which probably was counterproductive. Fixed now. Also added commented out sections for setting upload ports and the builtin led in ESP32.


ba58smith:
Looks good to me!

mairas:
Cool, merging!

PR #183, opened 2020-10-04
ESP32AnalogInput addition

GaryWSmith:
The ESP32 requires a characterisation / calibration of the ADC before reads can be called. So accomodating it is not a matter of modifying the analog_voltage code.

It also has options that are not available with the ESP8266 which are correctly handled in this code.


GaryWSmith:
Contained personal details. 
Will resubmit

PR #184, opened 2020-10-04
New ESP32AnalogInput sensor

GaryWSmith:
The correct handling of the ESP32 analog inputs requires a separate unit.

### DIFF
--- a/examples/esp32analog_input.cpp
+++ b/examples/esp32analog_input.cpp
@@ -33,10 +33,10 @@ ReactESP app([] () {
   SensESPAppBuilder builder;
 
   sensesp_app = builder.set_hostname("test_unit")
-                ->set_wifi("MysticBlue", "passwors1234")
+                ->set_wifi("your_ssid", "your_password")
                 //->set_standard_sensors(IP_ADDRESS)
                 ->set_standard_sensors(NONE)
-                ->set_sk_server("192.168.1.141", 3000) 
+                ->set_sk_server("xxx.xxx.xxx.xxxx", 3000) 
                 ->get_app();
   
 // In the test setup the input voltage was 0.651 in which case all of the options below all 
--- a/examples/esp32analog_input.cpp
+++ b/examples/esp32analog_input.cpp
@@ -1,0 +1,56 @@
+#include <Arduino.h>
+
+#include "sensesp_app.h"
+#include "sensesp_app_builder.h"
+
+#include "signalk/signalk_output.h"
+
+#include "sensors/esp32_analog_input.h"
+
+ReactESP app([] () {
+
+  // Some initialization boilerplate when in debug mode...
+  #ifndef SERIAL_DEBUG_DISABLED
+  Serial.begin(115200);
+
+  // A small delay and one debugI() are required so that
+  // the serial output displays everything
+  delay(100);
+  Debug.setSerialEnabled(true);
+  #endif
+  delay(100);
+  debugI("Serial debug enabled");
+
+  // Create the global SensESPApp() object. 
+  // If you add the line ->set_wifi("your ssid", "your password") you can specify
+  // the wifi parameters in the builder. If you do not do that, the SensESP 
+  // device wifi configuration hotspot will appear and you can use a web 
+  // browser pointed to 192.168.4.1 to configure the wifi parameters. 
+  // You can use NONE, UPTIME, FREQUENCY, FREE_MEMORY, WIFI_SIGNAL or ALL 
+  // instead of IP_ADDRESS in set_standard_sensors().
+
+  // Create a builder object
+  SensESPAppBuilder builder;
+
+  sensesp_app = builder.set_hostname("test_unit")
+                ->set_wifi("MysticBlue", "passwors1234")
+                //->set_standard_sensors(IP_ADDRESS)
+                ->set_standard_sensors(NONE)
+                ->set_sk_server("192.168.1.141", 3000) 
+                ->get_app();
+  
+// In the test setup the input voltage was 0.651 in which case all of the options below all 
+// produced a similar ouput although the difference in resolution between the first and 
+// last statement is clear - as expected.
+
+  auto* pAnalogInput = new ESP32AnalogInput(34, ADC_WIDTH_BIT_12, ADC_ATTEN_DB_0, 1000);
+//  auto* pAnalogInput = new ESP32AnalogInput(34, ADC_WIDTH_BIT_9, ADC_ATTEN_DB_0, 1000);
+//  auto* pAnalogInput = new ESP32AnalogInput(34, ADC_WIDTH_BIT_12, ADC_ATTEN_DB_0, 1000);
+//  auto* pAnalogInput = new ESP32AnalogInput(34, ADC_WIDTH_BIT_9, ADC_ATTEN_DB_11, 1000);
+  
+  pAnalogInput->connectTo(new SKOutputNumber("voltage")); 
+
+
+  sensesp_app->enable();
+  
+});
--- a/src/sensors/esp32_analog_input.cpp
+++ b/src/sensors/esp32_analog_input.cpp
@@ -1,0 +1,87 @@
+#include "esp32_analog_input.h"
+
+#include "Arduino.h"
+#include "sensesp.h"
+#include <RemoteDebug.h>
+
+ESP32AnalogInput::ESP32AnalogInput(uint8_t pin, adc_bits_width_t adc_width, 
+                                   adc_atten_t adc_attenuation,
+                                   uint read_delay, String config_path)
+
+    : NumericSensor(config_path), pin{pin}, read_delay{read_delay} {
+
+  associatePin(pin, adc_width, adc_attenuation);
+  className = "ESP32AnalogInput";
+  load_configuration();
+}
+
+void ESP32AnalogInput::associatePin(int pin, adc_bits_width_t adc_width, 
+                                    adc_atten_t adc_attenuation) {
+
+  // Configure width across all channels on ADC1
+  adc1_config_width(adc_width);
+  
+  // Set channel attenuation
+  adc1_channel = (adc1_channel_t) digitalPinToAnalogChannel(pin);
+  adc1_config_channel_atten(adc1_channel, adc_attenuation);
+
+  // Characterize/calibrate the ADC
+  cal_type = esp_adc_cal_characterize(ADC_UNIT_1, adc_attenuation, 
+                                      adc_width, V_REF, &characteristics);
+  
+  if (cal_type == ESP_ADC_CAL_VAL_EFUSE_VREF) {
+      cal_type_str = "eFuse Vref";
+  } else if (cal_type == ESP_ADC_CAL_VAL_EFUSE_TP) {
+      cal_type_str = "Two Point";
+  } else {
+      cal_type_str = "Default";
+  }
+
+  // Setup the channel for calling the readVolts function
+  adc_channel = (adc_channel_t) digitalPinToAnalogChannel(pin);
+  
+  configured = true;
+}
+
+
+float ESP32AnalogInput::readVolts() {
+	uint32_t mV = 0;
+	esp_adc_cal_get_voltage(adc_channel, &characteristics, &mV); // returns mV
+	return  mV*0.001;
+}
+
+void ESP32AnalogInput::update() {
+  //output = analogRead(pin);
+  output = readVolts();
+  this->notify();
+}
+
+void ESP32AnalogInput::enable() {
+  app.onRepeat(read_delay, [this]() { this->update(); });
+}
+
+void ESP32AnalogInput::get_configuration(JsonObject& root) {
+  root["read_delay"] = read_delay;
+  root["value"] = output;
+};
+
+static const char SCHEMA[] PROGMEM = R"###({
+    "type": "object",
+    "properties": {
+        "read_delay": { "title": "Read delay", "type": "number", "description": "Number of milliseconds between each analogRead(32)" },
+        "value": { "title": "Last value", "type" : "number", "readOnly": true }
+    }
+  })###";
+
+String ESP32AnalogInput::get_config_schema() { return FPSTR(SCHEMA); }
+
+bool ESP32AnalogInput::set_configuration(const JsonObject& config) {
+  String expected[] = {"read_delay"};
+  for (auto str : expected) {
+    if (!config.containsKey(str)) {
+      return false;
+    }
+  }
+  read_delay = config["read_delay"];
+  return true;
+}
--- a/src/sensors/esp32_analog_input.h
+++ b/src/sensors/esp32_analog_input.h
@@ -1,0 +1,118 @@
+/*
+This sensor outputs a voltage unlike the AnalogInput which ouputs a raw ADC value 
+
+This sensor is required to deal with the specific setup of the ESP32 ADC's.
+
+The decisions are:
+1) Whether to use ADC1 or ADC2.
+
+This code does not support ADC2 because the use of ADC2 requires interaction with the WiFi 
+code as ADC2 is shared with the wifi driver.
+
+The pins associated with ADC1 are 32 to 39.
+
+If the internal Hall sensor is used then pins 36 & 39 should not be used. 
+Pins 36 & 39 are also associated with the power switching of ADC2 for wifi and will have 
+short "pull down" events of 89nS. While this is probably not critical to most applications 
+its advisable to only use these pins if no others are available.
+
+2) The next decision is what resolution is required for your application. The resolution 
+ranges from 9 bits (512) to 12 bits (4096). In most cases the highest resolution would be 
+applicable. The default is 12 bits.
+
+Note that the ESP32-S2 allows only a 13 bit (8192) resolution setting.
+!! NB When using the ESP32-S2 the default of 12 bits must be overidden with 
+   ADC_WIDTH_13Bit !!
+
+3) The most critical consideration is the attenuation option. The ESP32 ADC's allow the user 
+to apply an attenuator in order to increase the "full-scale range" of the input from 1.1v to
+3.9V.
+
+This obviously impacts on the resolution of the reading. For example, if your input is never 
+going to exceed 950mV then you would set for zero attenuation. The selected resolution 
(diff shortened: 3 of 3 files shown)


src/sensors/esp32_analog_input.h
```diff
@@ -0,0 +1,118 @@
+/*
+This sensor outputs a voltage unlike the AnalogInput which ouputs a raw ADC value 
+
```
ba58smith, line 3: If it specifically outputs a voltage, should it be called esp32_analog_voltage.*, and ESP32AnalogVoltage? Would it ever be used to ouput something other than voltage? What if I want to use an ESP32 to measure the output of an analog light sensor, which I want to convert to a 0 - 100 range representing pitch black and bright light? What Class should I use to do that? If not this one - because it's not trying to output a voltage measurement - then it seems like this is ONLY used for voltage, and should be named accordingly. But I may be missing something.

src/sensors/esp32_analog_input.h
```diff
@@ -0,0 +1,118 @@
...
+#define V_REF 1100
+
+class ESP32AnalogInput : public NumericSensor {
+
+public:
+  ESP32AnalogInput(uint8_t pin = 32, adc_bits_width_t adc_width = ADC_WIDTH_12Bit, 
+              adc_atten_t adc_attenuation = ADC_ATTEN_11db, uint read_delay = 200, 
+              String config_path = "");
+  void enable() override final;
+
```
ba58smith, line 94: Please format all the members, below, so they line up nicely on the same left margin.

src/sensors/esp32_analog_input.cpp
```diff
@@ -0,0 +1,87 @@
...
+                                      adc_width, V_REF, &characteristics);
+  
+  if (cal_type == ESP_ADC_CAL_VAL_EFUSE_VREF) {
+      cal_type_str = "eFuse Vref";
+  } else if (cal_type == ESP_ADC_CAL_VAL_EFUSE_TP) {
+      cal_type_str = "Two Point";
+  } else {
+      cal_type_str = "Default";
+  }
+
```
ba58smith, line 39: What's the purpose of setting cal_type, and cal_type_str? I don't see that either of them are used for anything other than the above code, which seems to be for the sole purpose of setting cal_type_str.

src/sensors/esp32_analog_input.cpp
```diff
@@ -0,0 +1,87 @@
...
+  // Setup the channel for calling the readVolts function
+  adc_channel = (adc_channel_t) digitalPinToAnalogChannel(pin);
+  
+  configured = true;
+}
+
```
ba58smith, line 45: It doesn't look like "configured" is ever used for any purpose. Is it supposed to be?

src/sensors/esp32_analog_input.cpp
```diff
@@ -0,0 +1,87 @@
...
+	return  mV*0.001;
+}
+
+void ESP32AnalogInput::update() {
+  //output = analogRead(pin);
+  output = readVolts();
+  this->notify();
+}
+
+void ESP32AnalogInput::enable() {
```
ba58smith, line 59: I don't see any reason to leave the comment - //output = analogRead(pin); - in the code, do you?

src/sensors/esp32_analog_input.cpp
```diff
@@ -0,0 +1,87 @@
...
+	esp_adc_cal_get_voltage(adc_channel, &characteristics, &mV); // returns mV
+	return  mV*0.001;
+}
+
+void ESP32AnalogInput::update() {
+  //output = analogRead(pin);
+  output = readVolts();
+  this->notify();
+}
+
```
ba58smith, line 58: I don't see any reason to leave the comment - //output = analogRead(pin); - in the code, do you?

src/sensors/esp32_analog_input.cpp
```diff
@@ -0,0 +1,87 @@
...
+  configured = true;
+}
+
+
+float ESP32AnalogInput::readVolts() {
+	uint32_t mV = 0;
+	esp_adc_cal_get_voltage(adc_channel, &characteristics, &mV); // returns mV
+	return  mV*0.001;
+}
+
```
ba58smith, line 52: Can you not just:
`return esp_adc_cal_get_voltage(adc_channel, &characteristics, &mV) * 0.001`? And if you can, then you really don't need readVolts() - ::update() can be just:
`output = esp_adc_cal_get_voltage(adc_channel, &characteristics, &mV) * 0.001`. Makes it much simpler.

examples/esp32analog_input.cpp
```diff
@@ -0,0 +1,56 @@
...
+  debugI("Serial debug enabled");
+
+  // Create the global SensESPApp() object. 
+  // If you add the line ->set_wifi("your ssid", "your password") you can specify
+  // the wifi parameters in the builder. If you do not do that, the SensESP 
+  // device wifi configuration hotspot will appear and you can use a web 
+  // browser pointed to 192.168.4.1 to configure the wifi parameters. 
+  // You can use NONE, UPTIME, FREQUENCY, FREE_MEMORY, WIFI_SIGNAL or ALL 
+  // instead of IP_ADDRESS in set_standard_sensors().
+
```
ba58smith, line 31: A couple days ago, I added a lot of detail to the wiki about using the "builder", and the "Standard Sensors". Rather than putting a lot of details in this file, you could simply refer to the Wiki, which will be updated as these things change.

mairas:
This PR implements a new sensor, against my explicit guidance on the #sensors Slack channel. Instead, as I told in the discussion, the existing `AnalogInput` class should have been modified to support the ESP32 specific functionality. By providing sensible defaults, even the existing function signatures could be used as is.

Any new code should follow the Google C++ Style Guide as described on the SensESP wiki: https://github.com/SignalK/SensESP/wiki/Contributing-to-the-SensESP-Project

At least some effort should be made to provide decent Git commits. As it is, the messages for two commits are "Add files via upload" and one states the name of the file being modified.

Since implementing the changes would imply changing everything in the PR, I'm closing this PR for now. Please open a new one once you have implemented the changes in the correct files.

GaryWSmith:
I advised that I was a part-timer and that I would need support when it was inferred/stated that I was demanding volunteers to do heavy work. 
This is my first use of git for collaborative work. Even though the two git comments are actually correct - this is the addition of two new files.
I missed the explicit advice in the sensor channel, but even then I don't why this is a problem when the solution is in fact the correct one - the difference between the two is fundamental. 
I provided a lengthy explanation in the header file to back this up.
I am happy to make the adjustments suggested by Brian and to format the code appropriately.

But as it is, I do not believe that you understand the implications of trying to make the existing analog_input unit work for the ESP32. You seem to still believe that it's just a matter of multipliers? I suggest you read the documentation relating to the characterisation call and the calibration type. 

I am assuming that you would insist on having the same type of output for both sensor types - the raw adc output?
How would you accomodate the impact of the attenuation setting in the voltage_divider transport which does not appear from the documentation to have a linear impact? They would not go to the effort of providing a calibration/characterisation call if it was not required - despite your assertions. 

If you still feel that this could/should be handled by compiler directives then I'm afraid that's a step too far for me, and a solutin which overly complicates the analog_input structure and goes against the stated goal of keeping code atomic.

This solution is clear and succinct.

Please advise whether you would accept correctly formatted ESP32_analog_input files.

examples/esp32analog_input.cpp
GaryWSmith: I shall do that. 
This was just a copy from the original temperature_sender example.

I had intended to update the wiki and also do PR for some work on the README file.

Thanks for you helpful feedback. I appreciate it.

src/sensors/esp32_analog_input.cpp
GaryWSmith: Yep, that makes sense. I'll make the changes.

src/sensors/esp32_analog_input.cpp
GaryWSmith: Nope, it was part of my debugging / investigation process.
I'll remove it.

src/sensors/esp32_analog_input.cpp
GaryWSmith: Another artifact from the debugging process.

src/sensors/esp32_analog_input.cpp
GaryWSmith: They are related to my question in the channel about the abiliity to read the outputs of the transports. 

I used them with debugD() to ensure that I understood the characterisation API call. I left them in as I thought the might be useful for anyone using and debugging their examples down the line and understanding what type of calibration/characterisation their board supports. But essentaily they don't add anything directly to the code so I'll remove them.

src/sensors/esp32_analog_input.h
GaryWSmith: Will do

src/sensors/esp32_analog_input.h
GaryWSmith: At the end of the day the input to any of the analog ADC inputs is a voltage.

The output of the existing unit is a raw 0-1023 value which is then fed to AnalogVoltage to get back to the voltage value. This is a straight ratio calculation. But essentially you need to get back to a voltage before you can move forward.

After the existing AnalogInput you would use the VoltageDivider to determine the resistance of the sensor - or you just simply use the voltage depending on what the sensor type / objective is.

No matter what sensor you are using it must be varying the voltage as seen by the input pin either through voltage dividers or directly from active components.

This unit avoids the need for the AnalogVoltage unit. But it is not a mere convenience. 

The ESP32 does more than that. It has a number of resolution options and if that was the only difference then it could still be catered for in the AnalogVoltage unit. 

But it also has an attenuation setting. So the AnalogVoltage unit would need to be modified for a possible combination of  5 resolution settings and 4 attenuation settings - none of which is required for the ESP8266 apparently. Do you want to add this complexity to an existing unit?

But that is not the end of it either.

The relationship between attenuation, resolution, vref and adc output is NOT LINEAR. Which is why they provide the characterisation call which is necessary before any reads are called. I don't know how you would achieve that from the AnalogVoltage unit???     

It seems that Matti doesn't believe their documentation despite my experience where I show a 20% difference at one observed voltage and observations of divergence from linear multipliers.

Personally, I trust the documentation.

JohnySeven:
@GaryWSmith maybe it would be helpful if you can write down diference between analog_input.cpp and ESP32_analog_input.cpp? I as user of SensESP will have to decide whether I use the normal analog input or ESP32 specific input, and it's just little odd to have two analog inputs in first place. Also if you can replace one or the other why don't we use yours on ESP32 and the other on ESP8266?

This is what came to my mind when I saw this PR.

Thanks for clarification!

GaryWSmith:
@JohnySeven 

I've shown that the existing analog input does not produce the expected results on the ESP32 (nor for that matter does the bulk of Arduino code which also neglects the ESP32 requirements). In my case I had a 20% difference between using the existing unit and the code provided esp-idf at my "reference voltage". 

Not only is there a difference which can only be characterised by observation but correcting for the difference with multipliers did not work as I started to see divergences as the input moved away from the reference voltage.

The differences are: 
The ESP32 has multiple resolution settings: 9-13 bits
The ESP32 has 4 attenuation settings: 0dB to 11dB
The ESP32 REQUIRES a "characterisation" call which characterises the ADC with the selected resolution, attenuation and the eFuse 1 point or two coefficient values if they are present. That is, the relationship between raw ADC readings and output is NOT LINEAR and undefined without the characterisation call.

The ESP32AnalogInput produces an accurate output without the use of "opaque" multipliers which would be confusing to most users.

I provided the explanation and more detail in the header file and would have made a PR for an updated RERADME as well as updating the wiki.

GaryWSmith:
@JohnySeven 

Jan, Regarding your question of whether the ESP32 unit would suffice for the 8266. 

I am completely unfamiliar with the 8266. I must accept that whoever authored that unit would have applied any characterisation call type requirements if they existed.

I suppose that both sets of code could be gathered into one unit and compiler directives used to select the correct configuration. But do you want excess / irrelevant input parameters to be floating around?

To me the simplest, cleanest most succinct solution is the dedicated ESP32 unit.  
I don't think that

mairas:
@GaryWSmith please do not characterise what I understand and what I don't.

From a software architecture point of view, you want to define clear APIs which abstract away the hardware differences. You don't add new interfaces at will. So, for something that reads MCU native analog inputs, it will be the AnalogInput class. 

At the moment AnalogInput only returns the raw input values but it is trivial to add an output scaling coefficient as a constructor argument which will then allow the user to define the output range. As for the different settings for the ESP32 ADC, you'll be just fine by selecting reasonable defaults. There is no need to expose all the details of the ADC to the user. Once that is done, the internal implementation for the different MCUs can differ, but the programming interface remains the same.

ba58smith:
@GaryWSmith and I are going to collaborate on this, and I hope we'll come up with an elegant solution that will be more than satisfactory to all concerned. Now... everyone, be nice.

GaryWSmith:
@mairas 

I understand that my initial post(s) may have caused some concern. This was certainly not my intention. I can only state that I was quite surprised to find that this obvious flaw had not been addressed earlier. That sentiment no doubt came through. When it explained that I was one of the first users of the ESP32 it became clear.

I saw comments that I was demanding something. This is not close to the truth at all. At that stage I considered myself in the user/consumer role - not at all comfortable to delve into the lower level code. I offered my assistance as best I could envisage. At that point I was also having build issues which also impacted on my view of my ability to assist. With feedback from Brian and looking at the unit code I realised that I could in fact contribute but made it clear that I was a "part-timer" and would need assistance. 

I trust that this context clears up any misconceptions about my motive or attitude. 

Matti wrote:
"At the moment AnalogInput only returns the raw input values but it is trivial to add an output scaling coefficient as a constructor argument which will then allow the user to define the output range. As for the different settings for the ESP32 ADC, you'll be just fine by selecting reasonable defaults. There is no need to expose all the details of the ADC to the user. Once that is done, the internal implementation for the different MCUs can differ, but the programming interface remains the same."

Unfortunately that is not correct. The documentation is clear that an  "ADC characterisation" call is required. The call takes as inputs the attenuation, resolution and Vref (user provided if the eFuse alternatives are not available).

In order to calculate the voltage form the raw adc reading the characterisation variable must be applied and this call used:
uint32_t voltage = esp_adc_cal_raw_to_voltage(reading, adc_chars);

It is NOT a case of using the ratio of the ADC reading / resolution * full scale voltage.

**So if you wish to limit the output to match that of the existing unit then you must devise a method of carrying the characterisation variable through to any unit which wishes to determine the CORRECT input voltage.**

Also, I wrote a fairly lengthy explanation of the selection of the appropriate attenuation values. I placed it in the header file - probably incorrectly by the guidelines but intended to add this to the wiki and README

GaryWSmith:
@mairas 

This is why you don't want to minimise the significance of the attenuation value. In my test example I showed that the difference was notable

3) The most critical consideration is the attenuation option. The ESP32 ADC's allow the user 
to apply an attenuator in order to increase the "full-scale range" of the input from 1.1v to
3.9V.

This obviously impacts on the resolution of the reading. For example, if your input is never 
going to exceed 950mV then you would set for zero attenuation. The selected resolution 
would apply to the range corresponding with the selected attenuation.

If your range exceeds 950mV then you have the option to apply attenuation as per the diagram 
below. The ability to set attenuation also may have an impact on the accuracy of your 
readings as per this diagram:

    +----------+------------+--------------------------+
    |   SoC    | attenuation|   suggested range (mV)   |
    +==========+============+==========================+
    |          |  0dB       |     100 ~ 950            |
    |          +------------+--------------------------+
    |          |  2.5dB     |     100 ~ 1250           |
    |   ESP32  +------------+--------------------------+
    |          |  6dB       |     150 ~ 1750           |
    |          +------------+--------------------------+
    |          |  11dB      |     150 ~ 2450           |
    +----------+------------+--------------------------+
    |          |  0dB       |     100 ~ 800            |
    |          +------------+--------------------------+
    |          |  2.5dB     |     100 ~ 1100           |
    | ESP32-S2 +------------+--------------------------+
    |          |  6dB       |     150 ~ 1350           |
    |          +------------+--------------------------+
    |          |  11dB      |     150 ~ 2600           |
    +----------+------------+--------------------------+

The default is 11db attenuation (ADC_ATTEN_DB_11)

The diagram above indicates that using the portion of the range indicated is highly
recommended. The careful selection of reference/volatage divider resistors in conjunction 
with attenuation settings can greatly improve the accuracy of the ESP32 analog input.

The main difference between ESP32 and ESP8266 is the requirement to "characterise" the ADC.
This is essentially an internal calibration of the ADC using the selected ADC resolution, 
selected ADC attenuation and the voltage reference which can be a nominal 1100mV value, a
single point coefficient value stored in eFuse memory or a two point reference coefficients 
also stored in eFuse memory. A calibration call must be made prior to calling any read 
functions.

The lack of applying the characterisation of the ADC has greatly contributed to the bad rap 
of the ESP32 for analog input.

mairas:
Gary, there is nothing preventing you implementing any possible configuration routines within the existing framework. That is not difficult. And that's what I referred to by having different implementations for the different MCUs. Again, do not underestimate my understanding on the topic.

Anyway. I'm not going to quarrel further on this topic. I'm putting strong emphasis on code maintainability and project quality, and no PR will get merged by simply arguing with the maintainers.

GaryWSmith:
Perhaps you can help me out then.

How do I access the characterisation variable in AnalogVoltage transport?

mairas:
You don't. The `AnalogVoltage` transform is irrelevant and basically a redundant version of the `Linear` transform. Read what I wrote above. You instead add an output scaling coefficient in `AnalogInput`.

GaryWSmith:
So you ignore the characterisation variable?

GaryWSmith:
Or are you saying that you ignore the clearly documented requirement to perform a characterisation call prior to making a read?

GaryWSmith:
The only way that I can see this working is:

Call the charcterisation function
Read the voltage correctly with the 
Assume an arbitrary full scale voltage (another input parameter?)
Determine the linear "adc value" from the ratio * the selected resolution value
Pass the derived linear adc value out

Then in the next unit calculate the voltage by reversing the above. 

That would work but it just feels so wrong ... The full-scale voltage is an arbitrary value and yet another input parameter.

PR #186, opened 2020-10-06
ESP32 support for AnalogInput

mairas:
This class implements ESP32 support for AnalogInput. Some sensible defaults are used at the moment for attenuation and bit width values.

Tested both on ESP8266 and ESP32.

### DIFF
--- a/src/sensors/analog_input.cpp
+++ b/src/sensors/analog_input.cpp
@@ -3,19 +3,25 @@
 #include "Arduino.h"
 #include "sensesp.h"
 
-AnalogInput::AnalogInput(uint8_t pin, uint read_delay, String config_path)
-    : NumericSensor(config_path), pin{pin}, read_delay{read_delay} {
-  pinMode(pin, INPUT);
+AnalogInput::AnalogInput(uint8_t pin, uint read_delay, String config_path,
+                         float output_scale)
+    : NumericSensor(config_path),
+      pin{pin},
+      read_delay{read_delay},
+      output_scale{output_scale} {
+  analog_reader = new AnalogReader(pin);
   load_configuration();
 }
 
 void AnalogInput::update() {
-  output = analogRead(pin);
+  output = output_scale * analog_reader->read();
   this->notify();
 }
 
 void AnalogInput::enable() {
-  app.onRepeat(read_delay, [this]() { this->update(); });
+  if (this->analog_reader->configure()) {
+    app.onRepeat(read_delay, [this]() { this->update(); });
+  }
 }
 
 void AnalogInput::get_configuration(JsonObject& root) {
--- a/src/sensors/analog_input.h
+++ b/src/sensors/analog_input.h
@@ -1,22 +1,24 @@
 #ifndef _analog_input_H_
 #define _analog_input_H_
 
+#include "analog_reader.h"
 #include "sensor.h"
 
 class AnalogInput : public NumericSensor {
-
-public:
-  AnalogInput(uint8_t pin = A0, uint read_delay = 200, String config_path = "");
+ public:
+  AnalogInput(uint8_t pin = A0, uint read_delay = 200, String config_path = "",
+              float output_scale = 1024.);
   void enable() override final;
 
-private:
+ private:
   uint8_t pin;
   uint read_delay;
+  float output_scale;
+  BaseAnalogReader* analog_reader;
   virtual void get_configuration(JsonObject& doc) override;
   virtual bool set_configuration(const JsonObject& config) override;
   virtual String get_config_schema() override;
   void update();
 };
-
 
 #endif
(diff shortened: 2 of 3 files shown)


src/sensors/analog_input.cpp
```diff
@@ -1,21 +1,28 @@
 #include "analog_input.h"
 
 #include "Arduino.h"
+#include "analog_reader.h"
 #include "sensesp.h"
```
ba58smith, line 6: Do you need to` #include "analog_reader.h"` here, since it's also in the .h file?

ba58smith:
In anticipation of someone asserting that the results of this aren't as accurate as they could be, because it doesn't use the actual measured VREF of the chip, consider two things:
1. The ADC on the ESP32 is a very cheap "instrument" that is prone to a lot of noise, and at its very best, it's still not super accurate, and its output is not linear.
2. A chip with a measured VREF that is off from the norm (1100 mv) as much as possible is off by only 100 mv. (Because the range is from 1000 to 1200.) This graph from the ESP32 documentation shows how little that impacts the results:
image

Bottom line: if you want super accurate, use an external ADC, such as an ADS1015 / ADS1115, which is already supported by SensESP. But consider what you're using this for before you do that - the built-in ADC may be more than accurate enough for your purposes. (The ADC on the ESP8266 is also not especially accurate, BTW.)

GaryWSmith:
In response to the "anticipation of someone asserting" ... :-)

Brian, after our lengthy discourse on the topic, you surely realise by now that the graph you present is only ONE element of the bias and non-linearities that the manufacturers API adc characterisation call addresses?

You also realise that the manufacturers documentation clearly explains that the use of the attenuation setting allows the user to enhance the accuracy of the device where appropriate?

I am intrigued that you use the supposed inherent inaccuracy of the device as justification to implement a solution which contributes to the the degradation of the results in at least some instances.

Do you have any data which demonstrates the accuracy or lack thereof of the device? I ask this because I am seeing results within 3 mV of measurements performed by a quality Fluke multimeter - **when the device is used as per the manufacturers guidelines** - multi-sampling, a capacitor and most importantly, **configured as per the manufacturers guidelines**.

**Bottom line: if you want the best results from any hardware it is always sensible to follow their usage guidelines UNTIL/UNLESS you have incontrovertible data which proves that you know the product better than the manufacturer.**

This will be my last input on the subject.

Once again, I am very impressed with this project - except for this single issue.  Please don't infer more into my stance than that I simply wish to contribute in a positive manner.

src/sensors/analog_input.cpp
mairas: Sharp eyes! Will fix.

ba58smith:
I deleted my comment about not using the eFuse calibration because I was wrong about that. This implementation does use the eFuse calibration if it's available on the chip - that's inherent in the call to `esp_adc_cal_characterize`.

mairas:
> In response to the "anticipation of someone asserting" ... :-)
> 
> Brian, after our lengthy discourse on the topic, you surely realise by now that the graph you present is only ONE element of the bias and non-linearities that the manufacturers API adc characterisation call addresses?

Give it a rest already... If you'd have read the [espressif adc implementation](https://github.com/espressif/esp-idf/blob/f91080637c054fa2b4107192719075d237ecc3ec/components/esp_adc_cal/esp_adc_cal.c), you'd understand that their calibration function is a simple two-point linear transformation:

```
static uint32_t calculate_voltage_linear(uint32_t adc_reading, uint32_t coeff_a, uint32_t coeff_b)
{
    //Where voltage = coeff_a * adc_reading + coeff_b
    return (((coeff_a * adc_reading) + LIN_COEFF_A_ROUND) / LIN_COEFF_A_SCALE) + coeff_b;
}
```

That's ALL that happens, and that's not all. `coeff_b` is non-zero only if the user has saved the two-point calibration values in the eFuses themselves. So, your fancy bias and non-linearity manufacturer characterisation code boils down to a simple multiplication by one correction coefficient. And that's it.

mairas:
OK, I also wrongly stated that the calibration would be a simple one-coefficient multiplication, but there was a 20-point lookup table enabled, after all.

mairas:
Removed the extra include. Merging.

GaryWSmith:
Too late for this PR but I recommend another ifdef for the ESP32-2S. 

It supports **only** a 13 bit resolution. They probably aren't that common yet but I assume that you don't want to have to revisit this later. 

"enum adc_bits_width_t

    ADC resolution setting option.

    Note

        For ESP32-S2. Only 13 bit resolution is supported. For ESP32. 13 bit resolution is not supported.
"

ba58smith:
@GaryWSmith - if that's the only modification for the ESP32-2S, I think you can wrap the whole ESP32AnalogReader class, with `#ifdef(ESP32) || defined(ESP32-2S)`. Then, use #ifdef and #else around only the line that defines the resolution: ```#ifdef(ESP32-2S)
adc_bits_width_t bit_width = ADC_WIDTH_BIT_13;
#else
adc_bits_width_t bit_width = ADC_WIDTH_BIT_12;
#endif```

PR #189, opened 2020-10-06
Added new Numeric Transform - LogFunction

GaryWSmith:
This Transform applies the log function to an input i.e. output = (multiplier * LN(input) + offset)
It has been copied directly from a similar Transform so hopefully it meets all of the guidelines.
It is useful as a simple method of characterising NTC thermistors although being generic it has many potential uses.

### DIFF
--- a/src/transforms/logfunction.cpp
+++ b/src/transforms/logfunction.cpp
@@ -1,6 +1,6 @@
 #include "logfunction.h"
 
-LogFunction::LogFunction(float constant_1, float constant_2,
+LogFunction::LogFunction(float multiplier, float offset,
                              String config_path)
     : NumericTransform(config_path),
       multiplier{multiplier},
--- a/src/transforms/logfunction.cpp
+++ b/src/transforms/logfunction.cpp
@@ -23,8 +23,8 @@ void LogFunction::get_configuration(JsonObject& root) {
 static const char SCHEMA[] PROGMEM = R"({
     "type": "object",
     "properties": {
-        "constant_1": { "title": "multiplier", "type": "number", "description": "The (m)ultiplier in the equation f(x) = m*log(x) + o" },
-        "offset": { "title": "offset", "type": "number", "description": "The (o)ffset in the equation f(x) = m*log(x) + 0" },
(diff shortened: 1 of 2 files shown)


src/transforms/logfunction.cpp
```diff
@@ -0,0 +1,44 @@
...
+
+LogFunction::LogFunction(float constant_1, float constant_2,
+                             String config_path)
+    : NumericTransform(config_path),
+      multiplier{multiplier},
+      offset{offset}{
+  className = "LogFunction";
+  load_configuration();
+}
+
```
ba58smith, line 11: I think you mean for constant_1 and constant_2 to be multiplier and offset, don't you?

src/transforms/logfunction.cpp
```diff
@@ -23,8 +23,8 @@ void LogFunction::get_configuration(JsonObject& root) {
 static const char SCHEMA[] PROGMEM = R"({
     "type": "object",
     "properties": {
-        "constant_1": { "title": "multiplier", "type": "number", "description": "The (m)ultiplier in the equation f(x) = m*log(x) + o" },
-        "offset": { "title": "offset", "type": "number", "description": "The (o)ffset in the equation f(x) = m*log(x) + 0" },
```
ba58smith: I just noticed this for the first time: right above this comment, shouldn't "constant_1" be "multiplier"? Sorry I didn't catch it before.

ba58smith:
I also just noticed that you created this PR on your "master" branch, rather than on a specific branch created for this PR. As noted in the wiki: 

"For your own benefit, it is recommended that you create a new Git branch for each PR. It doesn't necessarily change the end result, but it's very likely that you'll be asked to make some changes in the PR, and if you're working out of your master branch, you might end up in a world of pain. Hence, make a unique branch for each PR. ***If you don't already have a Git process that works perfectly, and that includes making a new branch for each PR, then please follow the steps below exactly.***"

So, please put this PR into a new branch, rather than have it in "master".

GaryWSmith:
I did, thank you Brian.

GaryWSmith:
Re the specific branch - I'll close this and try again.

PR #190, opened 2020-10-06
Added LogFunction Transform

GaryWSmith:
A simple LogFunction Transform

### DIFF
--- a/src/transforms/logfunction.cpp
+++ b/src/transforms/logfunction.cpp
@@ -23,8 +23,8 @@ void LogFunction::get_configuration(JsonObject& root) {
 static const char SCHEMA[] PROGMEM = R"({
     "type": "object",
     "properties": {
-        "multiplier": { "title": "multiplier", "type": "number", "description": "The (m)ultiplier in the equation f(x) = m*log10(x) + 0" },
-        "offset": { "title": "offset", "type": "number", "description": "The (o)ffset in the equation f(x) = m*log10(x) + 0" },
+        "multiplier": { "title": "multiplier", "type": "number", "description": "The (m)ultiplier in the equation f(x) = m*log10(x) + o" },
+        "offset": { "title": "offset", "type": "number", "description": "The (o)ffset in the equation f(x) = m*log10(x) + o" },
         "value": { "title": "Last value", "type" : "number", "readOnly": true }
     }
   })";
--- a/src/transforms/logfunction.cpp
+++ b/src/transforms/logfunction.cpp
@@ -23,7 +23,7 @@ void LogFunction::get_configuration(JsonObject& root) {
 static const char SCHEMA[] PROGMEM = R"({
     "type": "object",
     "properties": {
-        "constant_1": { "title": "multiplier", "type": "number", "description": "The (m)ultiplier in the equation f(x) = m*log10(x) + o" },
+        "multiplier": { "title": "multiplier", "type": "number", "description": "The (m)ultiplier in the equation f(x) = m*log10(x) + 0" },
         "offset": { "title": "offset", "type": "number", "description": "The (o)ffset in the equation f(x) = m*log10(x) + 0" },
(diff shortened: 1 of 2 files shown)


GaryWSmith:
I trust that I have done this correctly this time :-)

src/transforms/logfunction.cpp
```diff
@@ -0,0 +1,44 @@
...
+
+static const char SCHEMA[] PROGMEM = R"({
+    "type": "object",
+    "properties": {
+        "constant_1": { "title": "multiplier", "type": "number", "description": "The (m)ultiplier in the equation f(x) = m*log10(x) + o" },
+        "offset": { "title": "offset", "type": "number", "description": "The (o)ffset in the equation f(x) = m*log10(x) + 0" },
+        "value": { "title": "Last value", "type" : "number", "readOnly": true }
+    }
+  })";
+
```
ba58smith, line 31: Shouldn't `constant_1` be `multiplier`? (Same pattern as the "offset" line right below it.)

GaryWSmith: Yep, thanks. 

I fixed this via the edit function on the github site. Will that pull through to the pull request?

ba58smith: I don't know - I've never done that. But there's no new commit here, so maybe not? I think this is the correct way:
https://github.com/SignalK/SensESP/wiki/Contributing-to-the-SensESP-Project#update-an-existing-pull-request-before-it-is-merged

ba58smith: It looks like it did - see the two new commits below? Those are the fixes, I think.

GaryWSmith: I've only ever personally used github as a repository for my own stuff - never for collaboration, hence the lack of comments on the first attempt and subsequent failings.

mairas:
Hey. Content-wise this looks as good as it gets, thanks for that! I just have another issue with these simple transforms you unwittingly became a victim of: I'm concerned having a lot of these very simple transforms because they have 70-something lines of almost unchanged boilerplate for just a couple of lines of actual code. This reduces maintainability because any change in e.g. the configuration marshalling code would need to be repeated a couple of dozen times.

Instead, I'm planning to implement a `LambdaTransform` template class soon. In your case, you could then implement the new transform straight in `main.cpp` as follows:

```
auto* analog_input =
    new AnalogInput(pin, read_delay, analog_in_config_path, output_scale);
auto* log_function = [](float input, float multiplier, float offset) -> float {
  return (multiplier * log(input)) + offset;
};
// template types below are:
// 1. input
// 2. output
// 3. value_1 (multiplier)
// 4. value_2 (offset)
auto* log_transform = new LambdaTransform<float, float, float, float>(
  log_function, "multiplier", "offset", "/config/path");
analog_input
  ->connect_to(log_transform, def_multiplier, def_offset)
  ->connect_to(new SKOutputNumber(sk_path));
```

This will define a new function called `log_function` - you could do it with a regular function definition, too:
```
float log_function(float input, float multiplier, float offset) {
  return (multiplier * log(input)) + offset;
}
```
and then create a new transform that calls the function to do its thing. The syntax is a bit more involved (and will likely be somewhat different when it's actually done) but the end result will be that simple transforms can be defined on the go, without having to create new classes for it. So, instead of merging this PR and adding yet another public transform class that needs to be maintained forever, I'm asking for a bit patience until I've managed to implement the lambda transform class. Of course, you can keep using your local implementation forever, it won't be invalidated by the new developments.

GaryWSmith:
@mairas That sounds cool. 

This morning I was thinking the same thing, you could generate many function transforms and still not cover everything.
I was thinking about using an "expression evaluator" library. I used one in some basic node.js plugin.

But your approach seems far better. I look forward to it.

JohnySeven:
@mairas great stuff with lambdas, I like it! 👍

Issue #179, opened 2020-10-01
Various compiler warnings that should be dealt with


ba58smith:
1. Library Manager: Warning! More than one package has been found by OneWire requirements:
- It lists a bunch, but defaults to the first one, by "paulstoffregen". It works, and is the one we've always used, so we should specify this one: `paulstoffregen/OneWire` will always use the latest version of that library, which is what we should probably use.

2. bool HTTPClient::begin(String) is deprecated.
- This refers to the several lines in `ws_client.cpp` that call the above, with `http.begin(url)`. We need to figure out what the current signature is for `begin()` and use it instead.

3. 'SPIFFS' is deprecated: SPIFFS has been deprecated. Please consider moving to LittleFS or other filesystems.
- In `sensesp_app.cpp`, `configurable.cpp`, `spiffs_storage.cpp`.

mairas:
SPIFFS deprecation is going to be interesting. I had a quick look at LittleFS and while it seems technologically superior, every file reserves a minimum of 4 KiB of "disk space". Since we in a typical configuration only have 64 KiB to begin with, it means there's only room for 16 files. At the moment we save the sensor/transform config filenames equal to their config paths, which in SPIFFS are single files but in LittleFS would be several (I assume one directory would be one file). So, it's possible we'd easily run out of disk space unless the config saving approach gets changed. Not a showstopper but something to keep in mind.

ba58smith:
@mairas - Would we change the config saving so that there would be only one file, which would be a single JSON file, with all the JSON representing the individual configuration items (each sensor and transform, the wifi credentials, hostname, etc.) being added together to make one big nested array, or an array of nested arrays?

ba58smith:
Looking at https://arduino-esp8266.readthedocs.io/en/latest/filesystem.html#filesystem, I wonder if we really do need to worry about using too much memory. Maybe with an ESP8266 with 1 meg, but we've set the minimum to be the 4 meg version, so if it doesn't work with the 1 meg version, I wouldn't feel bad about that. And with the 4 meg version, the minimum file system size is 1 meg. That's 256 files if each one is 4K.

mairas:
A single config file might be awkward to deal with. You'd need to store all of that in memory and parsing would be more complex. But you're right; I was calculating based on 64k filesystem size and in practice it's always larger than that. On ESP8266 our default is 1M and while the currently used min_spiffs partition layout on ESP32 provides only 180K for the filesystem, it would be quite simple to customize that.

[...]

PR #191, opened 2020-10-06
Stop using deprecated HTTPClient::begin(String)

ba58smith:
Discussed w/ @mairas - we're not sure what the `WiFiClient client` does, but it allows the use of the current `HTTPClient::begin()`, and gets rid of the compiler warning about using a deprecated version (Issue #179). 

Tested with ESP8266 and ESP32 - data flows all the way to the SK Server, so it seems to work as it did before.

### DIFF
--- a/src/net/ws_client.cpp
+++ b/src/net/ws_client.cpp
@@ -227,16 +227,13 @@ void WSClient::connect() {
 
 void WSClient::test_token(const String server_address,
                           const uint16_t server_port) {
-  // Needed for HTTPClient::begin, below
-  WiFiClient client;
-  
   // FIXME: implement async HTTP client!
   HTTPClient http;
 
   String url = String("http://") + server_address + ":" + server_port +
                "/signalk/v1/api/";
   debugD("Testing token with url %s", url.c_str());
-  http.begin(client, url);
+  http.begin(wifi_client, url);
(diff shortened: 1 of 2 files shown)


ba58smith:
I've searched quite a lot and can't figure out how this works, but every example I can find of using HTTPClient does it this way. This is the closest thing I can find to an explanation: https://arduino.stackexchange.com/questions/21265/difference-between-wificlient-and-httpclient

ba58smith:
I finally found a good example of how to fix this correctly: https://github.com/earlephilhower/ESP8266Audio/commit/2d8a97767ce1e9ecec7494cb751a43b3d2fa872b

I've tested this on both ESP32 and ESP8266 - data is getting all the way to Signal K as it should. I think it's ready to merge, @mairas, but can you have a quick look?

mairas:
I've tried to look at the implementation itself to understand whether it's OK to instantiate an object multiple times, and while I've seen nothing conclusive, I'd tend to assume it's OK. And if you've tested it, let's declare it perfect. :-)

Merging.

ba58smith:
The second commit changed that "multiple instances of WiFiClient" approach. Now, `WiFiClient wifi_client` is a member of WSClient, and is used in all three instances of HTTPClient::begin(). (But the other way worked, too!)

PR #193, opened 2020-10-07
Fix bme280 config path

tedenda:
There was a missing / in config path for humidity value.


ba58smith:
Nice catch! 

Before I merge this, can you remove commit 474a05f, please? It doesn't have anything to do with this PR.

mairas:
Great!

PR #194, opened 2020-10-07
Debug function to clean up examples

tedenda:


ba58smith:
Can you describe why you did this, @tedenda? I see that it combines several lines of code into a single line, which I certainly appreciate. Is it for any other reason that I may be missing? Thanks.

tedenda:
I asked mairas today about this block of code on slack.
In almost all examples there was 2 lines outside the #ifndef statement, for no reason.
I asked if I should fix it, mairas then suggested that this block of code could move in to a function instead.
And here we are now, fixed and clean.

ba58smith:
Looks good to me, but since you talked to @mairas about it, I'll leave it for him to merge.

And I'm sure he'll want you to remove that extra commit, as you and I have talked about from your other PR.

mairas:
Yep, content-wise it looks good, but as Butch commented, could you drop the merge commit from the PRs?

I was wondering a bit would it be better to have the ifdef inside the function or around it. But I guess around it makes more sense the way you have it: it would be misleading to always have the function called but only have it do something if the definition was in place.

mairas:
It's golden now!

PR #195, opened 2020-10-07
Added the Hysteresis Transform unit files

GaryWSmith:
Classes NumericHysteresis and IntegerHysteresis.
Arguments are:
minValue - minimum value of the hysteresis range.
maxValue - maximum value of the hysteresis range.
belowRange - output value if input is smaller than the min value of the range.
aboveRange - output value if input is greater than the max value of the range.

### DIFF
--- a/examples/OneWire_temperatures.cpp
+++ b/examples/OneWire_temperatures.cpp
@@ -12,15 +12,8 @@
 
 ReactESP app([]() {
 #ifndef SERIAL_DEBUG_DISABLED
-  Serial.begin(115200);
-
-  // A small delay and one debugI() are required so that
-  // the serial output displays everything
-  delay(100);
-  Debug.setSerialEnabled(true);
+  SetupSerialDebug(115200);
 #endif
-  delay(100);
-  debugI("Serial debug enabled");
 
   sensesp_app = new SensESPApp();
 
--- a/examples/ads1x15_volt_meter.cpp
+++ b/examples/ads1x15_volt_meter.cpp
@@ -15,15 +15,8 @@
 
 ReactESP app([]() {
 #ifndef SERIAL_DEBUG_DISABLED
-  Serial.begin(115200);
-
-  // A small delay and one debugI() are required so that
-  // the serial output displays everything
-  delay(100);
-  Debug.setSerialEnabled(true);
+  SetupSerialDebug(115200);
 #endif
-  delay(100);
-  debugI("Serial debug enabled");
 
   sensesp_app = new SensESPApp();
 
--- a/examples/analog_input.cpp
+++ b/examples/analog_input.cpp
@@ -12,15 +12,8 @@ ReactESP app([]() {
 
 // Some initialization boilerplate when in debug mode...
 #ifndef SERIAL_DEBUG_DISABLED
-  Serial.begin(115200);
-
-  // A small delay and one debugI() are required so that
-  // the serial output displays everything
-  delay(100);
-  Debug.setSerialEnabled(true);
+  SetupSerialDebug(115200);
 #endif
-  delay(100);
-  debugI("Serial debug enabled");
 
   // Create the global SensESPApp() object.
   sensesp_app = new SensESPApp();
--- a/examples/bme280_example.cpp
+++ b/examples/bme280_example.cpp
@@ -48,7 +48,7 @@ ReactESP app([]() {
 
   // Do the same for the humidity value.
   auto* bme_humidity =
-      new BME280Value(bme280, humidity, read_delay, "Outside/Humidity");
+      new BME280Value(bme280, humidity, read_delay, "/Outside/Humidity");
 
   bme_humidity->connect_to(new SKOutputNumber("environment.outside.humidity"));
 
--- a/examples/bme280_example.cpp
+++ b/examples/bme280_example.cpp
@@ -12,15 +12,8 @@
 
 ReactESP app([]() {
 #ifndef SERIAL_DEBUG_DISABLED
-  Serial.begin(115200);
-
-  // A small delay and one debugI() are required so that
-  // the serial output displays everything
-  delay(100);
-  Debug.setSerialEnabled(true);
+  SetupSerialDebug(115200);
 #endif
-  delay(100);
-  debugI("Serial debug enabled");
 
   sensesp_app = new SensESPApp();
 
--- a/examples/bmp280_example.cpp
+++ b/examples/bmp280_example.cpp
@@ -12,15 +12,8 @@
 
 ReactESP app([]() {
 #ifndef SERIAL_DEBUG_DISABLED
-  Serial.begin(115200);
-
-  // A small delay and one debugI() are required so that
-  // the serial output displays everything
-  delay(100);
-  Debug.setSerialEnabled(true);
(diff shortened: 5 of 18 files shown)


ba58smith:
Can you give an example of how you intend to use this? I've looked at the code in depth, and I see what it does, but I don't know the reason for it. Thanks.

GaryWSmith:
Most of the current examples are related to reading input -> conditioning -> transforming -> send to skserver

My current applications are also concerned with control of devices 
input -> conditioning -> general transform -> send to server
                                  -> hysteresis transform -> "thermostat"

The transform is used where the boolean output purpose is to keep the input **within** a range.

In my example it effectively emulates an analogue thermostat but it would be useful in any application where you want the switching to maintain a range rather than a threshold on/off condition - such as a thermostat, pressure switch etc.

I imagine that it would be possible to use a couple of threshold transforms and a debounce but I felt that this is a more direct and "user friendly" implementation of an analogue device. Also a debounce provides hysteresis based on time, which won't always work.

ba58smith:
So what will be in your "train" of sensors and transforms AFTER this transform? For example, in the relay_control example below, a SKListener is set up for a certain path, then the value of that path is sent to the NumericThreshold, and if the value is within the defined range, then true or false is sent to DigitalOutput (depending on the 3rd parameter), to turn something on or off. In your case, I would guess you would replace the SKListener's output with the output from a Sensor, after it's been through 0 or more Transforms, then go into the HysteresisTransform, and if the value is outside the defined range, then it sends true or false (depending on the values of aboveRange and belowRange) to... what? DigitalOutput, to turn a pin HIGH or LOW?

```auto* listener = new SKNumericListener(sk_path);
  listener->connect_to(new NumericThreshold(0.0f, 100.0f, true, config_path))
      ->connect_to(new DigitalOutput(5));```

ba58smith:
A massive PR was just merged to get as much of the code as possible in alignment with the Google C++ Style Guide. I notice a lot of variables that are mixed case (likeThis), but should be lower case with underscores (like_this). Please address those issues.

Similarly, if you didn't do the automatic formatting of your two files as described in the Wiki, please do that. ("To quickly reformat existing source files in Visual Studio Code according to the preferred style, press Shift-Alt-F to apply the "Reformat Document" command.") (I can't tell if you did or didn't just by looking.)

I know these comments have nothing to do with functionality, but SensESP is getting to be quite large these days, with lot of files and lots of contributors, and if we don't all use the same naming and formatting, it's much more difficult to do code review than it needs to be. Thanks!

GaryWSmith:
Re for the question on the functionality. Yes, it's goes to a digital output which is likely used to control an "analogue" device such as a water pump, compressor heater, air con etc . Think about the typical paint spray air compressor in your garage. It runs then stops when the tank upper pressure threshold is reached. It actually overshoots that threshold slightly. It only starts up when the pressure drops to a lower threshold again with a slight overshoot.

That functionality is exactly what this transform does. While the input is within the range the input (previous output)  is passed straight through - no change. When the input is on the low side of the range the state is (say) high and the compressor is running. As it passes into the range nothing happens until it passes out of the other side in which case the output is set low (and the compressor stops). Similarly as it passes back into the range nothing happens until it hits the lower threshold again.

So it is a switch with range hysteresis rather than one based on debounce time & trial and error - hence the selection of the name.

Thinking about that example, it is possible to achieve similar functionality with 
analogInput -> 
upperThreshold (someMeaninglessLargeValue, upperthresholdVal) -> 
debounce(suitableValDeterminedFromObservation) -> 
lowerThreshold( lowerThresholdValue, someMeaninglessSmallValue) ->
debounce(suitableValDeterminedFromObservation) -> 
applianceControlOutput

Now the user needs to use one config path per threshold with one "meaningless" value per threshold, and two debounce transforms with values that are selected by observation. 

This one transform simply takes the range input and inherently deals with overshoot/debounce. Nice & clean.

Re the formatting ...

I used the ThresholdTransform as the base for this and was careful not to make any formatting changes assuming that it would therefore comply. Sadly not ...

That raises a question from me (again my inexperience with github), - how can I keep my remote fork equal with the master without deleting and re-forking? All of the guides talk about keeping my local repository current - rather than simply the keeping the github one current? 

Currently I'm waiting for resolution of this before I do that.

But I guess more pertinently I should really know whether this transform is acceptable in principle before I go wasting time on it. Is it useful?

ba58smith:
It seems useful, and is kind of the opposite of, or at least a significant twist on, the NumericThreshold Transform. Do you have a specific use in mind for it on your boat? (I'm curious, and also, it seems like any enhancement to SensESP ought to have some specific boat use, as opposed to just being a cool function.)

RE: the formatting, don't feel bad. It's going to take me a loooong time to stop naming variables likeThis, and start naming them like_this. A whole bunch of the changes made in that massive PR were changes to my code.

RE: keeping your fork updated with the main repo (I'm not saying "master", as that refers to a branch, not a repo) - let's move this discussion to Slack. I'll tag you there.

GaryWSmith:
Re the usage. I replaced a failed thermostat on my fridge with a spare themistor (from my mppt's), a relay and the ESP32 

Now I have the fridge temp in Grafana and much finer temp control than before. No more frozen eggs at the bottom and much less icing of the evaporator box. "If you can't measure it you can't manage it". :-)

ba58smith:
Great example!

mairas:
If I understood correctly, there's no state, so this is just a threshold transform with a very complex implementation. Not hysteresis. Also, most of the commits are unrelated.

OK if I close?

GaryWSmith:
Not sure that I agree with your understanding.

A threshold has a state that is determined by whether the input is in_range.

This has a state that remains unchanged while the input is in_range - which is where the hysteresis name is derived. 
When the input is out of range the state will be dependent on whether the range is hi or low of the range.  

Despite my comments earlier I realise that it is not possible to emulate this functionality with existing transforms at all.

If you consider the fridge thermostat example ... Implementing the hysteresis required to allow reasonable switching of the compressor requires the use of a debounce with a trial & error derived value which has no relationship to the condition which one is trying to control.

The selected value may work well in warm environmental conditions but could be inappropriate in cooler conditions (and vice versa) - which is exactly why I produced this transform - frozen eggs.

With the range based hysteresis implementation the "hysteresis" is directly tied to the input to be controlled.

Maybe a more apt name is RangeSwitch?

The reason that there are a number of unrelated commits is because I am trying to get my "git process" in line with the guide. So If you wish for me to resubmit it in a clean PR - that's fine by me.

mairas:
I really don't see any state whatsoever in the `::set_input` logic. This is simply a threshold function which we already have. Given the unnecessarily complex implementation, it won't be accepted. There is no need for the `NumericHysteresis` and `IntegerHysteresis` classes.

And yes, those extra commits must go, and the code must conform to the Google C++ Style Guide.

mairas:
Ah, now I get it: you're using output as state and keep outputting the same value until the trigger condition is fulfilled. I find that somewhat misleading because output is expected to be just the output. Also, output is not initialized by the constructor (except that you load the value from the saved configuration, which shouldn't be done). Use a separate state variable instead; that's much more obvious. And the implementation can be done with a few lines of code with the LambdaTransform.

Closing.

PR #196, opened 2020-10-07
Fix debug function

tedenda:
I had missed one example and missed to delete some rows.
Also tweaked the "Serial debug enabled" with new line so it is visible in terminal output.


tedenda:
Removed my commit that fixed new line.
I make a new PR for that change.

mairas:
Thanks!

PR #197, opened 2020-10-07
Rename README.md files

tedenda:
I renamed these files so they should automatically load in GitHub.


mairas:
Works on your branch. Thanks! Merging.

tedenda:
I think the readme with pictures really help to get people interested/motivated to try SenseESP sensors for themselves. Maybe something to do with other sensors also, will think about that...

PR #198, opened 2020-10-08
Define SetupSerialDebug in a cpp file

mairas:
Fixed a compile error: SetupSerialDebug was defined in the .hpp file, while it should've been just declared there and defined in the .cpp. file.

### DIFF
--- a/src/sensesp_app.cpp
+++ b/src/sensesp_app.cpp
@@ -23,6 +23,17 @@
 RemoteDebug Debug;
 #endif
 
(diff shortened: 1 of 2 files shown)


ba58smith:
I noticed that, but thought it had been done for some particular reason. @mairas , can you briefly explain why it doesn't work to have the method declared and defined in the .h file? I'm SURE I've seen that in some places in SensESP - in some cases, there is ONLY a .h file (sensesp_app_builder.h, for example). Why does it sometimes work, but not in this case?

PR #200, opened 2020-10-09
Lambda Transform

mairas:
Lambda Transform is a transform that can be defined using a transform function, without having to create a new transform class for each new transform. The transforms also allow for defining a number of configuration parameters that can be modified using the web interface.

The implementation could be cleaner: in particular, variadic template parameters could be used to remove pretty much all code duplication related to differing parameter list lengths. However, the current version of the GCC compiler provided by the Arduino Framework exposes a bug which I couldn't work around. Changing the implementation later wouldn't affect the public interface, though.

### DIFF
--- a/examples/lambda_transform.cpp
+++ b/examples/lambda_transform.cpp
@@ -1,0 +1,68 @@
+#include <Arduino.h>
+
+#include "math.h"
+#include "sensesp_app.h"
+#include "sensors/analog_input.h"
+#include "signalk/signalk_output.h"
+#include "transforms/lambda_transform.h"
+
+ReactESP app([]() {
+  SetupSerialDebug(115200);
+
+  sensesp_app = new SensESPApp("sensesp-illum-example", "My WiFi SSID",
+                               "my_wifi_password", "skdev.lan", 80, NONE);
+
+  const char* sk_path = "indoor.illumination";
+  const char* analog_in_config_path = "/indoor_illumination/analog_in";
+
+  uint read_delay = 500;
+
+#ifdef ESP8266
+  uint8_t pin = A0;
+#elif defined(ESP32)
+  uint8_t pin = 32;
+#endif
+
+  float output_scale = 3.3;
+
+  // Use AnalogInput as an example sensor. Connect it e.g. to a photoresistor
+  // or a potentiometer with a voltage divider to get an illustrative test
+  // input.
+
+  auto* analog_input =
+      new AnalogInput(pin, read_delay, analog_in_config_path, output_scale);
+
+  // This is our transform function. The example is artificial; a log transform
+  // with configurable multiplier, base, and offset parameters.
+
+  auto log_function = [](float input, float multiplier, float base,
+                         float offset) -> float {
+    return multiplier * log(input) / log(base) + offset;
+  };
+
+  // This is an array of parameter information, providing the keys and
+  // descriptions required to display user-friendly values in the configuration
+  // interface.
+
+  const ParamInfo* log_lambda_param_data = new ParamInfo[3]{
+      {"multiplier", "Multiplier"}, {"base", "Base"}, {"offset", "Offset"}};
+
+  // Here we create a new LambdaTransform objects. The template parameters
+  // (four floats in this example) correspond to the following types:
+  // 1. Output type of the transform function
+  // 2. Input type of the transoform function
+  // 3. Type of parameter 1
+  // 4. Type of parameter 2
+
+  auto log_transform = new LambdaTransform<float, float, float, float>(
+      log_function, 10, 2, 100, log_lambda_arg_data, "/transforms/log");
+      log_function, 10, 2, 100, log_lambda_param_data, "/transforms/log");
+
+  // Finally, connect the analog input via the freshly-generated log_transform
+  // to an SKOutputNumber object
+
+  analog_input->connect_to(log_transform)
+      ->connect_to(new SKOutputNumber(sk_path));
+
+  sensesp_app->enable();
+});
--- a/examples/lambda_transform.cpp
+++ b/examples/lambda_transform.cpp
@@ -52,14 +52,21 @@ ReactESP app([]() {
       {"multiplier", "Multiplier"}, {"base", "Base"}, {"offset", "Offset"}};
 
   // Here we create a new LambdaTransform objects. The template parameters
-  // (four floats in this example) correspond to the following types:
+  // (five floats in this example) correspond to the following types:
+  //
   // 1. Output type of the transform function
-  // 2. Input type of the transoform function
+  // 2. Input type of the transform function
   // 3. Type of parameter 1
   // 4. Type of parameter 2
+  // 5. Type of parameter 3
+  //
+  // The function arguments are:
+  // 1. The tranform function
+  // 2-4. Default values for parameters
+  // 5. Parameter data for the web config UI
+  // 6. Configuration path for the transform
 
-  auto log_transform = new LambdaTransform<float, float, float, float>(
-      log_function, 10, 2, 100, log_lambda_arg_data, "/transforms/log");
+  auto log_transform = new LambdaTransform<float, float, float, float, float>(
       log_function, 10, 2, 100, log_lambda_param_data, "/transforms/log");
 
   // Finally, connect the analog input via the freshly-generated log_transform
--- a/examples/lambda_transform.cpp
+++ b/examples/lambda_transform.cpp
@@ -9,6 +9,9 @@
 ReactESP app([]() {
   SetupSerialDebug(115200);
 
+  // Create a new SensESPApp object. This is the direct constructor call, and
+  // an equivalent alternative to using the SensESPAppBuilder class.
+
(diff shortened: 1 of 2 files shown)


src/transforms/lambda_transform.h
```diff
@@ -0,0 +1,364 @@
...
+                  T6 param6, const ParamInfo* param_info,
+                  String config_path = "")
+      : Transform<I, O>(config_path),
+        function4{function},
+        param1{param1},
+        param2{param2},
+        param3{param3},
+        param4{param4},
+        param5{param6},
+        param5{param6},
```
ba58smith, line 185: Typo? The first `param6` should be `param5`.

examples/lambda_transform.cpp
```diff
@@ -0,0 +1,67 @@
...
+#include "sensors/analog_input.h"
+#include "signalk/signalk_output.h"
+#include "transforms/lambda_transform.h"
+
+ReactESP app([]() {
+  SetupSerialDebug(115200);
+
+  sensesp_app = new SensESPApp("sensesp-illum-example", "My WiFi SSID",
+                               "my_wifi_password", "skdev.lan", 80, NONE);
+
```
ba58smith, line 14: I know this is a quicker approach to hard-coding certain details, but would it be better to use the Builder in examples, which seems a lot easier for a newbie to understand, and which is thoroughly documented in the Wiki?

examples/lambda_transform.cpp
```diff
@@ -0,0 +1,67 @@
...
+      new AnalogInput(pin, read_delay, analog_in_config_path, output_scale);
+
+  // This is our transform function. The example is artificial; a log transform
+  // with configurable multiplier, base, and offset parameters.
+
+  auto log_function = [](float input, float multiplier, float base,
+                         float offset) -> float {
+    return multiplier * log(input) / log(base) + offset;
+  };
+
```
ba58smith, line 42: Not a review comment - just a question about how this works: what is the purpose of `float` in this section of the code: `float offset) -> float {`? Does it set the return type of `multiplier * log(input) / log(base) + offset`?

examples/lambda_transform.cpp
```diff
@@ -0,0 +1,67 @@
...
+  // interface.
+
+  const ParamInfo* log_lambda_arg_data = new ParamInfo[3]{
+      {"multiplier", "Multiplier"}, {"base", "Base"}, {"offset", "Offset"}};
+
+  // Here we create a new LambdaTransform objects. The template parameters
+  // (four floats in this example) correspond to the following types:
+  // 1. Output type of the transform function
+  // 2. Input type of the transoform function
+  // 3. Type of parameter 1
```
ba58smith, line 54: Typo: "transoform".

ba58smith: This is the constructor for a LambdaTransform using a function with 3 parameters:
```LambdaTransform(O (*function)(I input, T1 param1, T2 param2, T3 param3),
                  T1 param1, T2 param2, T3 param3, const ParamInfo* param_info,
                  String config_path = "")```
That looks like the first parameter is the value of the function (`O (*function)(I input, T1 param1, T2 param2, T3 param3)`), the second, third, and fourth parameters are the types of the three parameters to the function (`param1`, `param2`,and `param3`), the fifth parameter is the pointer to the ParamInfo struct, and the last parameter is the config path.

You said you wanted comments on the public interface - my comment is that this is confusing to me because the comments in the code don't seem to correspond to the actual code.

File: src/transforms/lambda_transform.h [hunk 5]
```diff
@@ -0,0 +1,364 @@
...
+
+  LambdaTransform(O (*function)(I input), const ParamInfo* param_info,
+                  String config_path = "")
+      : Transform<I, O>(config_path),
+        function0{function},
+        num_params{0},
+        param_info{param_info} {
+    this->load_configuration();
+  }
+
```
ba58smith, line 93: What would be the purpose of the above method? If there are no parameters, why would you need a ParamInfo?

src/transforms/lambda_transform.h
```diff
@@ -0,0 +1,364 @@
...
+ private:
+  T1 param1;
+  T2 param2;
+  T3 param3;
+  T4 param4;
+  T5 param5;
+  T6 param6;
+  int num_params;
+  const ParamInfo* param_info;
+
```
ba58smith, line 338: For the public interface, a variable named "O" isn't so easy to understand for a new user - or at least to me. Could you make it more descriptive?

src/transforms/lambda_transform.h
mairas: Good catch! Thanks! The two last lines are both wrong. I initially had just four optional parameters but figured someone might actually run out of them and (too) quickly added a couple more.

examples/lambda_transform.cpp
mairas: I actually prefer this approach, as long as the argument list doesn't grow too long. Haven't just bothered adding a description to the wiki yet. In other words, I'd like both approaches to be illustrated in the examples. :-)

Should I add a comment describing what happens here? I'll do that.

examples/lambda_transform.cpp
mairas: C++ syntax is really all over the place. :-D Yes, it defines the return type. I'll add a comment about that. You can often leave it out and let the compiler deduce the correct type, but in this case when the expected return type is stated in the class constructors, it might be better to be explicit.

examples/lambda_transform.cpp
mairas: Ah. I added the Base parameter afterwards and apparently forgot to update the comments. Will fix.

src/transforms/lambda_transform.h
mairas: Strictly speaking, you don't. I didn't have this in the first place but the lack of symmetry offended my sensibilities and I decided to add it. Basically, if you implemented a transform without parameters, you could modify the example to just provide an empty ParamInfo array (or leave it intact and it'd just be ignored). I was going to offer to remove it but after writing that justification, I'd rather leave it. :-)

src/transforms/lambda_transform.h
mairas: For some reason it's customary to use very short type template names, but you're right, they're obscure. I'll make them OUTPUT, INPUT, P1, etc. I could also expand Pn to PARAMn but that'd make the quite heavy function signatures even longer. If you prefer the longer one, just say.

examples/lambda_transform.cpp
mairas: The comment listing was referring to the template type parameters, and actually the call missed one. It just happened to work because float is the default type. Basically, that's the drawback with my implementation approach: you have all the template type parameters always defined by they just default to float. The alternative would've been genuinely variable-length template parameter list but that didn't work with the compiler version and flags in use.

Should I make the default type something obscure to nudge people to be more explicit?

mairas:
Updated according to comments.

JohnySeven:
Very nice, I like it!

examples/lambda_transform.cpp
mairas: I made the default type `bool`. This will make the compiler throw an error if non-bool parameters are used in the lambda function and the template types are not explicitly defined.

mairas:
Barring no more comments, merging.

examples/lambda_transform.cpp
ba58smith: Yes, a comment that explains that this is one of threee approaches, briefly explains this approach, and points to the Wiki for all the documentation of thsi approach, seems like a very good idea.

examples/lambda_transform.cpp
ba58smith: That seems like a good solution, as the type is not likely to be `bool`.

examples/lambda_transform.cpp
mairas: Why three? You either instantiate the class directly or via the AppBuilder object.

I just added a mention about this being the direct instantiation alternative. I'll try to remember describing that in the wiki soon.

tedenda: Isn't the third option when you don't pass any settings and use captive portal to setup the sensor?

ba58smith: I already updated the wiki. And yes, @tedenda , that's the third option:
1. sensesp_app = new SensESPApp(); // the original way
2. sensesp_app = new SensESPApp( put your hard coding here);
3. using the Builder

mairas: Oh. I don't see any categorical difference between 1 and 2. Both constructor arguments and builder allow you to independently:
- define wifi or leave it undefined in which case the captive portal is enabled
- define the SK server or leave it undefined in which case mDNS is used
- define any of the other options

If you leave everything undefined, you're just going with the defaults but I wouldn't regard it as a categorical difference.

ba58smith: It's not a technical difference - it's just a "conversational" difference: no hard-coding, hard-coding with the SensESPApp constructor, and hard-coding with the Builder. It's not important - but is now all explained in the wiki.

Issue #173, opened 2020-09-26
Allow device to fill in requested permissions in token request to server


JohnySeven:
Hello,

it's just small think but it will allow devices to define what SK permissions they need in order to work, some of them needs just "readonly", mostly they need "readwrite" or "admin".

My sugestion is to change the token request to provide "permissions" in JSON token request and add new method in SensESP builder to configure this. Default permission will be "readwrite" as most devices needs to pull the data in.

This scenario is supported by SK server here:
(https://github.com/SignalK/signalk-server/blob/1c80214bd86272c04c47b8cf8be363e617d202a3/src/tokensecurity.js#L1035)

Just let me know what you think about this.

I'm willing to make PR.

Thanks,
Jan

ba58smith:
I like this - it will mean one less step in getting a new ESP sending data to Signal K.

mairas:
Oh, I didn't know that was supported! That's annoyed me forever. I'd love it too.

ba58smith:
Closed with PR #201.

PR #201, opened 2020-10-09
Allow device to fill in requested permissions in token request to server

JohnySeven:
Hi Guys,

I've implemented sending permissions in token request to SignalK server - tested with readwrite (default) permission and also with admin permission. Solves issue #173 

I've changed set_sk_server method on Builder and added third parameter permission with default READWRITE value. I've have changed SensESPApp ctor paramteters to add permision. Same with WSClient.

Let me know what you think!
Jan

### DIFF
--- a/src/sensesp_app.cpp
+++ b/src/sensesp_app.cpp
@@ -38,7 +38,8 @@ SensESPApp::SensESPApp(String preset_hostname, String ssid,
                        String wifi_password, String sk_server_address,
                        uint16_t sk_server_port, StandardSensors sensors,
                        int led_pin, bool enable_led, int led_ws_connected,
-                       int led_wifi_connected, int led_offline) {
+                       int led_wifi_connected, int led_offline,
+                       SKPermissions permissions) {
   // initialize filesystem
 #ifdef ESP8266
   if (!SPIFFS.begin()) {
@@ -84,8 +85,9 @@ SensESPApp::SensESPApp(String preset_hostname, String ssid,
     }
   };
   auto ws_delta_cb = [this]() { this->led_blinker->flip(); };
-  this->ws_client = new WSClient("/system/sk", sk_delta, sk_server_address,
-                                 sk_server_port, ws_connected_cb, ws_delta_cb);
+  this->ws_client =
+      new WSClient("/system/sk", sk_delta, sk_server_address, sk_server_port,
+                   ws_connected_cb, ws_delta_cb, get_permission_string(permissions));
 }
 
 void SensESPApp::setup_standard_sensors(ObservableValue<String>* hostname,
@@ -178,6 +180,26 @@ void SensESPApp::reset() {
   networking->reset_settings();
   SPIFFS.format();
   app.onDelay(1000, []() { ESP.restart(); });
+}
+
+String SensESPApp::get_permission_string(SKPermissions permission) 
+{
+  if(permission == READONLY)
+  {
+    return "readonly";
+  }
+  else if(permission == READWRITE)
+  {
+    return "readwrite";
+  }
+  else if(permission == ADMIN)
+  {
+    return "admin";
+  }
+  else
+  {
+    return "";
+  }  
 }
 
 String SensESPApp::get_hostname() { return networking->get_hostname()->get(); }
--- a/src/sensesp_app.cpp
+++ b/src/sensesp_app.cpp
@@ -34,6 +34,8 @@ void SetupSerialDebug(uint32_t baudrate) {
   debugI("\nSerial debug enabled");
 }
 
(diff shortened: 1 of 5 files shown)


src/sensesp_app.cpp
```diff
@@ -169,6 +171,26 @@ void SensESPApp::reset() {
   app.onDelay(1000, []() { ESP.restart(); });
 }
 
+String SensESPApp::get_permission_string(SKPermissions permission) 
+{
+  if(permission == READONLY)
```
mairas, line 176: You can achieve the same neater (and arguably more efficiently) by placing the permission strings in an array of strings and then referencing the array by coercing the enum to an int:

```
char* permission_strings[] = {
  "readonly",
  "readwrite",
  "admin"
};

...

permission_string = permission_strings[(int)permission];
```

(Not sure if the type definitions are correct in my example, but you get the point. Also, didn't check if the enums were in that order.)
};

mairas:
This looks golden, except for the small comment/suggestion I made. And seems like it's now conflicting with some prior changes. If you need help with resolving the conflict, send me a message.

ba58smith:
What is the default expiration of permissions if they are set this way? Hopefully "NEVER".

mairas:
> What is the default expiration of permissions if they are set this way? Hopefully "NEVER".

Blank, I suppose, if the server didn't support setting the expiration for auth requests. But possibly a better fix would be to set the default expiration on the server side to NEVER? @tkurki

src/sensesp_app.cpp
JohnySeven: Good point, will change that!

JohnySeven:
Yes, it's blank and it would be nice if the server would fill in default from configuration - "NEVER" would be fine. I hope somebody from node server team will maybe do it.

ba58smith:
Does it seem a bit risky to implement this change without knowing how the server is going to handle a blank expiration?

mairas:
> Does it seem a bit risky to implement this change without knowing how the server is going to handle a blank expiration?

This is only about an auth request. The user/admin still needs to go and approve the request. The only difference is that now the required permissions are pre-filled.

ba58smith:
> > Does it seem a bit risky to implement this change without knowing how the server is going to handle a blank expiration?
> 
> This is only about an auth request. The user/admin still needs to go and approve the request. The only difference is that now the required permissions are pre-filled.

Ah, I see. Just trying to prevent someone from approving a request for "ReadOnly" because they don't change it to "ReadWrite". Got it.

mairas:
It wouldn't be much of an authentication scheme if the client was allowed to authorize itself! :-D

ba58smith:
> It wouldn't be much of an authentication scheme if the client was allowed to authorize itself! :-D

Oh, you and your LOGIC! :-)

src/sensesp_app.cpp
JohnySeven: Done

JohnySeven:
@mairas it's ready to be merged, thanks!

mairas:
Before I merge, could you remove the last merge commit from the PR?

JohnySeven:
I've reverted merge commit from repo. Thanks

tedenda:
My guess is that something went wrong?
Now there is 153 files changed and conflicts?

mairas:
This PR indeed has quite a few extra commits and dozens of unrelated files on it. Instead of adding revert commits (which just add more stuff on the repo), remove the extra commits, either with `git reset --hard [...]` or `git rebase -i [...]`. Then rebase your branch against the latest master, and once that's done, force-push the changes to your forked repo. That'll fix the PR.

tedenda:
Fixes #173 when merged

JohnySeven:
I've done what @mairas suggested and PR is now in good shape! Thanks @mairas :-)

mairas:
I'm sorry for this to come in so late, but now that I began thinking about it, is there ever a reason for requesting any other than readwrite permission? Even if you implemented a pure control device, you'd still want to report back the state of the controlled interface, so readonly is not feasible. And on the other hand, "admin" is not relevant either because I don't see SensESP devices mucking with the server in such a manner.

Assuming this, I'd simply hard-code the value and remove all that complexity-adding configurability.

JohnySeven:
@mairas readonly can be used for pure displays that just show boat information, admin can be used for sending put requests to control another device on boat - I know it's not implemented right now, but it's something I would like to implement in future as the SK protocol allows this scenario and is used in KIP web app. The default value is readwrite and I don't think it's not a big change in code, nor some breaking change, it just adds more features to SensESP.

mairas:
OK, you've got a point there. Merging!

JohnySeven:
@mairas  thanks! :-)

PR #203, opened 2020-10-11
Stop using [[deprecated]] attributes due to too old GCC

mairas:
This PR takes care of a lot of compiler warnings.

Commenting code out should generally be avoided but I resorted to that with a clear FIXME message to indicate that those lines should be restored once the toolchain allows that.


ba58smith:
Looks good to me!

PR #205, opened 2020-10-13
Fix some variable names that dodged the first round

mairas:
Some member variables had still the Hungarian notation pointer prefixes. Removed in this PR.


ba58smith:
Looks OK to me.

PR #206, opened 2020-10-13
Linear transform reimplemented as a LambdaTransform

mairas:
This is a reimplementation of the `Linear` transform class, illustrating the potential benefits of implementing the basic transforms using `LambdaTransform`.

The first commit is a bugfix for `LambdaTransform` implementation. If desired, I can make it a separate PR but this Linear transform reimplementation depends on it.

### DIFF
--- a/src/transforms/lambda_transform.h
+++ b/src/transforms/lambda_transform.h
@@ -154,7 +154,7 @@ class LambdaTransform : public Transform<IN, OUT> {
         param2{param2},
         param3{param3},
         param4{param4},
-        param4{param5},
+        param5{param5},
         num_params{5},
         param_info{param_info} {
     this->load_configuration();
--- a/src/transforms/lambda_transform.h
+++ b/src/transforms/lambda_transform.h
@@ -19,29 +19,19 @@ static const char kLambdaTransformSchemaTail[] PROGMEM = R"(
  * template specializations.
  **/
 template <class T>
-const char* get_schema_type_string(const T dummy) {
-  return "string";
-}
+const char* get_schema_type_string(const T dummy);
 
 template <>
-const char* get_schema_type_string(const int dummy) {
-  return "number";
-}
+const char* get_schema_type_string(const int dummy);
 
 template <>
-const char* get_schema_type_string(const float dummy) {
-  return "number";
-}
+const char* get_schema_type_string(const float dummy);
 
 template <>
-const char* get_schema_type_string(const String dummy) {
-  return "string";
-}
+const char* get_schema_type_string(const String dummy);
 
 template <>
-const char* get_schema_type_string(const bool dummy) {
-  return "boolean";
-}
+const char* get_schema_type_string(const bool dummy);
 
 /**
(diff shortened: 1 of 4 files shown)


src/transforms/lambda_transform.h
```diff
@@ -19,29 +19,19 @@ static const char kLambdaTransformSchemaTail[] PROGMEM = R"(
...
-const char* get_schema_type_string(const String dummy) {
-  return "string";
-}
+const char* get_schema_type_string(const String dummy);
 
 template <>
-const char* get_schema_type_string(const bool dummy) {
-  return "boolean";
-}
+const char* get_schema_type_string(const bool dummy);
```
ba58smith, line 35: The constructor below for 5 parameters has a typo: `param4{param5},`.

src/transforms/linear.cpp
```diff
@@ -1,42 +1,17 @@
...
-static const char SCHEMA[] PROGMEM = R"({
-    "type": "object",
-    "properties": {
-        "k": { "title": "Multiplier", "type": "number" },
-        "c": { "title": "Constant offset", "type": "number" },
-        "value": { "title": "Last value", "type" : "number", "readOnly": true }
-    }
-  })";
+float (*Linear::function)(float, float, float) =
+    [](float input, float k, float c) { return k * input + c; };
```
ba58smith, line 12: This isn't a suggestion for you to change anything - it's just to help me understand what the above statement is and does.
Would the above method definition work without the parentheses around `*Linear::function`? Similarly, in the .h file, do you need the parenthesis around `*function` in `static float (*function)(float, float, float);`?
Lastly, why do you need the `*Linear::` before `function`? Is there some OTHER `function` that this definition could be confused with?

src/transforms/lambda_transform.h
mairas: Wow, still? Thanks.

src/transforms/linear.cpp
mairas: That's not a method, that's a member variable. :-) You just stumbled upon what's in my opinion probably the crappiest syntax in C++.
```
float (*function)(float, float, float);
```
defines a pointer to function that returns a float and takes three floats as an argument. I just then initialize the pointer variable with the lambda function.

This would be equivalent:
```
float linear_fn(float input, float k, float c) {
  return k * input + c;
}
float (*Linear::function)(float, float, float) = &linear_fn;
```

It's `Linear::function` because it's a member variable.

Did that help you get it?

mairas:
Yes, my eventual idea is to refactor all simple transforms to be implemented as a LambdaTransform. It reduces a lot of repetition in the transform implementations.

In principle, I'd love to combine those classes to a single .h/.cpp pair but the header files are already part of the interface. :-( Basically, we'd break a lot of existing code if we moved the classes around.

src/transforms/lambda_transform.h
mairas: Added a new commit with fix.

mairas:
Merging.

PR #207, opened 2020-10-13
Hysteresis transform implementation using LambdaTransform

mairas:
Work in progress:

- has extra commits from existing PRs
- No documentation

### DIFF
--- a/src/transforms/hysteresis.h
+++ b/src/transforms/hysteresis.h
@@ -1,0 +1,54 @@
+#ifndef _hysteresis_H_
+#define _hysteresis_H_
+
+#include "lambda_transform.h"
+
+/**
+ * @brief Hysteresis function
+ * 
+ * Hysteresis is a threshold switch with a dead zone: to prevent switching
+ * back and forth at the threshold value, input has to rise above the
+ * upper threshold before being switched high, and has to decrease below
+ * the lower threshold before being switched back low.
+ * 
+ * @tparam IN Input variable type
+ * @tparam OUT Output variable type
+ * @param[in] lower_threshold Lower threshold value
+ * @param[in] upper_threshold Upper threshold value
+ * @param[in] low_output Output value when hysteresis function output is low
+ * @param[in] high_output Output value when hysteresis function output is high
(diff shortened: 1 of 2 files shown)


src/transforms/hysteresis.h
```diff
@@ -0,0 +1,41 @@
...
+  const ParamInfo param_info[4] = {
+      {"lower_threshold", "Lower threshold"},
+      {"upper_threshold", "Upper threshold"},
+      {"low_output", "Low output"},
+      {"high_output", "High output"}};
+
+  static OUT function(IN input, IN lower_threshold, IN upper_threshold,
+                      OUT low_output, OUT high_output) {
+    static OUT last_value = low_output;
+    if (upper_threshold <= input) {
```
ba58smith, line 26: This is REALLY trivial... but it seems the logic is easier to follow if the above is `if (input >= upper_threshold)`.

src/transforms/hysteresis.h
```diff
@@ -0,0 +1,41 @@
...
+      return high_output;
+    }
+    if (lower_threshold <= input && input < upper_threshold) {
+      return last_value;
+    }
+    if (input < lower_threshold) {
+      last_value = low_output;
+      return low_output;
+    }
+    return low_output;
```
ba58smith, line 37: When would this last `return low_output;` ever be reached? The three if statement cover all possible bases, don't they?

ba58smith: Can't it be even simpler?
If input >= upper_threshold, last_value = high_output and return high_output.
If input < lower_threshold, last_value = low_output and return low_output.
Otherwise, return last_value, because the only remaining choice is that input is between upper and lower.

mairas: The last return would never be reached but the compiler didn't know it and was complaining about not having a return value. But yes! You're right. I was being silly and blindly followed the hysteresis definition. I'll update.

ba58smith:
Looks good to me!

mairas:
OK, I'll add some comments and class documentation and merge after PR#206.

mairas:
Merging based on a previous review.

PR #215, opened 2020-10-19
SK Paths are now valid

### LINKED ISSUE #212: Update all examples to use correct SK path
When this is done, add link to to documentation.

tedenda:
There are some paths that not exists in documentation, I have not touched those.
Fixes #212

### DIFF
--- a/examples/temperature_sender.cpp
+++ b/examples/temperature_sender.cpp
@@ -11,8 +11,7 @@
 
 /*
   Illustrates a custom transform that takes a resistance value in ohms and
-  returns the estimated temperature in Kelvin. Sample data in this example were
-  taken from a Westerbeke generator temperature sender and gauge. Note that you
+  returns the estimated temperature in Kelvin. Note that you
   will never instantiate CurveInterpolator in your own SensESP project - you
   will always need to create a descendant class of it, like
   TemperatureInterpreter in this example. The constructor needs to do only two
@@ -79,8 +78,9 @@ ReactESP app([]() {
 
   // The "Signal K path" identifies the output of this sensor to the Signal K
   // network.
-
-  const char* sk_path = "electrical.generator.engine.water.temp";
+  // To find valid Signal K Paths that fits your need you look at this link:
+  // https://signalk.org/specification/1.4.0/doc/vesselsBranch.html
+  const char* sk_path = "electrical.alternators.12V.temperature";
 
   /*
   Connecting a physical temperature sender to the MCU involves using a "voltage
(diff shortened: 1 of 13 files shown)


examples/temperature_sender.cpp
```diff
@@ -79,8 +79,9 @@ ReactESP app([]() {
 
   // The "Signal K path" identifies the output of this sensor to the Signal K
   // network.
-
-  const char* sk_path = "electrical.generator.engine.water.temp";
+  // To find valid Signal K Paths that fits your need you look at this link:
+  // https://signalk.org/specification/1.4.0/doc/vesselsBranch.html
+  const char* sk_path = "propulsion.generator.coolantTemperature";
```
ba58smith, line 85: I don't suppose it matters very much, but there is no "generator" path in the spec. I may have created this example, and chose "electrical" rather than "propulsion" because a generator is more related to making electricity than it is to propelling the boat.

tedenda: My thinking was that it could be a genset unit. In that case it is a motor with coolant, rpm, exhaust etc.

Change it to an electrical.alternators.12V.temperature ?

ba58smith: Yes, it is a motor / engine with coolant, rpm, exhaust, etc. But its purpose is to generate electricity, not to propel the boat. So it seems more appropriate to call it "electrical.generator.coolantTemperature". It really doesn't matter - it's just an example, and an end user will call it whatever they want.

(I see your point, though - by putting it in the "propulsion" group, you then have all the other paths that relate to it being a motor / engine. I actually submitted a PR to the "specification" project - https://github.com/SignalK/specification/pull/584 - but I can't get it to pass the json test, so I've given up on it. That PR adds "generator" as a top-level group, like electrical and propulsion, and combines all of the electrical and the propulsion paths that are relevant to a genset.

tedenda: Yes, I agree.

Someday that PR will be taken care of, there must be a lot of boaters out there that have a genset they want to keep track of.

ba58smith: Since you changed it to "alternator", you should eliminate this sentence from the first big comment at the top of the file: "Sample data in this example were taken from a Westerbeke generator temperature sender and gauge.".

examples/thermocouple_temperature_sensor/max31856_thermocouple_example.cpp
```diff
@@ -39,9 +41,9 @@ ReactESP app([]() {
   // run-time configuration.
 
   const char* temperature_in_config_path =
```
ba58smith, line 43: Should the above be `exhaust_temp_config_path" to match the other change you made?

tedenda: Yes, thanks!

ba58smith:
This all looks good to me, but I'm not sure what commit 12b70e9 is. Is it to update the files in this branch for changes that were made in the master branch after you started working on these files? That's what it looks like, and I suppose that's how it works - but I'm not going to merge this one. I'll leave it up to @mairas.

mairas:
Yeah, it's a merge commit that shouldn't be part of the PR. :-)

examples/temperature_sender.cpp
tedenda: Corrected, thanks!

mairas:
Merging. Sorry for the delay!

PR #219, opened 2020-10-22
Missing magnetometer dependencies

mairas:
Some dependencies were missing from the previous magnetometer PR and new builds broke due to that. This PR fixes that. Unfortunately, PlatformIO's Library Dependency Finder (LDF) is no longer able to find all dependencies with the default settings, so changes to `platformio.ini` are required.


mairas:
Merging right away.

PR #221, opened 2020-10-22
Actions workflow in github to build examples

plerenius:
Added GitHub Actions flow to automatically build example files
Fixed some examples to work with all specified boards in platform.ini
I'm not sure all pin numbers are suitable for ESP32

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,41 @@
+name: SensESP Automatic Build
+
+on: [push]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        example: [
+          examples/ads1x15_volt_meter.cpp,
+          examples/ina219_example.cpp,
+          examples/analog_input.cpp,
+          examples/lambda_transform.cpp,
+          examples/bme280_example.cpp,
+          examples/bmp280_example.cpp,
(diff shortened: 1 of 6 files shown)


.github/workflows/test.yml
```diff
@@ -0,0 +1,41 @@
...
+    steps:
+    - uses: actions/checkout@v1
+    - name: Set up Python
+      uses: actions/setup-python@v1
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install platformio
+    - name: Run PlatformIO
+      run: platformio ci -v --lib . -c ./platformio.ini
```
mairas, line 39: This is really neat! I wasn't aware of the `ci` command!

mairas:
Cool stuff! Thanks a bunch! GPIOs 4 and 5 are OK for normal use on ESP32. Maybe in the long term (or never) it might be more efficient to start using pre-built docker images, but this is perfect for now! Couldn't find anything to whine about, will merge right away.

.github/workflows/test.yml
plerenius: I found it a few days ago and thought it was really handy for SensESP project :)
I haven't gotten examples/hysterisis.cpp to build for d1_mini as I stated in Slack, so I omitted it for now.

Issue #216, opened 2020-10-19
Update fx8700_heading_example to write to correct SK Path


tedenda:
When I was updating all examples to output valid SK Paths I found that fx8700_heading_example don't format the Roll/Pith/Yaw according to [documentation](https://signalk.org/specification/1.4.0/doc/vesselsBranch.html#vesselsregexpnavigationattitude).

I discussed this with @mairas and the conclusion is that it should be just one path that outputs a Json object of the given format.

Look at signalk_position.cpp as a template.

I don't find any valid SK paths for acceleration(x, y, z) and rate(roll, pitch, yaw) in the documentation, how we should handle those is still an open question.

BjarneBitscrambler:
Thanks for bringing this up. I can have a look at this issue in a couple days - I'll wait till PR #166 is closed first.

tedenda:
@BjarneBitscrambler I started to code, added the formatting files, but I don't know how to test it nor implement it in NXP FX87. Got some tips from @mairas how to test it via the LambdaTransform for example, but I didn't figure it out. Please continue this work and I will eventually learn a bit.

You find my work at my fork under branch AddAttidudeFormat.

DanielG86:
Regarding the SK-paths: I've noticed that it writes to "orientation.XXX", and not to "navigation.XXX". 
Is this the same remark you made, @tedenda ?

tedenda:
I'm aware of this but waiting to fix it. When we have fixed that this sensor can output Attitude data correctly we need to update this example anyway. There is other things that also need to be fixed.

I'll fix it very soon.

tedenda:
I have now updated the example to match valid SK paths, added attitude that will be added etc.
Work is done in PR #225 But merging that has to wait until this issue is solved and merged.

BjarneBitscrambler:
@tedenda and @DanielG86 - I can adjust the paths in the orientation example if you want. I'm editing anyway to incorporate the fusion library v7.  If you want to try out my first pass it's at https://github.com/BjarneBitscrambler/SensESP.git in the OrientationV7 branch. I only focussed on getting it working on the ESP32 - I haven't yet tried it on my ESP8266, nor have I started putting in the various changes that have accumulated in SensESP over the past few months...

tedenda:
I have a new ESP32 now so I can try it.

PR #225, opened 2020-10-24
Update FX8700 example

tedenda:
This is the updated example with correct SK paths and also have Attitude in place.

This PR can only be merged when #216 is done and this example is verified to work as expected.

### DIFF
--- a/examples/fx8700_heading_example.cpp
+++ b/examples/fx8700_heading_example.cpp
@@ -19,7 +19,7 @@
 // defines an "app" object vs defining a "main()" method.
 ReactESP app([]() {
 
-// Some initialization boilerplate when in debug mode...
+// Start serial debug if not disabled
 #ifndef SERIAL_DEBUG_DISABLED
   SetupSerialDebug(115200);
 #endif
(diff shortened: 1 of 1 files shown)


DanielG86:
Great.
I know whether the attitude approach works, hopefully you do. 
In line 44 the sk path shows “navigation/headingCompass”, where I believe the / should be a dot.
Furthermore, you could argue to make this headingMagnetic, since after calibration on the boat, the sensor should give magnetic heading data.

tedenda:
@DanielG86 Thanks, updated!
Regarding the headingMagnetic, you are probably right, changed to that for now.

BjarneBitscrambler:
@tedenda - The Sensor Fusion v7 library is working and I have started porting the SensESP Orientation example over to use this library. Would you like me to review your changes proposed in this PR and merge them into the revised Orientation example? That would allow us to retire this PR once SensESP is switched over to the new library.

tedenda:
Yes, that sounds like a good idea.

When you add your PR you just mention this PR with `Close #225` GitHub will pick that up and do that while merging your PR.

BjarneBitscrambler:
I now have a working example that uses the correct paths for *Attitude* and *headingCompass*. It's submitted as PR#349 if you could have a look... Thanks!

mairas:
This PR seems to have gone stale; closing.

PR #229, opened 2020-11-03
build_flags added, each board can have dedicated pin definitions

KEGustafsson:
Can we add following build_flags to platformio.ini to enable dedicated pin definition for each boards independently?
I added example usage of this modification to analog_input.cpp.

### DIFF
--- a/examples/analog_input.cpp
+++ b/examples/analog_input.cpp
@@ -6,6 +6,18 @@
 #include "signalk/signalk_output.h"
 #include "transforms/linear.h"
 
+#ifdef D1MINI
+#define ANALOG A0  //Always A0
+#endif
+
+#ifdef ESP32DEV
+#define ANALOG 36 //ADC1_0
+#endif
+
+#ifdef ESPWROVERKIT
+#define ANALOG 0 //ADC2_1
+#endif
+
 // SensESP builds upon the ReactESP framework. Every ReactESP application
 // defines an "app" object vs defining a "main()" method.
 ReactESP app([]() {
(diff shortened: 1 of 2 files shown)


ba58smith:
I'm not in favor of this PR. It essentially hard-codes an AnalogInput pin on the ESP32 boards, which isn't appropriate, as they can use at least 15 different pins for AnalogInput. It seems to me that the current analog_input.cpp (the example) adequately addresses the pin issue in the comment about which pin to use.

KEGustafsson:
Analog pin is just example use case for this build flag. With this you describe differences between boards and keep the code same. 
I'm not sure how ESP8266 and ESP32 codes are planned to be build now if you would like to use same code base?
Pins cannot be the same in ESP8266 and ESP32.

ba58smith:
@KEGustafsson - I don't see the benefit, at least not in the way I work with SensESP. That is, each ESP I use is a separate project, with separate physical wiring of sensors and resistors and other components, and I choose the various pins depending on the physical location of them relative to all the other components. To make the change in my main.cpp, where I can see at a glance which pin I've used for which sensor, seems way easier than adding it to the platformio.ini. But that's just me - we can of course agree to disagree. I'll leave it up to someone else to merge it, if they so desire.

KEGustafsson:
I see your point when each ESP is a separate project. I came up with this when I started to shift from ESP8266 to ESP32, moving old projects to run on a new platform. It might help during this kind of transition period, but in a longer run it might be not so usable after all.
This is just me thinking this way.

KEGustafsson:
Just noticed that this same is already in there, so this is not needed.
https://github.com/SignalK/SensESP/blob/master/examples/fuel_flow_meter.cpp#L16-L22.
I'll remove PR.

PR #230, opened 2020-11-03
Add load_configuration() to DigitalInputValue constructor

ba58smith:
While testing another PR, I noticed this issue. It goes back to PR #125, when DigitalInputValue was first made configurable by the addition of read_delay, but load_configuration() wasn't added to the constructor at that time.


JohnySeven:
Nice catch! 👍

mairas:
Thanks! Merging.

PR #233, opened 2020-11-04
fix: ESP32 hostname for mDNS name

KEGustafsson:
ESP32 SensESP is not using hostname for mDNS name at the moment. This fix the issue and show correctly user defined hostname also in mDNS name.


mairas:
Looks good! Merging.

PR #234, opened 2020-11-04
ValueProducer::emit(value)

mairas:
Most sensors keep repeating the following pattern:

    output = some_value;
    notify();

This PR implements `ValueProducer::emit(value)` that changes the pattern to the following:

    this->emit(some_value);

### DIFF
--- a/src/sensors/analog_input.cpp
+++ b/src/sensors/analog_input.cpp
@@ -14,8 +14,7 @@ AnalogInput::AnalogInput(uint8_t pin, uint read_delay, String config_path,
 }
 
 void AnalogInput::update() {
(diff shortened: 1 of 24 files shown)


JohnySeven:
Looking very nice! 👍

ba58smith:
Was this done simply to reduce two lines of code to one? (In a whole bunch of places, of course.) I'm not questioning it - just want to make sure I understand exactly what this PR does.

mairas:
Yes, pretty much so. And also to reduce the usage of an explicit output variable, which I think is a bit of an anti-pattern in our `ValueProducer` class implementations.

mairas:
Merging.

PR #236, opened 2020-11-05
Networking and WSClient objects report their state via ValueProducers

mairas:
Previously, Networking and WSClient objects reported their state using callbacks which wasn't in line with the rest of the architecture and was also quite limiting in what you could do.

Now, Networking and WSClient are both ValueProducers that notify their consumers of state changes. LedBlinker is one such consumer.

This change in itself is not a very significant one but paves way for refactoring the LedBlinker class and also, with a bit of further refactoring, allows for example for having the device state to be displayed on a small I2C OLED display.

### DIFF
--- a/src/net/networking.cpp
+++ b/src/net/networking.cpp
@@ -63,8 +63,7 @@ void Networking::setup() {
 
 void Networking::setup_saved_ssid() {
   WiFi.begin(ap_ssid.c_str(), ap_password.c_str());
-  this->output = kWifiDisconnected;
-  this->notify();
+  this->emit(kWifiDisconnected);
 
   uint32_t timer_start = millis();
 
@@ -83,8 +82,7 @@ void Networking::setup_saved_ssid() {
     debugI("Connected to wifi, SSID: %s (signal: %d)", WiFi.SSID().c_str(),
            WiFi.RSSI());
     debugI("IP address of Device: %s", WiFi.localIP().toString().c_str());
-    this->output = kWifiConnectedToAP;
-    this->notify();
+    this->emit(kWifiConnectedToAP);
     WiFi.mode(WIFI_STA);
   }
 }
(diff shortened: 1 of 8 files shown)


src/net/networking.cpp
```diff
@@ -20,6 +20,8 @@ Networking::Networking(String config_path, String ssid, String password,
     : Configurable{config_path} {
   this->hostname = new ObservableValue<String>(hostname);
 
+  this->output = kNoAP;
```
ba58smith, line 23: This is the only instance in this file in which `this->notify()` is not called immediately after setting a new value for `output`. Is that correct?

src/net/networking.cpp
```diff
@@ -100,14 +111,22 @@ void Networking::setup_wifi_manager(std::function<void(bool)> connection_cb) {
   config_ssid = "Configure " + config_ssid;
   const char* pconfig_ssid = config_ssid.c_str();
 
+  this->output = kWifiManager;
+  this->notify();
+
```
ba58smith, line 115: Trivial, I know: could `kWifiManager` be changed to `kRunningWifiManager` or `kExecutingWifiManager`? I'm thinking of how it will look when it's displayed on a little OLED screen, thinking it will be a little more obvious what the status means.

src/net/networking.cpp
mairas: Yes, that's because it's in the constructor and it's just about setting an initial value, not about some change that should be notified.

But now that you mentioned it, I hadn't paid attention to that file's repetition of that same pattern. I'll replace them with emit calls.

src/net/networking.cpp
```diff
@@ -100,14 +111,22 @@ void Networking::setup_wifi_manager(std::function<void(bool)> connection_cb) {
...
   if (!wifi_manager->autoConnect(pconfig_ssid)) {
     debugE("Failed to connect to wifi and config timed out. Restarting...");
+
+    this->output = kDisconnected;
+    this->notify();
+
```
ba58smith, line 121: For the same reason: how about `kWifiDisconnected`? Otherwise, it won't be obvious to someone who isn't familiar with the code WHAT is disconnected: is it referring to the connection to the SK Server? The ws connection?

src/net/networking.cpp
mairas: Yes, it's a good change. I'll do it. Note, though, that anything that gets displayed is unrelated to the enum name: it could be even a symbol or something in Finnish, for that matter. :-D

src/net/networking.cpp
mairas: Yes, absolutely. And for the rest of the symbols in the WifiState enum, too.

src/net/networking.cpp
ba58smith: Noooo!!!! Nothing in Finnish!!!! I have enough trouble understand C++. :-)

ba58smith:
Looks great!

PR #237, opened 2020-11-05
Split SensESPApp initialization to separate constructor and setup calls

mairas:
The SensESPApp constructor had a rather long argument list, with most arguments being rather inessential. This wasn't great as a public interface and would've caused future refactoring of the arguments always break the public interface.

This change is required to limit the growth of the constructor argument list length, to maintain separation of public and non-public interfaces and to only have the essential arguments in the public constructor.

### DIFF
--- a/examples/hysteresis.cpp
+++ b/examples/hysteresis.cpp
@@ -10,7 +10,7 @@ ReactESP app([]() {
   SetupSerialDebug(115200);
 
   sensesp_app = new SensESPApp("sensesp-hysteresis-example", "My WiFi SSID",
-                               "my_wifi_password", "skserver.lan", 80, NONE);
+                               "my_wifi_password", "skserver.lan", 80);
 
   const char* sk_path = "environment.indoor.illuminance";
   const char* analog_in_config_path = "/indoor_illuminance/analog_in";
--- a/examples/lambda_transform.cpp
+++ b/examples/lambda_transform.cpp
@@ -13,7 +13,7 @@ ReactESP app([]() {
   // an equivalent alternative to using the SensESPAppBuilder class.
 
   sensesp_app = new SensESPApp("sensesp-illum-example", "My WiFi SSID",
-                               "my_wifi_password", "skdev.lan", 80, NONE);
+                               "my_wifi_password", "skdev.lan", 80);
 
   // To find valid Signal K Paths that fits your need you look at this link:
   // https://signalk.org/specification/1.4.0/doc/vesselsBranch.html
--- a/src/sensesp_app.h
+++ b/src/sensesp_app.h
@@ -44,11 +44,10 @@ void SetupSerialDebug(uint32_t baudrate);
 
 class SensESPApp {
  public:
+  SensESPApp(bool defer_setup);
   SensESPApp(String hostname = "SensESP", String ssid = "", String wifi_password = "",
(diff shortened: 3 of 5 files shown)


src/sensesp_app.h
```diff
@@ -84,7 +83,36 @@ class SensESPApp {
...
+  const SensESPApp* set_led_wifi_connected(int led_wifi_connected) { this->led_wifi_connected = led_wifi_connected; }
+  const SensESPApp* set_led_offline(int led_offline) { this->led_offline = led_offline; }
+  const SensESPApp* set_requested_permissions(SKPermissions permissions) { this->requested_permissions = permissions; }
+
  private:
+  String preset_hostname = "SensESP";
+  String ssid = "";
+  String wifi_password = "";
+  String sk_server_address = "";
+  uint16_t sk_server_port = 3000;
```
ba58smith, line 107: This changes the default server port from 0 to 3000. I think there is code that looks at the server port, and if it's 0, that means it hasn't been configured yet. Did you intend to make this change?

examples/hysteresis.cpp
```diff
@@ -10,7 +10,7 @@ ReactESP app([]() {
   SetupSerialDebug(115200);
 
   sensesp_app = new SensESPApp("sensesp-hysteresis-example", "My WiFi SSID",
-                               "my_wifi_password", "skserver.lan", 80, NONE);
+                               "my_wifi_password", "skserver.lan", 80);
```
ba58smith, line 13: I think this changes the behavior of the Std Sensors in this example from "NONE" to "ALL". (Since "ALL") is the default.) 

If you add `sensesp_app.set_sensors(NONE);` right here - that keeps the same behavior, right? (And illustrates how the new setters are used.)

examples/lambda_transform.cpp
```diff
@@ -13,7 +13,7 @@ ReactESP app([]() {
   // an equivalent alternative to using the SensESPAppBuilder class.
 
   sensesp_app = new SensESPApp("sensesp-illum-example", "My WiFi SSID",
-                               "my_wifi_password", "skdev.lan", 80, NONE);
+                               "my_wifi_password", "skdev.lan", 80);
```
ba58smith, line 16: Same comment as above about the Std Sensors.

src/sensesp_app.h
mairas: Good catch! I didn't think it through. I'll revert back to the previous default.

examples/hysteresis.cpp
mairas: Note that those SensESPApp setters are not public - they're only available to the friend classes (or within the class itself, obviously). They rely on being called before `setup()` is being called, and the default constructor calls `setup()` immediately.

Instead, my thinking was that the default constructor should provide access to the most essential settings and rely on sensible defaults for the others. The builder class provides access to the more obscure settings. If you want to change the enabled sensors, you need to use the builder.

So, yes, the modification does change the behavior but because it's not really essential for the sake of the example, I didn't worry about it.

examples/lambda_transform.cpp
mairas: Same reply as above. :-D

examples/hysteresis.cpp
ba58smith: Got it. I like using the Builder anyway. Thx.

mairas:
Merging.

PR #238, opened 2020-11-05
Increase output buffer size on ESP32

mairas:
As reported by @KEGustafsson :

The web config UI wouldn't work with more than six sensors. The
culprit was found to be the output buffer size. It's now increased
to 4096 on ESP32. On ESP8266 memory is at premium and the amount
of sensors likely less anyway, so there the buffer size was left at
1024.

### DIFF
--- a/src/net/http.cpp
+++ b/src/net/http.cpp
@@ -175,10 +175,12 @@ void HTTPServer::handle_not_found(AsyncWebServerRequest* request) {
 }
 
 void HTTPServer::handle_config_list(AsyncWebServerRequest* request) {
+  // to save memory, guesstimate the required output buffer size based
+  // on the number of elements
+  auto output_buffer_size = 200 * configurables.size();
   AsyncResponseStream* response =
       request->beginResponseStream("application/json");
-  DynamicJsonDocument json_doc(1024);
-  // JsonObject root = json_doc.as<JsonObject>();
+  DynamicJsonDocument json_doc(output_buffer_size);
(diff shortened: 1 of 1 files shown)


ba58smith:
Looks good to me.

JohnySeven:
@mairas maybe it's time to add more defines for all the buffers we have? CONFIG, DELTA, TOKEN_TEST, etc?

Also could you make the define like:
```
#ifndef OUTPUT_BUFFER_SIZE
#ifdef ESP8266
#define OUTPUT_BUFFER_SIZE 1024
#elif defined(ESP32)
#define OUTPUT_BUFFER_SIZE 4096
#endif
#endif
```

So anyone can define the buffer size to respect to application they have? Lower it or make it even bigger? 

Thanks!

KEGustafsson:
If end user can define buffer size, that would be very handy. Already now I noticed that 4k limit is too small when I started to add more sensors to same device. I had to increased buffer to 8k and working fine. ESP32 is streaming now 29 paths and maybe many more could be added. Buffer size is the limiting factor now for config UI.

mairas:
Hmm, ok, seems like this is not a sufficient solution. I'm not excited about making the value user-configurable. Instead, in this case we can just take the size of the `configurables` map and estimate the size from that (say, 200 bytes per item).

mairas:
Finally got around to updating this. The amount of memory is now guesstimated from the number of configurables. This approach should work better on both ESP8266 and ESP32.

ba58smith:
Looks good!

PR #241, opened 2020-11-05
New blinkenlights

mairas:
This is a preview of new led controller and blinker implementation. There's now a dedicated LedController object that consumes notifications from Networking and WSClient objects and changes the led blinking pattern accordingly. 

Everything is hardcoded at the moment. Some refactoring should be done to allow changing or modifying the LedController object as well as connecting new consumers to the Networking and WSClient objects.

This is basically ready but waiting for PR#237 to be merged. This PR currently duplicates the commits of that PR -- if you want to look at the changes, they're all in the last commit.

### DIFF
--- a/src/sensesp_app.cpp
+++ b/src/sensesp_app.cpp
@@ -87,19 +87,17 @@ void SensESPApp::setup() {
 
   // create the websocket client
 
-
   this->ws_client =
       new WSClient("/system/sk", sk_delta, sk_server_address, sk_server_port,
                                  permission_strings[(int)requested_permissions]);
 
-  // create a led blinker and connect it to its data sources
+  // create a led controller and connect it to its data sources
 
-  led_blinker = new LedBlinker(led_pin, led_ws_connected, led_wifi_connected,
-                               led_offline);
+  led_controller = new LedController(led_pin);
+  this->networking->connect_to(led_controller);
+  this->ws_client->connect_to(led_controller);
+  this->ws_client->get_delta_count_producer().connect_to(led_controller);
 
-  this->networking->connect_to(led_blinker);
-  this->ws_client->connect_to(led_blinker);
-  this->ws_client->get_delta_count_producer().connect_to(led_blinker);
 }
 
 void SensESPApp::setup_standard_sensors(ObservableValue<String>* hostname,
@@ -134,8 +132,6 @@ void SensESPApp::setup_standard_sensors(ObservableValue<String>* hostname,
 }
 
 void SensESPApp::enable() {
-  this->led_blinker->set_wifi_disconnected();
-
   // connect all transforms to the Signal K delta output
 
   // ObservableValue<String>* hostname = networking->get_hostname();
--- a/src/system/led_blinker.cpp
+++ b/src/system/led_blinker.cpp
@@ -14,8 +14,14 @@ BaseBlinker::BaseBlinker(int pin) : pin{pin} {}
  */
 void BaseBlinker::set_state(bool state) {
   this->state = state;
+#ifdef ESP32
   digitalWrite(pin, state);
+#elif defined(ESP8266)
+  // LED (at least on ESP-01 and Wemos D1 mini) is active low
+  digitalWrite(pin, !state);
+#endif
   update_counter++;
+
 }
 
 /**
--- a/src/system/led_blinker.cpp
+++ b/src/system/led_blinker.cpp
@@ -22,7 +22,39 @@ void BaseBlinker::set_state(bool state) {
  * Invert the current LED state.
  */
 void BaseBlinker::flip_state() { this->set_state(!this->state); }
+
+/**
+ * Flip the LED off and on for `duration` milliseconds.
+ */
+void BaseBlinker::blip(int duration) {
+  // indicator for a blip being in progress
+  static bool blipping = false;
+
+  // only allow one blip at a time
+  if (blipping) {
+    return;
   }
+  blipping = true;
+
+  bool orig_state = this->state;
+  this->set_state(false);
+  int current_counter = this->update_counter;
+  app.onDelay(duration, [this, duration, orig_state, current_counter]() {
+    // only update if no-one has touched the LED in the meanwhile
+    if (this->update_counter == current_counter) {
+      this->set_state(true);
+      int new_counter = this->update_counter;
+      app.onDelay(duration, [this, orig_state, new_counter]() {
+        // again, only update if no-one has touched the LED
+        if (this->update_counter == new_counter) {
+          this->set_state(orig_state);
+        }
+        blipping = false;
+      });
+    } else {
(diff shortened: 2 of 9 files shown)


src/system/led_controller.cpp
```diff
@@ -0,0 +1,87 @@
...
+      break;
+    case kWifiManager:
+      this->set_wifimanager();
+      break;
+    default:
+      this->set_wifi_disconnected();
+      break;
+  }
+}
+
```
ba58smith, line 63: I assume you need to change the above cases to match the changes you made to the various Wifi statii, like making kDisconnected into kWifiDisconnected.

src/system/led_blinker.cpp
```diff
@@ -5,102 +5,89 @@
...
-    this->set_state(0);
-  }
+#define max(a,b) ((a)>(b)?(a):(b))
+
+BaseBlinker::BaseBlinker(int pin) : pin{pin} {}
+
+void BaseBlinker::set(bool state) {
+  this->state = state;
+  digitalWrite(pin, state);
 }
```
ba58smith, line 15: Could you name this `set_state`, so we know what's being set?

src/system/led_blinker.cpp
```diff
@@ -5,102 +5,89 @@
...
-    case kWifiManager:
-      this->set_wifi_disconnected();
-      break;
-    default:
-      this->set_wifi_disconnected();
-      break;
-  }
+void BaseBlinker::flip() {
+  this->set(!this->state);
 }
```
ba58smith, line 19: Could you name this `flip_state`, so we know what's being flipped?

src/system/led_controller.cpp
```diff
@@ -0,0 +1,87 @@
+#include "led_controller.h"
+
+// PATTERN: *___________________
+int no_ap_pattern[] = {50, 950, -1};
```
ba58smith, line 4: // Every pattern[] needs to end with -1. That's how PatternBlinker knows when to restart the pattern.

src/system/led_blinker.cpp
```diff
@@ -5,102 +5,89 @@
...
 
-void LedBlinker::set_wifi_connected() {
-  this->remove_blinker();
-  this->blinker =
-      app.onRepeat(wifi_connected_interval, [this]() { this->flip(); });
+void PatternBlinker::tick() {
+  if (!enabled) {
+    return;
+  }
```
ba58smith, line 77: Add comment here:
// When pattern[pattern_ptr] == -1, that's the end of the pattern, so start over at the beginning.

(Otherwise, there's no clue what it means for this to equal -1.)

src/sensesp_app.cpp
```diff
@@ -81,26 +81,25 @@ void SensESPApp::setup() {
...
+                                 permission_strings[(int)permissions]);
+
+  // create a led blinker and connect it to its data sources
+
+  led_blinker = new LedBlinker(led_pin, led_ws_connected, led_wifi_connected,
+                               led_offline);
+
+  this->networking->connect_to(led_blinker);
+  this->ws_client->connect_to(led_blinker);
+  this->ws_client->get_delta_count_producer().connect_to(led_blinker);
```
ba58smith: Does this do anything yet? I see code that sets delta_count_producer to 0 or 1, code that retrieves its value, and this code, which (I think) sets up the led_blinker to watch for changes to delta_count_producer. But I don't see the code that actually blinks the LED when something happens.

src/system/led_controller.cpp
JohnySeven: It would be nice to have some #define PATTERN_END -1

ba58smith: A much better solution than my suggestion to just add a comment, JohnySeven.

JohnySeven: @ba58smith I would add both :-)

src/system/led_controller.cpp
mairas: That should've been changed already; must've bungled up something when committing the changes. It wouldn't even compile like that, and I _did_ test the new features, at least a bit. 😉

src/system/led_blinker.cpp
mairas: Sure thing.

src/system/led_blinker.cpp
mairas: Roger.

src/system/led_controller.cpp
mairas: I'll do both. :-)

src/sensesp_app.cpp
mairas: This is again showing some older version of the file; these `connect_to`s should've pointed to `led_controller` instead of `led_blinker`.

FWIW, maybe it's not the greatest design in the world to use `connect_to` to connect to three different overloaded variants of one class method. I'll experiment with doing more manual connections with `attach`.

mairas:
This is good for proper review now.

src/system/led_controller.cpp
```diff
@@ -0,0 +1,103 @@
...
+// repeating until PATTERN_END is reached
+
+// PATTERN: *___________________
+int no_ap_pattern[] = {50, 950, PATTERN_END};
+
+// PATTERN: ******______________
+int wifi_disconnected_pattern[] = {300, 700, PATTERN_END};
+
+// PATTERN: **************______
+int wifi_connected_pattern[] = {700, 300, PATTERN_END};
```
ba58smith, line 13: Will it be possible to determine between a 300 / 700 pattern, and a 700 / 300 pattern?

src/system/led_blinker.cpp
```diff
@@ -22,7 +22,39 @@ void BaseBlinker::set_state(bool state) {
  * Invert the current LED state.
  */
 void BaseBlinker::flip_state() { this->set_state(!this->state); }
+
+/**
+ * Flip the LED off and on for `duration` milliseconds.
+ */
```
ba58smith, line 34: I can't figure out how / when "blip" is going to be used. It seems to turn the LED on/off really quickly (every 20 ms), but when will that happen? Every time an LED status indicator is changed? Is the blip used to indicate that the status is changing?
Can you add a bit of a comment to explain it? Thx.

mairas: Sent deltas were having the LED to flip. I was annoyed about that messing up the LED blink patterns. Instead, I wanted to have a way to have the LED flicker whenever a delta was sent, but resuming the original blink pattern afterwards. This is what I came up with. Even though the time is short, the effect is quite visible. It's called by `LedController::set_input(int ...)`, at the end of `led_controller.cpp`. I'd quite like to make it simpler but couldn't quite figure, how.

ba58smith: I like that! Thanks for explaining it.

mairas:
Merging.

PR #245, opened 2020-11-08
Blinkenlights controller

mairas:
Generalized `LedController` into a `VisualOutputController` that can be used to implement e.g. display output. It's also possible to provide the controllers via `SensESPAppBuilder`.


mairas:
Given no comments, merging.

PR #246, opened 2020-11-08
Update platformio.ini instructions in the README

ba58smith:
After spending a lot of time talking a new user through the process of creating a working `platformio.ini` file for a new board that's not in the example file, I rewrote the instructions in much greater detail.

I also updated the README to drop the references to the two unhandled exceptions for ESP32's, since those problems have now been resolved. 

I added a reference to the Wiki instructions for hard-coding the wifi credentials and the server address and port.


README.md
```diff
@@ -35,40 +35,47 @@ the Wemos D1 mini, an implementation of the ESP8266 chip. If you're using a diff
 board dropdown. If you're using an ESP32 that isn't in the board list, choose "Espressif ESP32 Dev Module".)
 The Arduino framework should become automatically selected. Complete the New Project dialog, then open the project you created.
 
-Once you have your new project open, open the `platformio.ini` file that's in your project's directory (NOT the one that you find if you go down into the .pio/libdeps/... folders). Add the SensESP dependency to the section for your board:
+Once you have your new project open, open the `platformio.ini` file that's in your project's directory (NOT the one that you find if you go down into the .pio/libdeps/... folders). Use the File_Save As... menu option to save the file as "platformio.ini.auto" (because it's the one that was automatically generated).
```
mairas, line 38: I'd much prefer to have the Getting Started section describe what _what_ needs to be done, not the specific details _how_ it is done. For that purpose, I think the section is already too detailed. I personally find it annoying and patronizing if the instructions try to walk me through the menus of some specific editor when I just want to know that I need to copy the example platformio.ini in place and modify it. It's of course fine and very useful to have such instructions, but they could be further down the file and in their own section (e.g. "Detailed Instructions").

ba58smith: With all due respect, I didn't write these instructions for you. Or jkoz, or Jan. I know you can figure out how to accomplish this task on your own. But I've spent far too much time, and typed detailed instructions into Slack far too many times, helping new SensESP users get a proper platformio.ini. I don't want to do that anymore.

I could write a version of these instructions for someone with your experience, and then ANOTHER set of instructions for someone who needs all the details, but why? Is an experienced developer really going to be that annoyed by having to read a dozen extra words of detail?

Look at this example: `Use File_New File to create a new, empty file.` If I don't issue that explicit instructions, and I say only `Create a new empty file.` - I can guarantee you that we'll get the following questions: "Where should I create the file?" and "What kind of file should I create?" and "What editor should I use to create the new file?"

I'll go through the instructions again, looking for the most condescending parts, and change them if I can do so in a way that will prevent questions like the above. But I don't think that will be very many changes.

mairas: I'm not saying we can't or shouldn't have instructions for total beginners. That's perfectly fine. What I tried to say was that there are pretty strong conventions what people with even a little software development background expect to see in the project documentation. And it's pretty essential those expectations are met; people are impatient and have short attention spans and easily continue looking for a project that they feel caters them better. And developers, especially experienced developers is what we need in this project. At the moment I'm pretty much alone working on improving the core codebase and debugging the build system issues and whatnot, and that doesn't scale, especially if we only want to get more total newbies onboard and ignore people with engineering skills.

As I suggested, this is not an either-or situation. We can cater for both, simply by first providing a project overview and further down in the document the newbie instructions.

ba58smith:
After I posted my comment yesterday, I thought about another approach: put "expert" instructions in README, with a reference to newbie instructions in the Wiki. Would you prefer that?

I'm fine writing the newbie instructions, and I'll make a stab at the expert instructions.

mairas:
Sure, that's one workable solution! The README should of course have a prominent link to the wiki in that case.

README.md
ba58smith: I moved the details to the Wiki, and tried to write a much simpler version of the instructions for the README. I may have made them TOO simplified, however. Please review, @mairas.

ba58smith:
Merging.

PR #247, opened 2020-11-09
Changed to a non-global fusion filter

BjarneBitscrambler:
The default fusion filter selected in the orientation sensor files has been changed from Adafruit_NXPSensorFusion (which was global) to Adafruit_Mahony (which is part of the class). This can save runtime RAM in cases where SensESP is not using the orientation sensor yet the orientation code is linked in anyway. Testing suggests this doesn't happen on ESP32 boards but may on ESP8266.

### DIFF
--- a/src/sensors/sensor_nxp_fxos8700_fxas21002.cpp
+++ b/src/sensors/sensor_nxp_fxos8700_fxas21002.cpp
@@ -15,10 +15,6 @@
 #define FXAS_CR1_ODR_BITS_LOC (2) //ODR bits are in positions 4:2
 #define N2K_INVALID_FLOAT (-1e-9) //NMEA2000 value for unavailable parameters
 
-Adafruit_NXPSensorFusion filter;  // when placed inside class, results from this
-                                  // filter return Heading=0.0 always. Other two
-                                  // filters (see below) work OK inside class
-
 //  Constructor creates an accelerometer/magnetometer object (fxos_)
 //  and a gyroscope object (fxas_)
 SensorNXP_FXOS8700_FXAS21002::SensorNXP_FXOS8700_FXAS21002()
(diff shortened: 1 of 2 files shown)


src/sensors/sensor_nxp_fxos8700_fxas21002.cpp
```diff
@@ -15,9 +15,9 @@
 #define FXAS_CR1_ODR_BITS_LOC (2) //ODR bits are in positions 4:2
 #define N2K_INVALID_FLOAT (-1e-9) //NMEA2000 value for unavailable parameters
 
-Adafruit_NXPSensorFusion filter;  // when placed inside class, results from this
+//Adafruit_NXPSensorFusion filter;  // when placed inside class, results from this
```
mairas, line 18: Please remove code instead of commenting it out. Code that is commented out reduces readability, and history is easily available on git, anyway.

src/sensors/sensor_nxp_fxos8700_fxas21002.h
```diff
@@ -28,7 +28,7 @@
 //   please support PJRC and open-source hardware by purchasing products
 //   from PJRC!  Written by PJRC, adapted by Limor Fried for Adafruit Industries.
 
-// Last edits by Bjarne Hansen 2020-10-20
+// Last edits by Bjarne Hansen 2020-11-09
```
mairas, line 31: This is redundant; you can get the information for each individual edit from Git.

mairas (changes requested):

src/sensors/sensor_nxp_fxos8700_fxas21002.cpp
BjarneBitscrambler: OK - done.

src/sensors/sensor_nxp_fxos8700_fxas21002.h
BjarneBitscrambler: Also gone.

BjarneBitscrambler:
It looks OK to merge now. I rebased the original edits and your requested changes into one (should have done it before pushing to my repository - hence the force-push).

mairas:
Splendid, thanks a lot! I'm relieved now. 😊 And yes, force-pushing is perfectly fine to clean PR commits!

Merging.

PR #248, opened 2020-11-11
VisualOutputController fixes

mairas:
Some modifications were required to make the VisualOutputController properly usable in external libraries


ba58smith:
A lot happening here, most of it over my head. I see nothing to comment on.

PR #252, opened 2020-11-14
Fixed a bug in Integrator Transform; gave it a reset() method

ba58smith:
Fixed a bug in the Integrator Transform - it didn't use the multiplier (`k`) in the calculation of `value`.

Added a public method to reset `value` to 0. Will be used in conjuction with a new LambdaConsumer to reset the value when the user does something (pushes a button, for example).


ba58smith:
No comments, and the changes are very small. Merging.

PR #253, opened 2020-11-17
Add a 100 ms timeout to the call to pulseIn. 

RBerliner:
The project will hang in a Security Token loop if the timeout is not present

### DIFF
--- a/src/sensors/ultrasonic_distance.cpp
+++ b/src/sensors/ultrasonic_distance.cpp
@@ -17,12 +17,13 @@ UltrasonicDistance::UltrasonicDistance(int8_t trig_pin, int8_t input_pin,
 void UltrasonicDistance::enable() {
   app.onRepeat(read_delay, [this]() {
     digitalWrite(trigger_pin, HIGH);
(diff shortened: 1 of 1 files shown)


RBerliner:
Something is wrong with this code. I will need to resubmit the request aqfter I figure out what it is

ba58smith:
I didn't look very closely, but I think you need to use an `unsigned long` to work with `micros()`.

RBerliner:
You are right - but my misunderstanding was more general. The delay in pulseIn is to be in microsec - not millisec. The code works with 50000 as the delay.

I have another problem - the pulseIn and notify statements in the original code have been replaced with 

this->emit(pulseIn(input_pin, HIGH));

I am trying to find where this function is documented and if it supports a timeout.

RBerliner:
It appears I am making too much out of it - I can probably just use the timeout in the emit(pulseIn...) but I need to point things to the head of SignalK/SensESP for testing. My platformio.ini is currently pointed to SignalK/Sensesp

PR #254, opened 2020-11-17
Add a 50 ms timeout to the pulseIn(...) statement.

RBerliner:
The code will hang with a Security Token loop when the sensor is not connected if this delay is not present.

### DIFF
--- a/src/sensors/ultrasonic_distance.cpp
+++ b/src/sensors/ultrasonic_distance.cpp
@@ -22,7 +22,7 @@ void UltrasonicDistance::enable() {
       yield();
     }
     digitalWrite(trigger_pin, LOW);
-    this->emit(pulseIn(input_pin, HIGH));
+    this->emit(pulseIn(input_pin, HIGH, 50000));  // 50 ms timeout
   });
 }
 


src/sensors/ultrasonic_distance.cpp
```diff
@@ -22,7 +22,7 @@ void UltrasonicDistance::enable() {
       yield();
     }
     digitalWrite(trigger_pin, LOW);
-    this->emit(pulseIn(input_pin, HIGH));
+    this->emit(pulseIn(input_pin, HIGH, 50000));  // 50 ms timeout
```
mairas, line 25: Given that this is a synchronous operation (everything stops until a response is received) and you're normally measuring short distances, wouldn't it be preferable to make the timeout as short as feasible? 50 ms corresponds to 8.5m back-and-forth in open air and far more in water. Assuming a absolute maximum range of 4.3 m for the sensor discussed in the example file, that would correspond to 25 ms, so maybe a 30 ms timeout would suffice?

mairas (commented):
If you prefer, I can merge as-is, too.

src/sensors/ultrasonic_distance.cpp
RBerliner: I looked again at the waveforms I observe from the device. The figure below is schematic. In practice 

PWM-Waveform

In practice I find that T2 is 15 ms and T3, could be as much as 25 ms (corresponding to 4.3 m) so I think that the 50ms delay is not out of place. I don't know how much these devices vary given that the datasheet gives T2 as 5 ms.

mairas: OK, thanks for the clarification! Merging.

PR #255, opened 2020-11-17
Make Integrator a template class

mairas:
Allow Integrator to handle more types than just floats.


ba58smith:
Since it now inherits from Symmetric, `Accumulator` is going to want to take an `int` in, and output an `int`. And `k` is an `int`. Which is OK as long as someone doesn't need a `k` of 1.54. My first thought is "Just run the output of `Acculator` through a `Linear` - but `Linear` is defined as expecting a `float` as input. 

Cross that bridge when we come to it, possibly with the "Typecast template transform" that you mentioned on Slack?

mairas:
Right... It won't be difficult to inherit Integrator from `Transform<C,P>` instead. Meaning that we'd define the input and output types separately. Just more moving parts in that case, but maybe it that wouldn't matter as long as there are typedefs for the simple cases available?

ba58smith:
"It won't be difficult to inherit Integrator from Transform<C,P> instead." 
I like that idea. There probably aren't too many cases for Integrator not to be Symmetric: int to float, bool to int, maybe int to bool. Certainly NOT float to int, or anything with String as either input or output.

mairas:
I'll make it inherit from `Transform<C,P>` then!

ba58smith:
@mairas - just a nudge. I'm waiting on this PR for something else I'm working on. :-)

ba58smith (approved):
This works as-is, but as we discussed, it's not as flexible as it could be. If you don't have time to change it from Symmetric to Transform<C, P>, you can merge this, then I'll do another PR to make that change. (I'm waiting on this for something else I'm working on.)

mairas:
Oh, sorry! I managed to miss your nudges here... Will do it right away.

ba58smith:
Looks good to me! To be clear, someone can now use `IntegratorT<int, float>` and it will work? (That was my reason for wanting this to inherit from Transform<C, P>.)

mairas:
> Looks good to me! To be clear, someone can now use `IntegratorT<int, float>` and it will work? (That was my reason for wanting this to inherit from Transform<C, P>.)

Yes, as long as you can multiply the C and P types together and the result is P, anything goes.

mairas:
Merging then.

PR #257, opened 2020-11-17
Add steps to README to deal with PlatformIO bug

ba58smith:


mairas:
The text is good, but it seems like the issue actually got fixed in PlatformIO 5.0.3 that was released a week ago. At least my test case failed to fail any longer. :-) I've closed the upstream bug report. I'll close this PR as well; let's reopen if the issue still persists.

ba58smith:
I was also able to successfully build my ESP32 Project in PlatformIO 5.0.3 without first having to build the ESP8266 version of it. So it does appear to no longer be a problem.

PR #258, opened 2020-11-18
Split frequency from Wire.begin() for ESP8266 compatibility

### LINKED ISSUE #256: TwoWire::begin() has conflicting definitions in ESP32 and ESP8266 versions of Wire library
## Problem
The Wire libraries for the ESP32 and ESP8266 differ in an important manner, despite being written by the same author.  The `TwoWire::begin()` method for the ESP32 has prototype:
```
bool TwoWire::begin(int sdaPin, int sclPin, uint32_t frequency);
```
whereas for the ESP8266 it is:
```
void TwoWire::begin(int sda, int scl, uint8_t address);

```
## Impact
The consequence is that when calling TwoWire::begin() for an ESP32 with a reasonable frequency (say 400000 Hz) parameter, it will on the ESP8266 instead be interpreted as starting a connection with address 0x80 (i.e. 0x061a80 is truncated to a uint8_t), at whatever the default frequency is, which likely will not be what is desired. 

## Fix
One workaround is to split up the `begin()` call into two:  a call to begin() specifying only the SDA and SCL pins, and a second call to `setClock()`.  In this form, both calls have the same meaning in both libraries.

Another workaround (less preferable I'd say) is to have conditional-compilation #ifdef / #endif around two versions of the code.

## History

@ba58smith pointed out (thanks!) a warning seen in when compiling the orientation code for the ESP8266, which I hadn't seen when targetting the ESP32. 
Re: https://github.com/SignalK/SensESP/issues/249#issuecomment-729140276_
```warning: large integer implicitly truncated to unsigned type [-Woverflow]
   Wire.begin(pin_i2c_sda, pin_i2c_scl, 400000 );
```

BjarneBitscrambler:
Issue #256 noted that the `begin()` method of the Wire library differs between the ESP8266 and ESP32 versions. This PR removes the frequency from the begin() call, and instead makes a separate call to `setClock()`  The code is now compatible with both versions of the Wire library and fixes #256.

I've verified that it compiles without warnings for the d1_mini and ESP32-Wrover, and runs on the Wrover. I don't have a board to test the d1_mini executable (but have one ordered).

### DIFF
--- a/src/sensors/sensor_nxp_fxos8700_fxas21002.cpp
+++ b/src/sensors/sensor_nxp_fxos8700_fxas21002.cpp
@@ -42,9 +42,10 @@ bool SensorNXP_FXOS8700_FXAS21002::connect(int pin_i2c_sda,
     }
   }//end of loading calibration
 
-  Wire.begin(pin_i2c_sda, pin_i2c_scl, 400000 );
+  Wire.begin(pin_i2c_sda, pin_i2c_scl);
+  Wire.setClock(400000);  // separate from begin() for ESP82666 compatibility
   if (!startSensors()) {
-      debugE("Failed to find sensors");
+    debugE("Failed to find sensors");
     return false;
   }
 


ba58smith:
Looks good to me!

mairas:
Yep, this was a good solution! Thanks! Merging.

PR #260, opened 2020-11-19
Initial version of SkPutRequestListener<>

joelkoz:
A ValueProducer that listens to incomming PUT requests. This was required because the Signal K spec requires devices that publish paths and expect to be manipulated via PUT requests need to respond to a websocket broadcast different from an "update" from a subscription, aas well as send a response

### DIFF
--- a/src/signalk/signalk_put_request_listener.h
+++ b/src/signalk/signalk_put_request_listener.h
@@ -1,0 +1,67 @@
+#ifndef _signalk_putrequestlistener_H_
+#define _signalk_putrequestlistener_H_
+
+#include <ArduinoJson.h>
+
+#include <set>
+
+#include "Arduino.h"
+#include "sensesp.h"
+#include "system/configurable.h"
+#include "system/observable.h"
+#include "system/valueproducer.h"
+
+/**
+ * A Signal K listener is one that listens for Signal K stream deltas
+ * and notifies of value changes
+ */
+class SKPutListener : virtual public Observable {
+ public:
+  /**
+   * The constructor
+   * @param sk_path The Signal K path that identifies the signal k path
+   * this particular listener responds to
+   */
+  SKPutListener(String sk_path);
+
+  String& get_sk_path() { return sk_path; }
+
+  virtual void parse_value(JsonObject& put) = 0;
+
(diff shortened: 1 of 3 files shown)


JohnySeven:
Looking good 👍

src/signalk/signalk_put_request_listener.h
```diff
@@ -0,0 +1,67 @@
...
+#include "Arduino.h"
+#include "sensesp.h"
+#include "system/configurable.h"
+#include "system/observable.h"
+#include "system/valueproducer.h"
+
+/**
+ * A Signal K listener is one that listens for Signal K stream deltas
+ * and notifies of value changes
+ */
```
ba58smith, line 17: Will this style of comment result in a Doxygen description of this class? We're trying to adhere to the Doxygen format for all new code (and to update existing code as it's modified in any way).

src/signalk/signalk_put_request_listener.h
```diff
@@ -0,0 +1,67 @@
...
+ * for the specified SignalK path. The value is then emitted for
+ * further processing.
+ */
+template <class T>
+class SKPutRequestListener : public SKPutListener, public ValueProducer<T> {
+ public:
+  SKPutRequestListener(String sk_path)
+      : SKPutListener(sk_path) {
+    if (sk_path == "") {
+      debugE("SKPutRequestListener: User has provided no sk_path to respond to.");
```
ba58smith, line 53: Minor comment, but I don't think you need to put "SKPutRequestListener" in the string of output, because `debug` always outputs the name of the calling function. (I think.)

ba58smith (approved):
I don't pretend to be familiar enough with Json to adequately review all of this. So I'll approve it, assuming @mairas will review it before merging.

src/signalk/signalk_put_request_listener.h
joelkoz: Well, that is the original JavaDoc format, which I used to generate the Doxygen docs when I last did it. Unless something changed explicitly, Doxygen recognizes JavaDoc format.

mairas:
LGTM. SK specification put documentation wasn't rendering correctly so I couldn't cross-check against that, but given that you were testing with external components and doing some back-and-forth with Scott, I'll take your word for it! Merging.

PR #267, opened 2020-11-24
Fix EspSoftwareSerial compile error

joelkoz:
EspSoftwareSerial should only be explicitly included when compiling for ESP32

### DIFF
--- a/library.json
+++ b/library.json
@@ -68,6 +68,10 @@
       "version": "https://github.com/sivar2311/ESPTrueRandom"
     },
     {
+      "name": "elapsedMillis",
+      "version": "https://github.com/pfeerick/elapsedMillis"
+    },  
+    {
       "name": "Wire"
     },
     {
@@ -105,9 +109,6 @@
     },
     {
       "name": "Adafruit AHRS"
-    },
-    {
-      "name": "EspSoftwareSerial"
     },
     {
       "name": "RemoteDebug",
--- a/library.json
+++ b/library.json
@@ -111,6 +111,10 @@
       "name": "Adafruit AHRS"
     },
     {
+      "name": "EspSoftwareSerial",
+      "platforms": ["espressif32"]
+    },
+    {
       "name": "RemoteDebug",
       "version": "https://github.com/JoaoLopesF/RemoteDebug.git#0b5a9c1a49fd2ade0e3cadc3a3707781e819359a" }
   ],
--- a/platformio.ini
+++ b/platformio.ini
@@ -25,6 +25,8 @@ default_envs =
 framework = arduino
 lib_ldf_mode = deep
 monitor_speed = 115200
+
+[common_env_data]
 lib_deps =
    ReactESP
    ESP8266WebServer
@@ -39,6 +41,7 @@ lib_deps =
    OneWire
    DallasTemperature
    https://github.com/sivar2311/ESPTrueRandom
+   https://github.com/pfeerick/elapsedMillis
    Wire
    Adafruit ADS1X15
    Adafruit AHRS
@@ -50,7 +53,6 @@ lib_deps =
    Adafruit SHT31 Library
    Adafruit INA219
    Adafruit MAX31856 Library
-   EspSoftwareSerial
    https://github.com/JoaoLopesF/RemoteDebug.git#0b5a9c1a49fd2ade0e3cadc3a3707781e819359a
 
 [espressif8266_base]
@@ -61,6 +63,8 @@ build_flags =
    -Wall
    -Wno-reorder
 extra_scripts = extra_script.py
+lib_deps=
+   ${common_env_data.lib_deps}
 
 [env:d1_mini]
 extends = espressif8266_base
@@ -78,6 +82,9 @@ platform = espressif32
 build_unflags = -Werror=reorder
 board_build.partitions = min_spiffs.csv
 monitor_filters = esp32_exception_decoder
+lib_deps=
+    ${common_env_data.lib_deps}
+    EspSoftwareSerial
 
 [env:esp32dev]
 extends = espressif32_base


library.json
```diff
@@ -67,6 +67,10 @@
       "name": "ESPTrueRandom",
       "version": "https://github.com/sivar2311/ESPTrueRandom"
     },
+    {
+      "name": "elapsedMillis",
```
mairas, line 71: This should probably be in a later PR, but no biggie.

library.json
```diff
@@ -106,9 +110,6 @@
     {
       "name": "Adafruit AHRS"
     },
-    {
```
mairas, line 109: Instead of unconditionally removing the dependency and thus breaking builds that depend on the library, keep the entry and add `"platforms": "espressif32"` to it.

platformio.ini
```diff
@@ -61,6 +63,8 @@ build_flags =
    -Wall
    -Wno-reorder
 extra_scripts = extra_script.py
+lib_deps=
```
mairas, line 66: I think the config items are inherited by default. But no harm done, of course.

mairas (commented):
OK, except for the unconditional EspSoftwareSerial removal from library.json which breaks builds.

library.json
joelkoz: Except this doesn't unconditionally break builds.  EspSoftwareSerial is ALREADY part of the esp8266 framework tree.  That is why this causes an error in the first place - it it getting included again.  It does NOT exist in the ESP32, which is why its needed in the first place.

To see this in action, you may want to do a full side-by-side view of the files vs. just the changes. What I did was move the list of dependencies to a `[common-env-values]` section, and then put a references to that in the esp8266 section, and again in the esp32 section - but ADDING the EspSoftwareSerial reference where it is needed.

I was able to build the libraries for both esp8266 as well as esp32 by doing this.  I was also able to build to test projects that I have - one for wemos, the other for a heltec32, and they both built without error or need for modification.

platformio.ini
joelkoz: Again, the fix is to put the list of dependencies in a new "common" section that is NOT inherited, and then to reference it in the two platform specific sections.  This has the net effect of removing the EspSoftwareSerial reference from esp8266 builds, but not from esp32 builds

library.json
joelkoz: I agree. Its really there for the RepeatReport fix, but it was already enough work to break the big PR into lots of smaller ones that I thought this one time it would be ok

platformio.ini
```diff
@@ -25,6 +25,8 @@ default_envs =
 framework = arduino
 lib_ldf_mode = deep
 monitor_speed = 115200
+
+[common_env_data]
 lib_deps =
    ReactESP
    ESP8266WebServer
```
joelkoz, line 32: THIS is the key to this PR. This `lib_deps` used to be in `[env]`, which is inherited by all.  Now its separated

library.json
joelkoz: Ok - so now I see that you can specify platform dependencies in both library.json, as well as platformio.ini.  Do you know the difference between the two? Certainly with this PR as is, I am able to build both the SensESP library, as well as two projects that are dependent upon it, both using both esp8266 and esp32 devices.  I can go ahead and add it back to library.json was well, but it doesn't seem to make a difference as far as I can tell.

library.json
```diff
@@ -110,6 +110,10 @@
     {
       "name": "Adafruit AHRS"
     },
+    {
+      "name": "EspSoftwareSerial",
+      "platforms": ["espressif32"]
```
joelkoz, line 115: Added back in per PlatformIO specs.

ba58smith (approved):
Looks like you two have hammered this out, and I see nothing to add.

library.json
mairas: Yes, I do know the difference of those two files, thank you very much for asking.

platformio.ini
mairas: OK, I didn't look at the espressif8266_base section close enough. I was assuming it was extending the section you defined. And FWIW, you could've just kept the lib_deps in their original env section and then overridden that in espressif32_base.

library.json
ba58smith: Just in case you thought @joelkoz was being literal, and was questioning your basic understanding of how to configure PlatformiIO... let me interpret, and rephrase, his question: "Do you know the difference between the usage of library.json and platformio.ini for configuring libraries? Because I'm confused." :-)

joelkoz: @mairas Butch is right - I was not questioning your knowledge.  I DON'T know the difference.  I spent a little time reading up, and I THINK library.json is for building the library when used externally, and the platformio.ini is for the IDE, but it seems like there may be more to it. One would think the IDE would be capable of reading library.json also.  Having to put entries in two places seems to violate the "don't repeat yourself" rule.  In any event, I added it back to library.json and used your suggestion to limit it.

platformio.ini
joelkoz: To override in espressif32_base, wouldn't I have had to copy/paste the entire lib_deps section and then drop the one line we don't want?  Is there a way to remove an entry already specified?  I took this approach from the platformio docs example on how to manage multiple platforms in the same .ini file.

joelkoz: Oh wait - you meant "go ahead remove the `EspSoftwareSerial` reference BUT leave it in the [env] section", and then just extended it in espressif32_base (not overridden it completely).

mairas: Yeah, that was what I meant but it doesn't really matter. This is fine.

mairas:
Good now. Merging.

ba58smith:
This doesn't seem to fix the problem, but only because of the square brackets in this:
```
      "name": "EspSoftwareSerial",
      "platforms": ["espressif32"]
```
If you look higher up in the file, you see another instance of specifying a platform, and it doesn't use the square brackets:
```
      "name": "ESP8266mDNS",
      "platforms": "espressif8266"
```
I made that change locally, deleted both `/EspSoftwareSerial` directories, and rebuilt for both ESP32 and ESP8266, and it worked.
@joelkoz - can you test this again?

joelkoz:
That is weird. I thought I read in the specs you could specify one or more platforms using an array, which is the syntax I used. It built for me the way it is.  You are building just the library itself?  Or are you building an external project?

What version of PIO are you using? I just upgraded to 1.36 last week. 

I will try it again. Can you put it back and try again to make sure it breaks? It would make sense that one should be able to specify more than one platform to use, so the array should work, but since in practice we only need this for one, it doesn’t hurt to use the alternate syntax. Still, I’d like to know for sure what is going on.

ba58smith:
Well, now I can't get my D1 Mini to compile at all if EspSoftwareSerial is in library.json - with or without the square brackets around espressif32.
HOWEVER, I looked at PlatformIO documentation and found another way to exclude a library from a particular platform, and it seems to work:
```
[espressif8266_base]
;this section has config items common to all ESP8266 boards
platform = espressif8266
board_build.ldscript = eagle.flash.4m1m.ld
build_flags =
; deprecated:   -Wl,-Teagle.flash.4m1m.ld
   -Wall
   -Wno-reorder
lib_ignore = EspSoftwareSerial 
```
And that's without any qualifier under "EspSoftwareSerial" in `platformio.ini`. It's just in there, like any other library name.

joelkoz:
@ba58smith - For clarification - what won't compile?  When you build from the SensESP dev lib (as if you were testing the library)? Or when you build an external project?  I am assuming its an "external project", which means its looking at SensESP's `library.json` file for the dependency list.

Can you verify that the `library.json` in the .pio/libdeps/d1_mini/SensESP` dir has the brackets in your specific project?

ba58smith:
@joelkoz - I did a fresh pull of the latest code from SensESP, deleted the .pio directory, then let PIO do the Intellisense Index Rebuild to bring in all the files. Then I looked at the `library.json` in my `d1_mini` Project, and the one in my `esp32dev` Project. Both of them had the square brackets around `espressif32`. Then I tried to build both Projects. The d1_mini build failed with the error about SoftwareSerial, but the esp32dev build succeeded.

Then I deleted .pio, and let the Intellisense Index Rebuild run again, and then I removed the brackets around `espressif32` in the `library.json` of each Project. Then I tried to build both Projects, and the d1_mini build failed again, with the same error.

Then I deleted .pio, and let the Intellisense Index Rebuild run again. I removed the whole `"platforms": ["espressif32"]` line from the `library.json` of both Projects. Then I modified the `platformio.ini` of the Project (not the `platformio.ini` of SensESP) to include `lib_ignore = EspSoftwareSerial` as noted above. Then I built them both successfully.

ba58smith:
I verified this one more time, and then created PR #283 to fix it.

PR #268, opened 2020-11-24
LoadController with truth input

joelkoz:
LoadController to manage a digital output connected to a relay, and a transform with dealing with "on/off" in various string formats.


mairas:
I may not be following but isn't the LoadController merely a controller (consumer) that sets the state of a digital output pin based on the input? That's unquestionably a valuable entity to have, but couldn't the name be simply, um, maybe DigitalOutput? Or DigitalOutputController?

Anyway, we should establish vocabulary for the output side of things. They're consumers, but that's just describing their relation to the data flow. What are controllers? In MVC, it's something that "Accepts input and converts it to commands for the model or view." Do we want that to refer to something that merely updates the state of some output device or performs more complex interpretation of the input and takes some action accordingly. I've used the term in LedController and VisualOutputController (as well as SSD1306DisplayController which is in a separate repo), and they're more of the latter kind. I now realize, though, that those names are pretty bad and would be happy to change them. They're controllers, for sure, but they output some very specific kind of information (device status) which should be present in the name.

ba58smith:
If you want to give it a very literal name, isn't it a "DigitalOutputPinController"? Kind of a long name, but it's accurate.

Other possible names: DigitalPinController, DigitalOutputController, DigitalPinStateController, DigitalPinStateSetter.

(I have to admit, "LoadController" always seemed a bit "off".)

joelkoz:
Getting back to `LoadController` here.  Now that I think about it, after all the refactoring that has occurred, the only difference between `LoadController` and the currently existing `DigitalOutput` is `DigitalOutput` is a ValueConsumer<bool> only, and `LoadController` is a BooleanTransform.  A simple refactor of DigitalOutput to pass its value along in addition to setting the pin solves this problem.  Any objections?

joelkoz:
Re: the use of "Controller", I do understand that it is ambiguous in the context of MVC, but we don't really use MVC in SensESP.  We don't have any models and views, so IMO we can claim the "Controller" definition for our own use.  Changing it to simply "Control" might eliminate the problem.  Even with the elimination of `LoadController` for an enhanced `DigitalOutput`, I still have a `SmartSwitchController`.  It more closely follows the definition of "Complex interpretation of inputs to outputs"

joelkoz:
Closing this unmerged as I had a problem with rebasing.  Will re-open as a new PR.

PR #269, opened 2020-11-24
Transform to repeat the reporting of a value at regular intvervals

joelkoz:
RepeatReport() is a transform for values that do not change frequently, but to still ensure they can be reported to another consumer (like the SK server) at frequent intervals.


joelkoz:
This PR depends on update to library. Going to resubmit

PR #270, opened 2020-11-24
Added RepeatReport<> transform

joelkoz:
RepeatReport() is a transform for values that do not change frequently, but to still ensure they can be reported to another consumer (like the SK server) at frequent intervals.

### DIFF
--- a/src/transforms/repeat_report.h
+++ b/src/transforms/repeat_report.h
@@ -6,20 +6,22 @@
 #include "transforms/transform.h"
 
 /**
- * RepeatReport ensures that values that do not change frequently are still reported at
- * at a specified maximum silence interval. If the value has not changed in max_silence_interval 
- * milliseconds, the current value is emmitted again, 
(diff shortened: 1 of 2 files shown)


src/transforms/repeat_report.h
```diff
@@ -0,0 +1,35 @@
...
+ * at a specified maximum silence interval. If the value has not changed in max_silence_interval 
+ * milliseconds, the current value is emmitted again, 
+ */
+template <typename T>
+class RepeatReport : public SymmetricTransform<T> {
+ public:
+  RepeatReport(long max_silence_interval = 15000, String config_path = "") :
+     SymmetricTransform<T>(config_path),
+     max_silence_interval{max_silence_interval} {
+     this->load_configuration();
```
mairas, line 19: Could you reformat to fix the indentation?

mairas (commented):
This is fine once reformatted and once the commits of the preceding PR are removed.

PR #271, opened 2020-11-24
Support for color LEDs

joelkoz:
LEDLight handles RGB leds on both esp8266 and esp32


src/system/led_light.cpp
```diff
@@ -0,0 +1,150 @@
...
+
+  this->load_configuration();
+}
+
+#ifdef ESP32
+
+// Code for analogWrite() simulation on ESP32 inspired by
+// https://github.com/ERROPiX/ESP32_AnalogWrite
+
+int8_t channel_to_pin[16] = {-1, -1, -1, -1, -1, -1, -1, -1,
```
mairas, line 28: Remove the globals.

src/system/led_light.cpp
```diff
@@ -0,0 +1,150 @@
...
+                             -1, -1, -1, -1, -1, -1, -1, -1};
+
+#define CHANNEL_FREQUENCY 5000
+#define CHANNEL_RESOLUTION 13
+
+static int pin_to_channel(uint8_t pin) {
```
mairas, line 34: Needs documentation.

mairas: PWMs might well be used outside of this code in which case the whole concept breaks. And forcing every user to reserve the channel using this function is somewhat too opinionated for my taste.

src/system/led_light.cpp
```diff
@@ -0,0 +1,150 @@
...
+        ledcAttachPin(pin, channel);
+        break;
+      }
+    }
+  }
+
+  return channel;
+}
+
+static void analogWrite(uint8_t pin, uint32_t value, uint32_t valueMax = 255) {
```
mairas, line 61: Add documentation.

src/system/led_light.cpp
```diff
@@ -0,0 +1,150 @@
...
+    uint32_t duty = ((levels - 1) / valueMax) * min(value, valueMax);
+
+    // write duty to LEDC
+    ledcWrite(channel, duty);
+  }
+}
+
+// Calculate the PWM value to send to analogWrite() based on the specified
+// color value.
+static int get_pwm(long rgb, int shift_right) {
```
mairas, line 76: Could the approach of using #ifdefs to select between two different implementations be reconsidered? It makes creating Doxygen documentation more difficult, following the code more difficult and reduces code maintainability because changes need to be implemented in two subtly different versions of the same function.

src/system/led_light.h
```diff
@@ -0,0 +1,95 @@
+#ifndef _led_light_H_
+#define _led_light_H_
+
+#include "system/configurable.h"
+#include "system/valueconsumer.h"
+
+/**
+ * LEDLight is a special device object that can be used to control
```
mairas, line 8: I think the class name is misleading. It makes me wonder is this about setting an individual RGB LED or controlling room illumination. Looking at the implementation, to me it seems this is just about setting RGB LED colors, with monochrome LED support added as an afterthought. The class name could be something like RgbLed (or RGBLED?).

Also, I think the monochrome LED case should be a class of its own and both RgbLed and MonochromeLed would inherit from the same base class.

src/system/led_light.h
```diff
@@ -0,0 +1,95 @@
...
+ * color to be displayed via the set_input(long), or by setting the
+ * on/off state via set_input(bool). When specifying a simple on/off
+ * via the the bool input, the default colors specified in the 
+ * constructor are used.
+ */
+class LEDLight : public Configurable,
+                 public ValueConsumer<long>,
+                 public ValueConsumer<bool> {
+                     
+    public:
```
mairas, line 25: Indentation off; pls reformat the file.

mairas (changes requested):

src/system/led_light.cpp
joelkoz: EVERYTHING in this section of the `#ifdef ESP32` should be considered "private, static".   This entire section is necessary to support PWM on an ESP32.  That fact that we even NEED PWM to control leds shouldn't be of concern to the end user IMO.  I wanted a simple class that was "display the rgb color number `0xrrggbb`. I wrote it for the Wemos, and all was fine until I went to contribute the code and discovered that the standard Arduino function analogWrite() has been "on the todo list" of the ESP32 devs for about 3 years now. :(

One simple solution is to conditionally include in the library.json, only for ESP32, the github project https://github.com/ERROPiX/ESP32_AnalogWrite .  That project is a simple pollyfill, and it adds analogWrite() back to the API.  If I included that library, this entire #ifdef section would disappear, and the implementation between esp8266 and esp32 would be identical, and there would be no need for the conditional compilation.  What I
 found, however, was that library had some other functionality that I don't think we needed, so I simply copied the interesting parts out of it, refactored it slightly, and put it here.

One thing I could do would be to move this code section to its own analog_write_esp32.h and .cpp file.  That would move it out of this area entirely, and also make the polyfill available to other transforms and devices in the future.  I personally like this approach the best.

src/system/led_light.cpp
joelkoz: This is a static method in the .cpp file, so it is not visible to any developer outside of someone looking at the SensESP source code.  It will not (and really should not) be picked up by Doxygen.  I can, however, add a line that indicates its a polyfill.

src/system/led_light.cpp
joelkoz: The more I think about it, the more I think I want to move this entire thing to `system/analog_write_esp32.h` and .cpp.  That way, no one else has to fight the fact that analogWrite() is missing on the ESP32.  Any objection to that?

src/system/led_light.h
joelkoz: It is exclusively about controlling a color rgb led by simply setting the color.  Monochrome wasn't really an afterthought, but I'll admit its squeezed in here a bit clumsy.  This code started its life inside no defunct SmartSwitch class, which was a much higher level function class.  I wanted that class to be flexible enough to handle switches that may or may not have a status light at all, or to handle a monochrome led if you couldn't afford to give up 2 additional pins on the MCU to control the other colors.

Now that this class is out on its own, I think I'll just remove the monochrome support entirely and just rename it either ColorLED, or LEDColor, of RgbLed, or whatever.  I really don't have a preference. If you prefer RgbLed, I'll go with that.

For monochrome led, users can use the pre-existing DigitalOutput, since we can now wire this stuff up in the `main.cpp` vs. it being hidden inside the implementation.

ba58smith (approved):
Looks like @mairas has touched on everything I would have commented on, and more.

I'll give my approval to merge, assuming that won't happen until after Matti has given his.

joelkoz:
Ok, lets give this version a try.  I have renamed LEDLight to RgbLed.  Support for monochrome leds have been removed (those are nothing more than DigitalOutput() and thus can be done elsewhere), and I have removed all of the PWM code for ESP32.  Instead, that code has been moved to a separate modue named `analog_write_esp32.h`  The sole purpose of that module is to add the missing analogWrite() API call back to the ESP32 framework, so SensESP code can be written for both platforms without any of the programming gymnastics I had to do before.  This makes any kind of PWM output available to other devices should the need arise.

src/system/analog_write_esp32.cpp
```diff
@@ -0,0 +1,66 @@
...
+
+
+// Code for analogWrite() simulation on ESP32 inspired by
+// https://github.com/ERROPiX/ESP32_AnalogWrite
+
+// ESP32 simulates PWM by connecting a 16 channel timer built
+// into the chip to a digital output pin. This array handles
+// the channel reservations for each pin the user tries
+// top write to using a lazy channel reservation system.
+static int8_t channel_to_pin[16] = {-1, -1, -1, -1, -1, -1, -1, -1,
```
mairas, line 13: You must refactor the code to remove the global array before this PR can be merged.

src/system/analog_write_esp32.cpp
```diff
@@ -0,0 +1,66 @@
...
+
+#define CHANNEL_FREQUENCY 5000
+#define CHANNEL_RESOLUTION 13
+
+// Translate the digital pin the user is writing to
+// to the timer channel the pin is attached to. This
+// function is guaranteed to succeed - if no previous
+// channel reservation has been made, one is made
+// automatically.
+static int pin_to_channel(uint8_t pin) {
```
mairas, line 24: No, you can't rely on your local channel_to_pins array for global resource reservations. You might have other code paths (imported libraries, whatever) using PWM and allocating PWM channels. You have to provide provisions for explicit channel allocation.

src/system/analog_write_esp32.cpp
joelkoz: Perhaps we have a different definition of "global array". This array is defined as `static`, so it is only visible to this particular .cpp file.  Since this module defines a global API function and not a class, there is no way to make this "not global."  It is really no different than a static member of a class.

src/system/analog_write_esp32.cpp
joelkoz: So here is the delimma: the timer channels is something that is unique to an ESP32.  There is no such concept on the esp8266.  Requiring the end user to do some sort of channel reservation in their code would require them to add ESP32 specific code to their app.  I would argue that the approach of adding a polyfill allows for better code portability. It allows users to share code and compile freely for the MCU of choice without any additional work.

I am open to suggestions on how to handle this in a way that doesn't have the RgbLed class use a completely different signature when compiling for ESP32 than for ESP8266. About the best I can come up with would be an extra parameter that allows an explicit channel to be specified IF the programmer wants to.  The default would be -1, which would mean they don't care and they want the system to pick one.  The parameter would be meaningless and thus ignored when compiling for ESP8266.  Doing so would add conditional compilation to the RgbLed class however, and I thought we wanted to avoid that.

One other approach would be to expose an additional function here that allows ESP32 developers to reserve channels ahead of time. That would allow those developers who have an awareness/need to say "pin X is attached to channel Y" to do so, but for the large majority of the hobbyist users who will be using SensESP, they will have no idea nor no need to even be aware of this issue.

src/system/analog_write_esp32.cpp
mairas: It has file scope, yes, but according to my understanding it still reserves that amount of memory for every single SensESP program, whether they use color leds or not. That's not good, and we've tried to systematically get rid of global variables from the codebase. This PR should not be an exception.

src/system/analog_write_esp32.cpp
mairas: Assuming that the RgbLed is the special and not the general case and if you want to follow the automatic reservation approach, you could do a ChannelReservation class that the user would instantiate to define the channels reserved for the RgbLed use. That could then be passed to the RgbLed constructor to provide it access to those channels. The default value of the object could be NULL for ESP8266.

However, since this is obviously becoming a contentious issue on a relatively specialized functionality, I wonder wouldn't it be better to keep the implementation separate of SensESP? Just like how I did with https://github.com/mairas/SensESPStatusDisplay. The code can be trivially taken into use but it doesn't add any maintenance burden to the SensESP codebase. That'd also allow you to use your exact current implementation without the global variable issues I raised.

joelkoz: I could do either way.  The RgbLed is not critical in the slightest to a smart switch implementation anymore, since this status display has been separated out from what is now the `LoadController`.

I don't really have time to mess with this right now, so for now we can leave this PR open in draft format.  I do think that rgb control is useful to others.

Maybe what we need is a class like `PWMController`.  Rather than trying to add analogWrite() to ESP32, we can add the concept of setting up to 16 channels that you can set a PWM value on for to ESP8266.  The ESP8266 implementation could be a simple pass through to analogWrite().  Any type of explicit channel reservation method would be optional, and for ESP8266, if it WAS used it wouldn't really do anything, as are there are no physical channels to reserve

src/system/analog_write_esp32.cpp
joelkoz: Ah!  Well, if that is the actual objection, this could be refactored to do a lazy allocation.  That changes a lot of my implementation thinking.  I thought there was just some type of ideological objection to global variables to avoid things like name clashing.

joelkoz:
Closing this implementation and adding new PR for new implementation

PR #272, opened 2020-11-24
Button and ClickType

joelkoz:
Classes for handling button presses


src/sensors/button.h
```diff
@@ -0,0 +1,35 @@
...
+/**
+ * Button represents a switch or button connected to a digital input
+ * pin. Button emits a value only when the state of the button changes
+ * (i.e. when the button changes from not pressed to pressed, and vice versa).
+ * However, if the button is pressed and held longer than repeat_start_interval 
+ * milliseconds, the press will be emitted once again, and then again every 
+ * repeat_interval milliseconds until the press is released. To
+ * disable this feature, set repeat_start_interval to -1.
+ */
+class Button : public DigitalInput, public BooleanProducer {
```
mairas, line 16: As discussed, `Button` is too broad (and misleading) a name for DigitalInput that merely repeats its output.

src/transforms/click_type.cpp
```diff
@@ -0,0 +1,128 @@
...
+    long_click_delay{long_click_delay},
+    double_click_interval{double_click_interval},
+    ultra_long_click_delay{ultra_long_click_delay},
+    queued_report{NULL} {
+   load_configuration();      
+}
+
+
+void ClickType::set_input(bool input, uint8_t inputChannel) {
+  if (input) {
```
mairas, line 18: As mentioned before, this needs reimplementing. Logic is convoluted, state is maintained (I think) in multiple variables that are not described, etc. I believe this could be cleanly expressed by a state machine, for example.

src/transforms/click_type.h
```diff
@@ -0,0 +1,60 @@
+#ifndef _click_types_H_
+#define _click_types_H_
+
+#include "transforms/transform.h"
+
+
+/**
+ * ClickTypes defines the types of clicks a ClickType transform can detect
+ */
+enum class ClickTypes { SingleClick, LongSingleClick, UltraLongSingleClick, DoubleClick };
```
mairas, line 10: As mentioned before, I think ButtonPress and ButtonRelease are also semantically UI events that could (and should) be reported because UI actions could also be based on them.

mairas (changes requested):

src/transforms/click_type.h
joelkoz: For clarification - you are interested in ButtonPress and ButtonRelease only, and not those differentiated for both single AND double clicks, right?  You expect to see:

```
ButtonPress -> ButtonRelease -> <short pause> ButtonPress -> ButtonRelease-> DoubleClick
```

and not

```
ButtonPress -> ButtonRelease -> <short pause> DoubleButtonPress-> DoubleButtonRelease-> DoubleClick
```

How about a compromise, and I add a boolean to the constructor that specifies if those should be sent?  Certainly, for my current use, having to deal with those actually complicates the implementation of SmartSwitchController, as now it will have to watch for those and throw them out.  SmartSwitchController is only interested in UltraLongButtonClick, and everything else. A single click and a double click both represent a simple single action (toggle the light), not five.

ba58smith: So you want a single click and a double click to do the exact same thing in the SmartSwitchController: toggle the switch? Why not a triple click, too? :-)
This seems strange to me, and unnecessary. If it's documented that a single click will toggle the status, that's what people will expect.

src/sensors/button.h
ba58smith: And I've been meaning to ask: what's the purpose of repeating the output every so often? What's the use case?

joelkoz: `PressRepeater`?

That only makes sense if the digital input functionality is removed entirely.  Perhaps `Button` is a candidate for one of those high level "composite classes" that wires itself to make it easier for the end user to use

src/transforms/click_type.h
joelkoz: Its not that I WANT a double click to do it. It's that I don't care.  I think it would be confusing to an end end user (that is, the user of a device created by one of our users) that if they double click on the light switch, it would do NOTHING.  I know that would confuse my wife!  We have automated dimmers all over the house that response to clicks and double clicks.

If its documented that a single click will toggle, and they never try a double click, then no harm no foul.  What behavior would you expect?  A quick on/off?

One could always go into the configuration and make the "DoubleClickInterval" so long that the ClickType transform would never register a double click.

src/sensors/button.h
joelkoz: @ba58smith Think of the "volume up" button on a remote control.  To raise or lower the volume, users usually just presses and holds the button.  If the receiver gets an "raise volume up one unit", repeated at normal intervals, you get the behavior one would expect.

ba58smith: Based on a conversation with @mairas , I've renamed DigitalInputValue to DigitalInputState. Is this a DigitalInputStateRepeater? (The more Sensors and Transforms we have, the harder it's going to be to name them.)

src/transforms/click_type.h
ba58smith: I'm the one who's always arguing for making things as simple to understand as possible for end-users. So I retract my comment: I'm perfectly OK for a click and a double-click to do the same thing.

ba58smith (approved):
My comments have been addressed. Merge approved, once a better name is decided upon, and @mairas 's comments are addressed.

src/transforms/click_type.h
mairas: Yes, I would like to see events for both button press and release events (the transitions) and the semantic click, double click and long click events. It is then up to the observer to ignore the irrelevant ones and act accordingly.

src/sensors/button.h
joelkoz: @ba58smith That sounds like a good candidate.  But to be clear - this class does not just repeat the state blindly.  It repeats the state ONLY IF it is "pressed" (be that HIGH or LOW).  If the digital input is in the opposite state, no repeating is done.  That is what makes this class different than say, the RepeatReport transform.

joelkoz: @mairas Does `DigitalInputStateRepeater` work for you?

ba58smith: OK, how about RepeaterOfTheDigitalPinStateIfTheButtonIsPressed? :-)

mairas: > @mairas Does `DigitalInputStateRepeater` work for you?

Sure but I don't quite see what benefit you'd get over just using the RepeatReport transform from the other PR. Is it that you don't want to repeat the unpressed state? I'd rather have as few and as generic classes implementing digital input reading as possible and then using transforms to enriching the input in meaningful ways.

joelkoz: @mairas That is EXACTLY the difference between the two.  The (currently named) Button only repeats is PRESSED state, and then only after it has been pressed for at least `repeat_start_interval`.  It is an entirely different thing from RepeatReport. It is why I think Button is actually appropriate, because this specialized behavior is unique (as far as I know) to buttons.  I don't mind making an intermediate transform that extracts out that behavior for use by other boolean producers, but I do think that a physical button, by far, is the most common use case. That is why if we do extract out this functionality, I still think we should have a high level `Button` class that combines the digital input and this repeater functionality.  Much easy for the hobbyist to use.

joelkoz:
Closing this PR and re-opening due to rebasing issues

PR #279, opened 2020-11-27
Refactored DigitalOutput to pass through input values

joelkoz:
DigitalOutput was originally designed as a ValueConsumer<bool>.  This minor refactor changes it to a BooleanTransform that passes the input value along after setting the pin, allowing for its use in more complicated wiring designs.

### DIFF
--- a/src/sensors/digital_output.cpp
+++ b/src/sensors/digital_output.cpp
@@ -7,6 +7,7 @@ DigitalOutput::DigitalOutput(int pin) {
   pinMode(pin, OUTPUT);
 }
 
-void DigitalOutput::set_input(bool newValue, uint8_t inputChannel) {
-  digitalWrite(pin_number, newValue);
+void DigitalOutput::set_input(bool new_value, uint8_t inputChannel) {
+  digitalWrite(pin_number, new_value);
+  this->emit(new_value);
 }
--- a/src/sensors/digital_output.h
+++ b/src/sensors/digital_output.h
@@ -4,9 +4,18 @@
 #include <ArduinoJson.h>
 
 #include "system/observable.h"
-#include "system/valueconsumer.h"
+#include "transforms/transform.h"
 
-class DigitalOutput : public BooleanConsumer {
+/**
+ * DigitalOutput configures the specified GPIO pin
+ * as an output pin, then sets the status of that
+ * pin to HIGH whenever set_input() is called
+ * and new_value is true. The pin is set to LOW when
+ * new_value is false. After the output pin's state
+ * is set, the new_value is passed through unmodified
+ * as the value returned by BooleanTransform::get().
+ */
+class DigitalOutput : public BooleanTransform {
  public:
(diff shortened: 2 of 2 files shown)


ba58smith:
I've had problems with Transforms that output booleans to Signal K, because there's no really good way to represent a boolean in Instrument Panel. (Months ago, I had to refactor DigitalInputValue to output an int instead of a boolean to get it to display properly in Instrument Panel.) How about making this either a SymmetricTransform<T> with typedefs for bool and Int, or Transform<C, P> that can take in either an int or a bool, and output an int or a bool?

mairas:
Normally you'd prefer to have the DigitalOutput and any other observer to be connected to the same source observable. Is the point here to guarantee that the GPIO pin state is set before the output value is emitted? If so, makes sense and feel free to merge (if Butch agrees). As for Butch's comment, I'd keep the type cast separate from this one because the observability of DigitalOutput is almost a side effect and having the type template might be confusing wrt. the main use case (setting output pins).

joelkoz:
Really, the point of this particular PR is to eliminate the need for my `LoadController` class.  The SmartSwitch implementation had its output pass from the SmartSwitchController to the relay to the SKOutput() to report its state.  I thinks the original implementation of this as a ValueConsumer<bool> assumed the pin would always be the end point, but my use case revealed that there are uses where it is just one of many along the way.  Yes, it is true that SmartSwitchController could be wired to the relay, and then separately be wired to the SKOutput(), but that approach kind of defeats the purpose of the chained `conntect_to()` feature.

@ba58smith - I would make it a `SymmetricTransform`, but that would break the original API signature.  The current/original implementation of this is its a BooleanConsumer.  I guess I could make it a template that is a `Transform<bool, T>`, where you can specify the PRODUCER side, but I think we have to keep the consumer side as a Bool least we break code in the wild.

joelkoz:
> I've had problems with Transforms that output booleans to Signal K, because there's no really good way to represent a boolean in Instrument Panel.

@ba58smith - Now that we have the TruthToText() transform, you can send any bool value through TruthToText() before SKOutput(), and that will send "ON"/"OFF" (or any other values for true/false) to the SK server.

I will say that I want to continue to hack away at the typecast problem.  I have this feeling that there are still some clever uses of template specifiers we can use in the ValueProducer template that will allow `connect_to()` to be used for asymmetric transforms like <bool,int> and <int,bool> and <int,float> etc. Given that, if no one objects, I'd rather say that "digital outputs are true/false, so they should producer true/false", and any limitations we have for using bools should be solved rather than accept the limitations and have that influence our API choice.  I don't want to discount the problem, I just think it is a problem separate from this one.

ba58smith:
I'm cool with that explanation, @joelkoz - merge away! (But to be clear, the problem is sending TRUE or FALSE to Instrument Panel. It works better to send 1 or 0.)

mairas:
Agree with Joel's sentiments on digital output typing. Merging.

PR #280, opened 2020-11-27
TextToTruth transform

joelkoz:
This is the remnants of the "LoadController and Truth" PR, which now contains only the TextToTruth() and TruthToText() transforms.  LoadController is no longer needed if DigitalOutput is refactored to be a BooleanTransform

### DIFF
--- a/src/transforms/truth_text.h
+++ b/src/transforms/truth_text.h
@@ -29,7 +29,7 @@ extern bool is_valid_false(String value);
  *   <li>On
  *   <li>Any string that represents a non-zero number
  * </ul>
- * 
+ * @see TruthToText
  */
 class TextToTruth : public Transform<String, bool> {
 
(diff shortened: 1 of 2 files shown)


src/transforms/truth_text.h
```diff
@@ -0,0 +1,60 @@
...
+   /**
+    * Returns TRUE if value represents one of truth values recognized by TextToTruth
+    */
+    static bool is_valid_true(String value);
+
+   /**
+    * Returns TRUE if value represents one of boolean values recognized 
+    * as a "false" value.  FALSE is returned if value value that is
+    * normally associated with a boolean false.
+    */
```
ba58smith, line 35: Typo: should be "if value is a value...".

mairas:
A-ok! Merging.

PR #281, opened 2020-11-27
ClickType Transform

joelkoz:
This is a refactor of the ClickType class. 

Changes based on our discussions:

1. ButtonPress and ButtonRelease ClickTypes have been added and are reported appropriately

2. The big convoluted IF statement from `set_input()` has been broken up into smaller functional methods for clarity

3. All protected state member variables have been documented for Doxygen.  Many have been renamed to clarify their use

4. elapsedMillis class library was used for timing routines.  Two member values that saved the current millis() value has been refactored to be elapsedMillis types

I did not use a state machine because really there are only two states: "idle" (`click_count == 0`) and "click in progress" (`click_count >= 1`).  The old code did use some implied values on the timer values to determine state, so I refactored the code to use these more explicit state values in the IF statements.

The Button class from the other PR has been dropped and I'll put similar but refactored functionality in a new PR

### DIFF
--- a/src/transforms/click_type.cpp
+++ b/src/transforms/click_type.cpp
@@ -1,148 +1,148 @@
 #include "transforms/click_type.h"
+
 #include "ReactESP.h"
 
-ClickType::ClickType(String config_path, long long_click_delay, long double_click_interval, long ultra_long_click_delay) :
-   Transform<bool, ClickTypes>(config_path),
-    click_count{0},
-    long_click_delay{long_click_delay},
-    ultra_long_click_delay{ultra_long_click_delay},
-    double_click_interval{double_click_interval},
-    delayed_click_report{NULL} {
-   load_configuration();
+ClickType::ClickType(String config_path, long long_click_delay,
+                     long double_click_interval, long ultra_long_click_delay)
+    : Transform<bool, ClickTypes>(config_path),
+      click_count{0},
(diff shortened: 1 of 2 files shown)


ba58smith:
Looks OK to me. I didn't do a deep dive into the logic - I assume you have tested it all thoroughly.

ba58smith (approved):
OK to merge (assuming you've thoroughly tested the functionality).

mairas:
Needs reformatting. The implementation is now readable enough, although I think if the implementation would've been formulated as a state machine _around the internal state_, not the input value, I believe it would've been quite a bit cleaner. And with internal state, I'm referring to states such as start, single_click, single_click_released, double_click, end.

But anyway, good enough for now once reformatted!

joelkoz:
I did actually consider that, but what I found was that the state goes from `button down`, to `button up`, and then at that point, after examining the elapsed duration of the press, the new "state" simply emits the discovered value, and then immediately goes back to the "waiting" state.  The only way I know to write simple state machines absent some state machine library is to use a loop and a switch statement for the different states, and I thought that would actually add complexity, not remove it.

I will check it in as is now (after reformatting), but I don't mind a future refactor.  Perhaps if you have some time, you could give it a go with how you thought it should be done.  I like learning new techniques! :)

joelkoz:
Merge initial approved version of ClickType class

PR #282, opened 2020-11-28
PressRepeater transform

joelkoz:
This is the main functionality of the old Button class, refactored to drop the digital input functionality and wrap the behavior in a BooleanTransform.

I opted to name this PressRepeater because upon further reflection, the suggested DigitalInputRepeater was too specific. While wiring this up to a digital input is by far the most common use case, I could see the need to use a different input, such as the boolean value coming in from a PUT request listener. That name may have made sense if we kept the "digital input" functionality in it, but as of now, this is a simple, generic transform.

### DIFF
--- a/src/transforms/press_repeater.cpp
+++ b/src/transforms/press_repeater.cpp
@@ -1,0 +1,86 @@
+#include "transforms/press_repeater.h"
+
+
+PressRepeater::PressRepeater(String config_path, int integer_false, int repeat_start_interval, int repeat_interval) : 
+    BooleanTransform(config_path),
+    integer_false{integer_false},
+    repeat_start_interval{repeat_start_interval},
+    repeat_interval{repeat_interval},
+    pushed{false},
+    repeating{false} {
+    load_configuration();
+}
+
+
+void PressRepeater::enable() {
+
+  app.onRepeat(10, [this]() {
+
+     if (pushed) {
+         // A press is currently in progress
+         if (repeating) {
+            if (last_value_sent > (unsigned long)repeat_interval) {
+                debugD("Repeating press report");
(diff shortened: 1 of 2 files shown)


src/transforms/press_repeater.cpp
```diff
@@ -0,0 +1,86 @@
...
+            }
+         }
+         else if (last_value_sent > (unsigned long)repeat_start_interval) {
+            debugD("Starting press report repeat");
+            repeating = true;
+            last_value_sent = 0;
+            this->emit(true);
+         }
+     }
+  });
```
ba58smith, line 35: Why do you need to cast `repeat_interval` and `repeat_start_interval` as `unsigned long`? Can you not simply declare / define them that way in the class definition? (This is not a criticism, but rather a simple question because I don't know.)

ba58smith (approved):
Looks good to me.

src/transforms/press_repeater.cpp
joelkoz: @ba58smith unsigned long takes up 4 bytes.  int takes up 2 is some 32 bit systems.  The intervals to be specified is on order of seconds, so there is no need to waste the extra storage.  Thus, the declaration.  On the other hand, when doing the comparison, a warning is generated about a 32 vs 16 bit comparison, and in many configurations, it is considered an error that fails to compile.  Thus, the type cast. Its the programmer's way of saying "I acknowledge the discrepancy, but in this case, it is safe).

Upon further review, the better declaration in the class should be int16_t, and not int, as that guarantees 16 bits in all environments.

mairas:
Looks good. Needs reformatting, though.

PR #283, opened 2020-11-30
Fix EspSoftwareSerial problem

ba58smith:
Use `lib_ignore` for ESP8266 in `platformio.ini` to finally resolve the EspSoftwareSerial problem.

### DIFF
--- a/library.json
+++ b/library.json
@@ -111,8 +111,7 @@
       "name": "Adafruit AHRS"
     },
     {
-      "name": "EspSoftwareSerial",
-      "platforms": ["espressif32"]
+      "name": "EspSoftwareSerial"
     },
     {
       "name": "RemoteDebug",
--- a/examples/platformio.ini
+++ b/examples/platformio.ini
@@ -35,6 +35,7 @@ board_build.ldscript = eagle.flash.4m1m.ld
 build_flags =
    -Wall
    -Wno-reorder
+lib_ignore = EspSoftwareSerial   
 
 [env:d1_mini]
 extends = espressif8266_base


library.json
```diff
@@ -111,8 +111,7 @@
       "name": "Adafruit AHRS"
     },
     {
-      "name": "EspSoftwareSerial",
-      "platforms": ["espressif32"]
+      "name": "EspSoftwareSerial"
     },
```
joelkoz, line 115: I tested this yet again on my machine, and I think its important to know that if you delete this reference to "EspSoftwareSerial" ENTIRELY, I'm pretty sure you can still build everything with what is currently in master.  I did this on my machine:

1. Clean copy of current SensESP/master
2. Delete .pio
3. Removed entire "EspSoftwareSerial" reference from library.json

I was able to build d1_mini and esp32dev build targets for SensESP library.

I then went to a project I have for an ESP32:

1. Deleted .pio entirely
2. Built for Heltec ESP32 (my own build environment in my project's platformio.ini)
3. build the esp32dev version of my project

Everything built fine, both for the library proper as well as for a project that uses the current version of the library.  I am not convinced we need this entry at all in library.json.

Can someone else try to repeat on their machine?

mairas: It does work for me as expected. If all references to EspSoftwareSerial are removed, it doesn't get pulled in (but the gps example fails to build). Double-check your project specific platformio.ini that it's using SensESP on your local filesystem and then the relevant library.json entries on that SensESP clone.

mairas:
I'd prefer not adding extra complexity to project platformio.ini files. That's not the level where internal project dependencies are handled.

I've posted a question on the PlatformIO forum about the library.json file syntax: https://community.platformio.org/t/platform-specific-dependencies-in-library-json/17483

I dug in to the PlatformIO Core code and found this smelly commit that would seem to be very much related: https://github.com/platformio/platformio-core/commit/39cb23813f713fd151a7018e81069e0d7f959a8d

If we need to resolve this locally, I think we can either:

- revert Bjarne Hansen's commit in which he changed the Software Serial library SensESP is depending on.

or

- refactor the gps example code to use the second HW serial interface on ESP32

library.json
joelkoz: > (but the gps example fails to build)

ah ha!  So, its the "gps example" that causes the error.  I asume you are talking about `examples/gps_compass.cpp`?

So, this brings up two questions:

1. I checked and `gps_compass.cpp` IS listed as one of the files to build in the automated test script, so why doesn't it fail when we check in these modifications to `library.json` and/or `platformio.ini`?  I am guessing maybe the automated testing only builds for one of the two ESP platforms?  Or, maybe it compiles but doesn't link?  Is this something we should take a look at?  Should we open an issue for it?

2. If the real problem is the `gps_compass.cpp` only, shouldn't the examples be considered "project specific", and thus if one were to build the example, shouldn't the EspSoftwareSerial be put in the `platformio.ini` of that "implementation's project?" and not `library.json`.  OR - is it that `sensors/gps.cpp` (which IS part of SensESP) requires EspSoftwareSerial on an ESP32?  If it is `sensors/gps.cpp` that is the requirement, that is part of the SensESP library, and thus it needs to be resolved in the `library.json`.  If it the `gps_compass.cpp`, then that is an implementation usage, and not part of the library proper IMO.

joelkoz:
> * revert Bjarne Hansen's commit in which he changed the Software Serial library SensESP is depending on

Are we talking about #145 ?

library.json
mairas: 1. Yes, the CI tests only build against the esp8266. It would be trivial to add esp32 there as well, at the cost of roughly doubling the build time. Now that we're getting errors in only one of the platforms, it might make sense. We could maybe limit the number of examples built (because they're not ideal as test cases, anyway) but then add esp32 to the mix.

2. Yes, now that you put it that way, you're absolutely correct! That library is _not_ a dependency for SensESP as is but for a specific example file. `sensors/gps.cpp` should happily use any stream and is not depending on any software serial implementation. Considering this, EspSoftwareSerial should indeed be dropped from `library.json`. And considering point no. 1 above, it won't even cause the tests to fail...

I apologize for my stubbornness and lack of thinking! 😁

FWIW, although I'm quite happy with that GPS code, I think it has no business being a part of SensESP. It's not an essential piece of the architecture and should rather be a "leaf" project of its own that could be pulled in as a dependency when needed. I'm hoping to cull that (and a lot of other code, too, for that matter) once we've done the 1.0 release.

mairas:
> Are we talking about #145 ?

Yes. Most of that PR was temporarily reverted after some time because of all the dependency issues it was triggering but seems like EspSoftwareSerial stayed.

ba58smith:
OK, so...
1. Don't merge this PR.
2. Make a PR that deletes the gps_compass example from the automated test script.
3. Make a PR that eliminates EspSoftwareSerial from library.json.
4. Make a PR that deletes a few more examples from the automated test script, and that includes a few examples to be built for ESP32.
That about it?

PR #284, opened 2020-11-30
Change debounce logic

ba58smith:
The logic for the Debounce Transform wasn't really doing a debounce. Now it is.
Also made it a Template class to handle all types of input.
Also made the debounce delay configurable.
Also updated the comments to Doxygen style.
(Oops - I guess this should have been four different commits. Sorry!)

### DIFF
--- a/src/transforms/debounce.cpp
+++ b/src/transforms/debounce.cpp
@@ -4,18 +4,32 @@
 // See the limitations and suggestion solution in SensESP Issue #287.
 // https://github.com/SignalK/SensESP/issues/287
 
-template<class T>
+template <class T>
 DebounceTemplate<T>::DebounceTemplate(int ms_min_delay, String config_path)
-    : SymmetricTransform<T>(config_path), ms_min_delay_{ms_min_delay} {
-      interrupt_timer_ = 0;
+    : SymmetricTransform<T>(config_path),
+      ms_min_delay_{ms_min_delay},
+      value_sent_{false},
+      stable_input_{false},
+      reaction_{NULL} {
+  load_configuration();
 }
 
-template<class T>
+template <class T>
 void DebounceTemplate<T>::set_input(T newValue, uint8_t inputChannel) {
-  if (interrupt_timer_ > ms_min_delay_) {
-    this->emit(newValue);
+  if (reaction_ != NULL) {
+    reaction_->remove();
+    reaction = NULL;
   }
-  interrupt_timer_ = 0;
+  // Input has changed since the last emit, or this is the first
+  // input since the program started to run.
+  if (input != stable_input_ || value_sent_ == false) {
+    reaction_ = app.onDelay(ms_min_delay_, [this, input]() {
+      this->reaction_ = NULL;
+      this->stable_input_ = input;
+      this->value_sent_ = true;
+      this->emit(input);
+    });
+  } 
 }
 
 template<class T>
--- a/src/transforms/debounce.cpp
+++ b/src/transforms/debounce.cpp
@@ -13,9 +13,9 @@ DebounceTemplate<T>::DebounceTemplate(int ms_min_delay, String config_path)
 template<class T>
 void DebounceTemplate<T>::set_input(T newValue, uint8_t inputChannel) {
   if (interrupt_timer_ > ms_min_delay_) {
-    interrupt_timer_ = 0;
     this->emit(newValue);
   }
+  interrupt_timer_ = 0;
 }
 
 template<class T>
--- a/src/transforms/debounce.cpp
+++ b/src/transforms/debounce.cpp
@@ -1,14 +1,48 @@
 #include "debounce.h"
 
-Debounce::Debounce(int ms_min_delay, String config_path)
-    : BooleanTransform(config_path), ms_min_delay{ms_min_delay} {
-  last_time = millis();
+template<class T>
+Debounce<T>::Debounce(int ms_min_delay, String config_path)
+    : SymmetricTransform<T>(config_path), ms_min_delay{ms_min_delay} {
+      interrupt_timer = 0;
 }
 
-void Debounce::set_input(bool newValue, uint8_t inputChannel) {
-  int elapsed = millis() - last_time;
-  if (newValue != output || elapsed > ms_min_delay) {
-    last_time = millis();
+template<class T>
+void Debounce<T>::set_input(T newValue, uint8_t inputChannel) {
+  if (interrupt_timer > ms_min_delay) {
+    interrupt_timer = 0;
     this->emit(newValue);
   }
 }
+
+template<class T>
+void Debounce<T>::get_configuration(JsonObject& root) {
+  root["min_delay"] = ms_min_delay;
+}
+
+static const char SCHEMA[] PROGMEM = R"###({
+    "type": "object",
+    "properties": {
+        "min_delay": { "title": "Minimum delay", "type": "number", "description": "The minimum time in ms between inputs for output to happen" }
+    }
+  })###";
+
+template<class T>
+String Debounce<T>::get_config_schema() { return FPSTR(SCHEMA); }
+
+template<class T>
+bool Debounce<T>::set_configuration(const JsonObject& config) {
+  String expected[] = {"min_delay"};
+  for (auto str : expected) {
+    if (!config.containsKey(str)) {
+      return false;
+    }
+  }
+  ms_min_delay = config["min_delay"];
+  return true;
+}
+
+// define all possible instances of a Debounce
+template class Debounce<bool>;
+template class Debounce<int>;
+template class Debounce<float>;
+template class Debounce<String>;
--- a/src/transforms/debounce.cpp
+++ b/src/transforms/debounce.cpp
@@ -1,5 +1,9 @@
 #include "debounce.h"
 
+// Developers: this isn't an ideal implementation of a templated Transform.
+// See the limitations and suggestion solution in SensESP Issue #287.
+// https://github.com/SignalK/SensESP/issues/287
+
 template<class T>
 DebounceTemplate<T>::DebounceTemplate(int ms_min_delay, String config_path)
     : SymmetricTransform<T>(config_path), ms_min_delay_{ms_min_delay} {
--- a/src/transforms/debounce.cpp
+++ b/src/transforms/debounce.cpp
@@ -1,13 +1,13 @@
 #include "debounce.h"
 
 template<class T>
-Debounce<T>::Debounce(int ms_min_delay, String config_path)
+DebounceTemplate<T>::DebounceTemplate(int ms_min_delay, String config_path)
     : SymmetricTransform<T>(config_path), ms_min_delay_{ms_min_delay} {
       interrupt_timer_ = 0;
 }
 
 template<class T>
-void Debounce<T>::set_input(T newValue, uint8_t inputChannel) {
+void DebounceTemplate<T>::set_input(T newValue, uint8_t inputChannel) {
   if (interrupt_timer_ > ms_min_delay_) {
     interrupt_timer_ = 0;
     this->emit(newValue);
@@ -15,7 +15,7 @@ void Debounce<T>::set_input(T newValue, uint8_t inputChannel) {
 }
 
 template<class T>
-void Debounce<T>::get_configuration(JsonObject& root) {
+void DebounceTemplate<T>::get_configuration(JsonObject& root) {
   root["min_delay"] = ms_min_delay_;
 }
 
@@ -27,10 +27,10 @@ static const char SCHEMA[] PROGMEM = R"###({
   })###";
 
 template<class T>
-String Debounce<T>::get_config_schema() { return FPSTR(SCHEMA); }
+String DebounceTemplate<T>::get_config_schema() { return FPSTR(SCHEMA); }
 
 template<class T>
-bool Debounce<T>::set_configuration(const JsonObject& config) {
+bool DebounceTemplate<T>::set_configuration(const JsonObject& config) {
   String expected[] = {"min_delay"};
   for (auto str : expected) {
     if (!config.containsKey(str)) {
@@ -42,7 +42,7 @@ bool Debounce<T>::set_configuration(const JsonObject& config) {
 }
 
 // define all possible instances of a Debounce
-template class Debounce<bool>;
-template class Debounce<int>;
-template class Debounce<float>;
-template class Debounce<String>;
+template class DebounceTemplate<bool>;
+template class DebounceTemplate<int>;
+template class DebounceTemplate<float>;
+template class DebounceTemplate<String>;
--- a/src/transforms/debounce.cpp
+++ b/src/transforms/debounce.cpp
@@ -11,14 +11,14 @@ DebounceTemplate<T>::DebounceTemplate(int ms_min_delay, String config_path)
       value_sent_{false},
       stable_input_{false},
       reaction_{NULL} {
-  load_configuration();
+  this->load_configuration();
 }
 
 template <class T>
-void DebounceTemplate<T>::set_input(T newValue, uint8_t inputChannel) {
+void DebounceTemplate<T>::set_input(T input, uint8_t inputChannel) {
   if (reaction_ != NULL) {
     reaction_->remove();
-    reaction = NULL;
+    reaction_ = NULL;
   }
   // Input has changed since the last emit, or this is the first
   // input since the program started to run.
@@ -63,4 +63,4 @@ bool DebounceTemplate<T>::set_configuration(const JsonObject& config) {
 template class DebounceTemplate<bool>;
 template class DebounceTemplate<int>;
 template class DebounceTemplate<float>;
-template class DebounceTemplate<String>;
+//template class DebounceTemplate<String>;
--- a/src/transforms/debounce.cpp
+++ b/src/transforms/debounce.cpp
@@ -2,21 +2,21 @@
 
 template<class T>
 Debounce<T>::Debounce(int ms_min_delay, String config_path)
-    : SymmetricTransform<T>(config_path), ms_min_delay{ms_min_delay} {
-      interrupt_timer = 0;
+    : SymmetricTransform<T>(config_path), ms_min_delay_{ms_min_delay} {
+      interrupt_timer_ = 0;
 }
 
 template<class T>
 void Debounce<T>::set_input(T newValue, uint8_t inputChannel) {
-  if (interrupt_timer > ms_min_delay) {
-    interrupt_timer = 0;
+  if (interrupt_timer_ > ms_min_delay_) {
+    interrupt_timer_ = 0;
     this->emit(newValue);
   }
 }
 
 template<class T>
 void Debounce<T>::get_configuration(JsonObject& root) {
-  root["min_delay"] = ms_min_delay;
+  root["min_delay"] = ms_min_delay_;
 }
 
 static const char SCHEMA[] PROGMEM = R"###({
@@ -37,7 +37,7 @@ bool Debounce<T>::set_configuration(const JsonObject& config) {
       return false;
     }
   }
-  ms_min_delay = config["min_delay"];
+  ms_min_delay_ = config["min_delay"];
   return true;
 }
 
--- a/src/transforms/debounce.h
+++ b/src/transforms/debounce.h
@@ -23,8 +23,8 @@ class Debounce : public SymmetricTransform<T> {
   virtual void set_input(T new_value, uint8_t input_channel = 0) override;
 
  private:
-  elapsedMillis interrupt_timer;
-  int ms_min_delay;
+  elapsedMillis interrupt_timer_;
+  int ms_min_delay_;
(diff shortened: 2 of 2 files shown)


src/transforms/debounce.cpp
```diff
@@ -1,14 +1,48 @@
 #include "debounce.h"
 
-Debounce::Debounce(int ms_min_delay, String config_path)
-    : BooleanTransform(config_path), ms_min_delay{ms_min_delay} {
-  last_time = millis();
+template<class T>
+Debounce<T>::Debounce(int ms_min_delay, String config_path)
+    : SymmetricTransform<T>(config_path), ms_min_delay_{ms_min_delay} {
```
joelkoz, line 5: So if this is changed from `Debounce` to `Debounce<T>`, it will break existing code. And alternative would be to name the templated class something unique (like `DebounceTemplate<T>`), and then `typedef DebounceTemplate<bool> Debounce`.  That would prevent the API from breaking. You could then define a second `typedef` like `typedef DebounceTemplate<int> DebounceInt`, etc.

src/transforms/debounce.h
```diff
@@ -2,21 +2,37 @@
...
+ private:
+  elapsedMillis interrupt_timer_;
+  int ms_min_delay_;
+  virtual void get_configuration(JsonObject& doc) override;
+  virtual bool set_configuration(const JsonObject& config) override;
+  virtual String get_config_schema() override;
 };
 
+typedef Debounce<bool> DebounceBool;
+typedef Debounce<int> DebounceInt;
```
joelkoz, line 34: I see now you have already done the `typedef`, so even more so, you are in perfect position to maintain backward compatibility by having the lone `Debounce` be a typedef for the previous boolean version.

src/transforms/debounce.cpp
```diff
@@ -1,14 +1,48 @@
...
+  }
+  ms_min_delay_ = config["min_delay"];
+  return true;
+}
+
+// define all possible instances of a Debounce
+template class Debounce<bool>;
+template class Debounce<int>;
+template class Debounce<float>;
+template class Debounce<String>;
```
joelkoz, line 48: So we are all aware, by putting the implementation of everything here in the .cpp file, the compiler will not be capable of generating a Debounce<> class for any future data types (e.g. `Debounce<SomeFutureDataType>`, as it will be blind to the implementation. On the other hand, if you put the entire implementation in the header file (esp. the configuration stuff) the compiler will generate a TON of redundant code for each class.  There is a bit of a compromise technique, but it can't really be used here unless we make some tweaks to the Transform class.  I'll open an issue for it now.  I am OK for now limiting Debounce templates to the "common" data types as long as we know the limitation is there.  Before we go too much further in expanding the other Transforms to use templates, however, we need to fix and make it "right."  Maybe add a comment here (or at the top of the CPP file) to indicate that this file should not be used as a pattern for other transforms (at least, not as is).

joelkoz (changes requested):
Rename the template to `DebounceTemplate<>` (or some other form), and typedef the boolean version to be simply `Debounce` to preserve backward compatibility.  I opened issue #287 for the second issue re: having the implementation in the .cpp file having some limitations, but I think those limitations are acceptable for now.

src/transforms/debounce.h
joelkoz: ...and actually, you can leave the new DebounceBool to be consistent with the others, and make an additional typedef for Debounce to also be bool.

src/transforms/debounce.cpp
ba58smith: Done.

src/transforms/debounce.h
ba58smith: Done.

src/transforms/debounce.cpp
ba58smith: Comment added to .h and .cpp file.

ba58smith:
Before I merge this...I think I have the debounce logic wrong. Instead of this:
```
void DebounceTemplate<T>::set_input(T newValue, uint8_t inputChannel) {
  if (interrupt_timer_ > ms_min_delay_) {
      interrupt_timer_ = 0;
      this->emit(newValue);
  }
}
```
I think it should be this:
```
void DebounceTemplate<T>::set_input(T newValue, uint8_t inputChannel) {
  if (interrupt_timer_ > ms_min_delay_) {
    this->emit(newValue);
  }
  interrupt_timer_ = 0;
}
```
The latter will output the very first state change at the beginning of the button press, and then ignore all the bouncing that occurs while the button is "settling in" to its new state, which is (I think) what we want. That's not what the former implementation does.

Right? I need to change it?

joelkoz:
You bring up a good point, but I think it still may be wrong, depending on what is feeding the input.  The implementation is correct only if Debounce() is being fed changes in state.  If it is being fed a steady stream of state, the later implementation will NEVER emit a value, even if the state is "stable". DigitalInputState sends a continuous stream every X ms. That version would never emit a value if the ms_min_delay_ is greater than the input interval.

ba58smith:
After a long discussion on the phone with @joelkoz , we decided to go with the new implementation, and added details to the comments in the .h file to explain how to set ms_min_delay properly.

ba58smith:
This article gave excellent information about bouncy buttons, and how to debounce them.
https://www.allaboutcircuits.com/technical-articles/switch-bounce-how-to-deal-with-it/

ba58smith:
OK, now I think this is ready for merging, unless someone sees something wrong in the last commit.

src/transforms/debounce.cpp
```diff
@@ -1,14 +1,52 @@
...
 }
 
-void Debounce::set_input(bool newValue, uint8_t inputChannel) {
-  int elapsed = millis() - last_time;
-  if (newValue != output || elapsed > ms_min_delay) {
-    last_time = millis();
+template<class T>
+void DebounceTemplate<T>::set_input(T newValue, uint8_t inputChannel) {
+  if (interrupt_timer_ > ms_min_delay_) {
     this->emit(newValue);
```
mairas, line 16: I don't get this logic. If the input value changes, say, from 0 to 1 and doesn't bounce, wouldn't that mean that this transform never emits the new value? Instead, the input would need to be resent before emitting happens.

ba58smith: I think it's right:
1. Debounce constructor runs, interrupt_timer_ is set to 0. 
2. Sometime time later, the button is pressed, and DigitalInputChange emits HIGH. interrupt_timer_ is now the total elapsed time since the constructor ran, which is of course bigger than ms_min_delay_, so Debounce emits HIGH and interrupt_timer_ is reset to 0.
3. If there's no bounce, no problem.
4. If there's bounce, and a LOW comes through in, say, 8ms, interrupt_timer_ is only 8, which is less than ms_min_delay_ (which is 20 by default), and nothing is emitted, but interrupt_timer_ is reset to 0.
5. Now, 12ms later, a HI comes through. That's less than 20, so nothing is emitted, and interrupt_timer_ is reset to 0.
6. Steps 4 and 5 can happen many times, and then, finally, the last HI comes through when the bouncing is over. Nothing is emitted, and interrupt_timer_ is reset to 0, and we're just waiting on the button to be released.
7. A half-second later, the button is released, and the above happens, only in reverse - a LOW is emitted (because the button was held down for a half a second, which is way longer than ms_min_delay), then a bunch of bounces are NOT emitted, and it settles into the "button not pressed" state of LOW.

Right?

mairas: Hmm, that makes sense. But what happens if you have a quick transient: input is 0 at steady state, then goes to 1, and after 1 ms returns back to 0, for good. The output would remain 1 indefinitely, wouldn't it?

ba58smith: Yes, it would. But the documentation for this Transform says "DebounceTemplate implements debounce code for a button or switch" - and there's no way you're going to have a button or a switch that goes on and off in 1 ms, or in 100 ms for that matter. Physical buttons and switches operated by humans just can't change state that fast. 
And every reference on the first two page of a Google search for "microcontroller debounce" is talking about buttons, switches, and keyboard keys - all things operated by humans at much slower than the debounce threshold.
Agree?

mairas: That kind of transient would most likely be a glitch, sure, but isn't glitch removal the whole point of debouncing? Those glitches can be a result of many things: a faulty momentary switch might not make a proper contact; the power leads of a big winch motor might go next to unprotected button wiring; someone is half depressing a momentary switch and then releasing it. Likewise, you could say any debouncing isn't needed because there's no way a person could flicker a switch or a button that fast. :-)

ba58smith: Are we really trying to solve every possible hardware issue in software with button debouncing code? I don't think so, but assume we are. Are you suggesting we capture the first state change after an idle period, then check the state again after an X ms "test period", and if they're the same, then emit the value? (But even that won't work if someone pushes and releases the button too fast.) I think that would work for all but the most horrible physical buttons / circuits, but then this code won't work as input for @joelkoz 's ClickType, because it will requre too long of a test period for him to be able to properly capture a double click.

ba58smith: Having said that, I'll now try to make the argument for the current logic, based on my assertion that we should be able to expect the hardware part of the circuit to be at least reasonably well executed. If someone is using the cheapest buttons on the planet, or running small wired next to giant ones, or is just getting random inputs to the ESP from any source... that's not really our problem to solve. Debounce is specifically in the code to debounce a physical button or swtich, and as it's written it does that.

mairas: But my point was that physical buttons, in the same conditions that produce spurious state changes that require debouncing, may end in either a new state or in the original state as the final stable state. The current PR always forces a state change, which is not good, in my opinion.

In either case, only a short debouncing test period is required and it should have no impact on any intentional double clicking, because that happens on a fleshy meat bag time scale, which, as you pointed out, is much slower than anything required for debouncing.

joelkoz: > Are you suggesting we capture the first state change after an idle period, then check the state again after an X ms "test period", and if they're the same, then emit the value? 

@ba58smith Actually, this may in fact be the correct logic.  When @mairas describe the debounce as "waiting for the signal to stabilize", that helped visualize what we are trying to do here.  However this only works if the debounce is fed a constant stream of input states, and not just the "changes".  So basically:

1. Whenever the state changes, reset the timer, and reset a flag that indicates the change has been emitted.
2. If the state has NOT changed since the last update, if the timer is now greater than the "steady state waiting period" and we have NOT already emitted the value, emit it, and set a flag that indicates it has been sent.

The steady state waiting period would still be short - shorter than any human could really change a button.  BUT - this requires that this logic be fed inputs at a rate that is shorter than the steady state interval.

I wonder if it would make sense to put this logic inside the DigitalInputChange() class.  We could add a parameter to the constructor that determines if debounce logic should be used.

The other approach is to NOT have a DigitalInputChange() class at all, or if we do, it would be one of the proposed "high level/composite" classes that is made up of `DigitalInputState -> Debounce -> ChangeFilter` (which ChangeFilter being refactored to also accept boolean values and to only send out the changes.

mairas: You _can_ do debouncing with changes. Since it's a transform and not an interrupt, we can quite easily do trickery with onDelay reactions. Something like this:

```
DelayReaction* reaction = NULL;

void set_input(bool input, uint8_t input_channel = 0) {
  if (reaction != NULL) {
    reaction->remove();
  }
  if (input != stable_input) {
    reaction = app.onDelay(20, [this]() {
      reaction = NULL;
      stable_input = input;
      this->emit(input);
    });
  } else {
    reaction = app.onDelay(20, [this]() {
      reaction = NULL;
      // no need to set stable_input or emit anything
  }
}
```

joelkoz: Ah yes!  I used a similar technique with ClickTypes. You set up what you think will happen provided no further input is given with a `onDelay()` (like the next click of a double click, or in this case, a status change prior to the debounce timeout), but you pull it back if further input DOES come in.

In this case, where you have "20" would instead be the user configurable `ms_min_delay_`. @ba58smith, I vote for this implementation. It allows your `DigitalInputChange()` class to work unmodified with `Debounce()`, and it has the added benefit of being tolerant of noise on the line.  I can testify that I have made circuits where out of nowhere, a quick on/off spike happens due to some type of transient noise.  If the "stable state" is _off_, and a single "on/off" comes thru with the _off_ happening 1 ms after the _on_, that is really noise and not a true button press. In this case, the behavior should be as if the signal never happened.  In @mairas implementation, the second _off_ would be what is queued up and emitted, which would simply match the "stable state" of _off_, where as the alternate implementation would emit the initial _on_ and I think would stay that way as the first _off_ would get thrown away because is came too soon, and DigitalInputChange() would not emit a second _off_ because it has not changed.  This actually keeps `Debounce` and `DigitalInputChange` in sync: `Debounce` will throw away previous values, not subsequent values, so they both are emitting the last value in the stream to have been output.

joelkoz: I just tested this with DigitalInputChange on my machine (I happen to be updating my generator gauge), and it works for the most part.  Minor changes to get it to compile.  Here is final impl:

```
void Debounce::set_input(bool input, uint8_t inputChannel) {
if (reaction != NULL) {
    reaction->remove();
  }
  if (input != stable_input) {
    reaction = app.onDelay(ms_min_delay_, [this, input]() {
      this->reaction = NULL;
      this->stable_input = input;
      this->emit(input);
    });
  } else {
    reaction = app.onDelay(ms_min_delay_, [this]() {
      this->reaction = NULL;
      // no need to set stable_input or emit anything
    });
  }
}

```

ba58smith: The constructor needs to `load_configuration();`. I will do that.

ba58smith: DigitalInputChange outputs no more often than once every `read_delay` ms (default is 50). Am I correct that it doesn't make sense for `Debounce::ms_min_delay` to be less than `DigitalInputChange::ms_min_delay`? If Debounce gets new input from DigitalInputChange, and waits only 20 ms for another change before emitting, it will always emit, correct? If that's correct, should I make DigitalInputChange::read_delay default to something less than the default for Debounce::ms_min_delay?

ba58smith: I think that this code assumes the button / switch starts with a value of LOW, and is looking for the first legitimate HIGH to do anything, correct? Should I add a constructor parameter for `button_resting_vaiue_` or something like that to indicate the initial state of `stable_input`? It can default to `false`, but many buttons and switches rest a HIGH, and go LOW when activated.
(I think I need to do the same to DigitalInputChange, for the same reason.)

mairas: > I think that this code assumes the button / switch starts with a value of LOW, and is looking for the first legitimate HIGH to do anything, correct? Should I add a constructor parameter for `button_resting_vaiue_` or something like that to indicate the initial state of `stable_input`? It can default to `false`, but many buttons and switches rest a HIGH, and go LOW when activated.

Yes, an initial value should be provided. Or, hm. You could rather just have a `value_sent` flag initially set to false. If the flag is false, you would debounce and send the value regardless of the end state, and after that follow the original logic. The better the less there are assumptions and initial values to be set.

ba58smith: I like that approach, @mairas - I will do it for Debounce and also for DigitalInputChange (if it works there).

Any comments on my second comment above, about the relationship between `DigitalInputChange::read_delay_` and `Debounce::ms_min_delay_`?

joelkoz: Yes, I think the Debounce delay needs to be longer than the read delay.  But now that I think about it, I am questioning the wisdom of DigitalInputChange without any Debounce code before it.  Since DigitalInputChange is REALLY emitting only at fixed intervals (because its an onRepeat(), correct?), it could, in theory, catch one of those transient signals and pass it along as a state change.  Its almost like the "has this changed since the last value" should be its own transform, that would come AFTER Debounce did its thing with the signal (i.e. DigitalInputState()->Debounce()->ChangeFilter().

At a very minimum, maybe the DigitalInputChange read rate should be "20", and the debounce delay be "50".

ba58smith:
It didn't like this template class to be used with <String>, because there's no `!= String`, or so it said. I'm going to call this PR finished, and will look later (after version 1.0 is released) to see how I can do this. Maybe a different `::set_input()` for Strings.

Please give my last two commits a look - I think it's ready to merge.

src/transforms/debounce.h
```diff
@@ -3,20 +3,51 @@
...
+  virtual bool set_configuration(const JsonObject& config) override;
+  virtual String get_config_schema() override;
 };
 
-#endif
\ No newline at end of file
+typedef DebounceTemplate<bool> DebounceBool;
+typedef DebounceTemplate<bool> Debounce; // for backward-compatibility with original class
+typedef DebounceTemplate<int> DebounceInt;
+typedef DebounceTemplate<float> DebounceFloat;
```
mairas, line 50: I'm not sure debouncing would work for continuous-value signals. For a discrete-valued signal represented as a float, sure, but ... dunno. Anyway, no real harm for defining this either, so feel free to disregard me.

mairas (approved):

src/transforms/debounce.h
ba58smith: I don't have a use case for debouncing a float - I just got carried away with the typedef's. I'll leave it for now, but when I revisit DebounceTemplate<String>, I'll also look more deeply at <float>.

src/transforms/debounce.cpp
ba58smith: I know we've beaten this to death, but is it possible that we're really overthinking it? I really think it works as-is (with the most recent commits), when used with Debounce as-is (with its most recent commits). How about I merge this, and then test the heck out of it with some different buttons and switches? Since it's a new class, no one is using it yet anyway, and I need it for my LambdaConsumer example.

src/transforms/debounce.h
```diff
@@ -50,6 +48,6 @@ typedef DebounceTemplate<bool> DebounceBool;
 typedef DebounceTemplate<bool> Debounce; // for backward-compatibility with original class
 typedef DebounceTemplate<int> DebounceInt;
 typedef DebounceTemplate<float> DebounceFloat;
-typedef DebounceTemplate<String> DebounceString;
+//typedef DebounceTemplate<String> DebounceString;
```
joelkoz, line 51: I don't think there is an actual definition for what a debounce of a String would be. You can probably remove this. Even if you were thinking about allowing this to accept String versions of numbers, it would be better to send those thru a Typecast.

joelkoz (approved):
Looks good to me.  One comment about removing the commented out String version.

src/transforms/debounce.h
ba58smith: What I was thinking was that there might be some weird sensor that would ouput "HIGH" or "LOW", or "ON" / "OFF", but you're right, that's going to be unlikely, and can be fixed w/ Typecast. I'll remove it.

ba58smith: I have my git in a weird state right now. So since this is only "comment cleanup", I'm going to merge this PR as-is, and then will do a new one for the comment clean-up.

joelkoz: Ok.  Your response actually brings up a question for us to ponder: should we have a template type at all for classes like this going forward?

Debounce really only makes sense in the context of a button, which only has two boolean states.  I understand that there was a need to have an `int` version of this, because there were problems with `DigitalInputState` being defined as an `int`. Of course, I'm pretty sure that was at one time it was defined as a `bool`, but changed because of other wresting that needed to be done.  But now that we have a functional `Typecast` class, would it be better to have the classes be consistent with their design?  That is, `DigitalInputState` and `DigitalInputChanged` only produce a `<bool>`.  `Debounce` only consumes and produces a `bool`.  If, at the last minute, you need to output the number as an `int`, you send it thru a `Typecast`.  

In your "HIGH/LOW" example, we should add those values to `TextToTruth` (along with "HI/LO"), and that transform could then be fed into `Debounce` as a `bool`

Something to ponder for a v2.0 refactor...

ba58smith: @joelkoz , the reason I made Debounce into DebounceTemplate is that I needed to debounce DigitalInputChange, which outputs an `int`. Debounce originally took input of only type `bool`. In the new example I'm currently writing (the chain counter, which illustrates LambdaConsumer), I wanted to connect DigitalInputChange to Debounce, and it wouldn't. Hence making it a template class so I can use `DebounceTemplate<int, bool>`.
I suppose DigitalInputState and DigitalInputChange could be changed to output `bool`, but the reason I originally changed DigitalInputState (then called DigitalInputValue) to output `int` was the way SK, and more specifically Instrument Panel, handles a `bool` : badly. There's no good instrument for displaying a `bool`. 
So without making Debounce a template, you'd have to have:
```
DigitalInputChange->connect_to(Typecast(bool))
                               ->connect_to(Debounce)
                               ->connect_to(Typecast(int)
                               -?connect_to(SKOutputInt)
``
Besides, I could argue that DigitalInputState / Change should output an `int` rather than a `bool`, to be consistent with how `HIGH` and `LOW` seem to always be represented when talking about the state of a pin: 1 or 0.

PR #285, opened 2020-11-30
DigitalInput fixes and updates

ba58smith:
Update comments to Doxygen style.
Change DigitalInputValue to DigitalInputState.
Properly initialize `triggered` variable.
Remove `value` from showing in the Config UI.
Add new class: DigitalInputChange.
Fix class member variable names.

### DIFF
--- a/src/sensors/digital_input.cpp
+++ b/src/sensors/digital_input.cpp
@@ -102,7 +102,8 @@ DigitalInputChange::DigitalInputChange(uint8_t pin, int pin_mode,
     : DigitalInput(pin, pin_mode, interrupt_type, config_path),
       IntegerProducer(),
       read_delay_{read_delay},
-      triggered_{false} {
+      triggered_{false},
+      last_output_{0} {
     load_configuration();    
     }
 
@@ -112,12 +113,13 @@ void DigitalInputChange::enable() {
       output = digitalRead(pin_);
       triggered_ = true;
     });
-  // app.onTick was tried for this, but for some reason, it doesn't work.
-  // app.onRepeat() does work.
+  
+  
   app.onRepeat(read_delay_, [this](){
-      if (triggered_) {
+      if (triggered_ && output != last_output_) {
         noInterrupts();
         triggered_ = false;
+        last_output_ = output;
         interrupts();
         notify();
       }
--- a/src/sensors/digital_input.cpp
+++ b/src/sensors/digital_input.cpp
@@ -10,37 +10,36 @@ DigitalInput::DigitalInput(uint8_t pin, int pin_mode, int interrupt_type,
   pinMode(pin, pin_mode);
 }
 
-DigitalInputValue::DigitalInputValue(uint8_t pin, int pin_mode,
+DigitalInputState::DigitalInputState(uint8_t pin, int pin_mode,
                                      int interrupt_type, int read_delay,
                                      String config_path)
     : DigitalInput{pin, pin_mode, interrupt_type, config_path},
       IntegerProducer(),
-      read_delay{read_delay} {
+      read_delay{read_delay},
+      triggered{false} {
   load_configuration();      
 }
 
-void DigitalInputValue::enable() {
+void DigitalInputState::enable() {
   app.onRepeat(read_delay, [this]() {
     emit(digitalRead(pin));
   });
 }
 
-void DigitalInputValue::get_configuration(JsonObject& root) {
+void DigitalInputState::get_configuration(JsonObject& root) {
   root["read_delay"] = read_delay;
-  root["value"] = output;
 }
 
 static const char SCHEMA2[] PROGMEM = R"###({
     "type": "object",
     "properties": {
-        "read_delay": { "title": "Read delay", "type": "number", "description": "The time, in milliseconds, between each read of the input" },
-        "value": { "title": "Last value", "type" : "number", "readOnly": true }
+        "read_delay": { "title": "Read delay", "type": "number", "description": "The time, in milliseconds, between each read of the input" }
     }
   })###";
 
-String DigitalInputValue::get_config_schema() { return FPSTR(SCHEMA2); }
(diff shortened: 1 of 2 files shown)


src/sensors/digital_input.cpp
```diff
@@ -94,6 +91,60 @@ bool DigitalInputCounter::set_configuration(const JsonObject& config) {
...
+void DigitalInputChange::enable() {
+  app.onInterrupt(pin_, interrupt_type_,
+    [this](){
+      output = digitalRead(pin_);
+      triggered_ = true;
+    });
+  // app.onTick was tried for this, but for some reason, it doesn't work.
+  // app.onRepeat() does work.
+  app.onRepeat(read_delay_, [this](){
+      if (triggered_) {
```
joelkoz, line 118: I had an idea about another possible implementation: What if the entire `onRepeat` block was changed to `onDelay(1`, and then put INSIDE the interrupt callback.  This would eliminate the need for the continuous checking of triggered in a repeat loop.  The downside of course is as currently implemented, if there are 1000s of interrupts coming in, the onDelay() version might fall over, while this current version would just ignore the changes.

It is just a thought. The current implementation works and has been tested, so I am good with it - the simple check of the triggered flag does not "cost" too much computation wise.

joelkoz (approved):
This looks good to me.  The one comment I made was more for experimentation, and not a requested change.

src/sensors/digital_input.cpp
```diff
@@ -56,32 +55,30 @@ DigitalInputCounter::DigitalInputCounter(uint8_t pin, int pin_mode,
                                          String config_path)
     : DigitalInput{pin, pin_mode, interrupt_type, config_path},
       IntegerProducer(),
-      read_delay{read_delay} {
+      read_delay_{read_delay} {
```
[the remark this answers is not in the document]
BjarneBitscrambler: The underscore indicates a member variable, and yes, that's from the style guide. I believe the recommended document is the [Google C++ Style Guide](https://google.github.io/styleguide/cppguide.html)

src/sensors/digital_input.cpp
ba58smith: So, replace `triggered_ = true;` with `app.onDelay(1, [this](){notify();}`? (No need for `triggered_` in this implementation, correct?)

Wouldn't that violate the concept of having the ISR do as little work as possible? I think that the execution of the `onDelay` code would happen outside of the ISR, but would setting up the `onDelay` reaction be a costly operation?

This class is specifically described as intended to be used to monitor a "button or other type of relatively slow changing switch". So if it's used for that, there wouldn't be more than a relatively small number of interrupts happening - only the number that occur during the pressing or releasing of a physical button, or the flipping of a physical switch.

I will try this. If it works, I like it, as it eliminates the need to use `noInterrupts();` and `interrupts();`, which (I think) introduces the possibility of a missed interrupt.

src/sensors/digital_input.cpp
```diff
@@ -94,3 +94,57 @@ bool DigitalInputCounter::set_configuration(const JsonObject& config) {
...
+                                       String config_path)
+    : DigitalInput(pin, pin_mode, interrupt_type, config_path),
+      IntegerProducer(),
+      read_delay_{read_delay},
+      triggered_{false} {
+    load_configuration();    
+    }
+
+void DigitalInputChange::enable() {
+  app.onInterrupt(pin, interrupt_type,
```
mairas: Now that ReactESP interrupt handling got changed/fixed, the callback should be a regular function with the ICACHE_RAM_ATTR attribute: ... except that it can't be, not without severely limiting the sensor functionality. Good as is, then. I'll create an issue about the problem.

```
static void ICACHE_RAM_ATTR DigitalInputChangeISR
```

mairas: Seems like I kicked the can forward instead of fixing the issue in ReactESP.

src/sensors/digital_input.cpp
mairas: I wouldn't do that. Adding an onDelay reaction is a whole lot more of function calls all over the place, and as Butch said, ISRs should be as small as possible.

mairas (approved):
Good as is. Kudos for nice commits!

src/sensors/digital_input.cpp
ba58smith: After discussing with @joelkoz, and input above from @mairas , I'll just leave it as-is.

ba58smith:
This class is DigitalInputCHANGE: so should it `notify` only if the new value to be output is different from the last output?

I came up with this scenario: we're monitoring a button. The button press begins, and it's bouncy. 10 ms into the button press, while it's still bouncing up and down every 3 ms, our `app.onRepeat` fires off (which it does every 50 ms). It sees that `triggered_` is `true`, so it does `notify()`. But because it happens right in the middle of all the bouncing, the last `digitalRead(pin_)` returned LOW, instead of HIGH. (HIGH indicates the button is pressed.) So now, DigitalInputChange is emitting LOW, when the button didn't really change yet.
Now, 10 more ms passes and the button stops bouncing. The last `digitalRead(pin_)` returned HIGH, and there are no more interrupts to change it. (Not until the button is released, 500 ms or more later.) The next time `app.onRepeat` fires off, `triggered_` will again be `true`, so it will call `notify()`, this time with HIGH.
If `notify()` is never called unless the current value of `output` is different from the previous value of `output`, then we truly only ever see CHANGES. 
What do you think, @mairas  and @joelkoz ?

mairas:
Yes, I think both the name and the doxygen doc imply that it should only report changes.

ba58smith:
OK, now I think it's ready to merge. If I don't see any more comments, I'll merge in about 24 hours.

ba58smith:
The failures in the CI don't make sense to me. I need to try to run them again (but don't know how to do that, without making another commit).

ba58smith:
I ran the CI checks again, and they all passed. (The same thing happened with another PR, and nothing changed.)

Merging.

PR #286, opened 2020-11-30
Create more interactive Doxygen documents

mairas:
Some improvements to our Doxygen config.

### DIFF
--- a/docs/RELEASE_PROCESS.md
+++ b/docs/RELEASE_PROCESS.md
@@ -24,12 +24,14 @@ The release process is as follows:
 
    Remember to commit the results.
 
-3. Update the doxygen docs (first bump the version number in `Doxyfile`)
+3. Update the Doxygen docs (first bump the version number in `Doxyfile`). To
+   run Doxygen, you need to have Graphviz installed on your system. It is 
+   available on your favorite package manager on any Linux distribution or Mac.
 
     rm -rf docs/generated/docs/
     doxygen
 
-   Again, commit the results.
+   Again, commit the changes.
 
 4. Push the changes to your local fork.
 
--- a/Doxyfile
+++ b/Doxyfile
@@ -350,7 +350,7 @@ AUTOLINK_SUPPORT       = YES
 # diagrams that involve STL classes more complete and accurate.
 # The default value is: NO.
 
-BUILTIN_STL_SUPPORT    = NO
+BUILTIN_STL_SUPPORT    = YES
 
 # If you use Microsoft's C++/CLI language, you should set this option to YES to
 # enable parsing support.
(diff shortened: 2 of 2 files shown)


joelkoz (commented):
Do any of these changes require additional software to be installed on the doc builder's machine to get this to work? I recall having to install some kind of graphics library so we could get nice class diagrams.  If this adds more software burden, maybe we should add to the readme file or wiki on what is needed to build the docs?  Or, is this covered already on the Doxygen site? Or better, there some automated build file for it?

mairas:
Graphviz is required for the graphs, and yes, it was an existing requirement. I did enable LaTeX support. LaTeX is readily available on Linux and Mac Homebrew and somehow on Windows as well. LaTeX is only used for rendering mathematical formula, and since I enabled Mathjax, on HTML they are rendered nicely in-page, so it might be a case of diminishing returns. I might as well disable it.

I documented the release process in `docs/RELEASE_PROCESS.md`. I'll add a mention of the requirements there. Doxygen itself doesn't benefit much from a build script because it's just a single command, and I think trying to automate the installation of the dependencies would be an exercise in futility because of the different package managers and even package names between Linux distro flavors and Mac Homebrew.

docs/RELEASE_PROCESS.md
```diff
@@ -24,12 +24,14 @@ The release process is as follows:
 
    Remember to commit the results.
 
-3. Update the doxygen docs (first bump the version number in `Doxyfile`)
+3. Update the Doxygen docs (first bump the version number in `Doxyfile`). To
+   run Doxygen, you need to have Graphviz installed on your system. It is 
+   available on your favorite package manager on any Linux distribution or Mac.
```
joelkoz, line 30: Did you remove the usage of LaTex in the Doxyfile? I don't see it there anymore.  I do see that MathJax is still there.  Does that need to be added to this section also, or does this just activate an external url in the HTML output?

I do agree that having a build script to install these would get hairy based on all the possible platforms. The mention of using whatever is most appropriate for your platform is more than adequate for my tastes.  As long as this document mentions all the _required_ software that needs to be installed to run Doxygen, I have no other feedback.

joelkoz (approved):
Looks good to me, save the one minor question I left about MathJax in the comment of RELASE_PROCESS.md

docs/RELEASE_PROCESS.md
mairas: Yup, it's gone. Mathjax is an unrelated beast; it accepts LaTeX math syntax as input (among others, I believe) but renders the output purely in browser using Javascript. And that's the beauty of it; it doesn't need any external dependencies.

mairas:
Replied to the MathJax question; no action required. Merging!

PR #289, opened 2020-12-02
PWMOutput class for platform independent pwm outputs

joelkoz:
Provides a platform independent mechanism for managing "pwm channels" on both esp8266 and esp32.  Outputs can be written via static class methods or with an instance of PWMOutput (which is a ValueConsumer<float>)


mairas:
Looks good to me!

PR #290, opened 2020-12-03
Support for color LED via RgbLed class

joelkoz:
A ValueConsumer for setting the color of an RGB led via `<bool>` or bit packed RGB color via `<long>`. This new version is implemented using PWMOutput()


ba58smith (approved):
Looks good. Nice comments!

mairas:
Looks good to me! Reformat before you merge!

PR #291, opened 2020-12-03
Expand the output_scale comment

ba58smith:
There was a lack of understanding of the purpose of the `output_scale` member of the AnalogInput class. These additional comments will hopefully resolve that.

### DIFF
--- a/src/sensors/analog_input.h
+++ b/src/sensors/analog_input.h
@@ -15,18 +15,20 @@ class AnalogInput : public NumericSensor {
    *
    * @param[in] pin The GPIO pin to read. On ESP8266, always A0. On ESP32, at
    * the moment only ADC channel 1 (pins 32..39) is supported.
-   * 
+   *
    * @param[in] read_delay Time delay between consecutive readings, in ms
-   * 
+   *
    * @param[in] config_path Configuration path for the sensor
-   * 
-   * @param[in] output_scale The maximum scale of the output. The default is 1024,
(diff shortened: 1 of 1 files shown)


src/sensors/analog_input.h
```diff
@@ -15,9 +15,18 @@ class AnalogInput : public NumericSensor {
...
    * the moment only ADC channel 1 (pins 32..39) is supported.
+   * 
    * @param[in] read_delay Time delay between consecutive readings, in ms
+   * 
    * @param[in] config_path Configuration path for the sensor
-   * @param[in] output_scale The maximum scale of the output
+   * 
+   * @param[in] output_scale The maximum scale of the output. The default is 1024,
+   * which is typical for Arduino AnalogInput pins. (Because of the way this has
+   * been implemented in SensESP, the default for ESP32 is also 1024.) If you want
```
joelkoz, line 25: I think we should be explicit about "the way this has been implemented in SensESP".  Suggestion:

`@param output_scale The scale of the converted input value AnalogInput produces. The maximum raw analog value returned by an esp8266 and an esp23 are different, but AnalogInput takes that into account by converting them to a number between zero and output_scale. This allows the same code to run on both platforms...`

then continue with your examples

joelkoz (approved):
Left a suggested modification. Do with it what you think is best...

src/sensors/analog_input.h
ba58smith: Agreed. I made the change.

mairas:
Merging.

Issue #278, opened 2020-11-27
SensESP's platformio.ini still uses a deprecated build flag


ba58smith:
```
[espressif8266_base]
;this section has config items common to all ESP8266 boards
platform = espressif8266
build_flags =
   -Wl,-Teagle.flash.4m1m.ld
```
The last line above should be removed, and the following line should be added above the `build_flags =` line: `board_build.ldscript = eagle.flash.4m1m.ld`.

ba58smith:
Closed with PR #292.

PR #228, opened 2020-11-03
Replace deprecated build flag

ba58smith:
Addresses Issue #156 .

After making this change in my project's platformio.ini, I erased the flash on my ESP8266 (D1 Mini) and reloaded. Then I brought up the Config UI and changed a couple of the configurable items in my project and saved them. Restarted the ESP and checked the Config UI to confirm my changes had been saved. Then I changed them back, restarted the ESP again, and made sure those changes had been saved. Everything worked as expected.

PR #292, opened 2020-12-03
Fix deprecated build flag for configuring flash

ba58smith:
Addresses Issue #278 .

Fixed the same way as it was fixed in the example platformio.ini file: PR #228 .

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -58,8 +58,8 @@ lib_deps =
 [espressif8266_base]
 ;this section has config items common to all ESP8266 boards
 platform = espressif8266
+board_build.ldscript = eagle.flash.4m1m.ld
 build_flags =
-   -Wl,-Teagle.flash.4m1m.ld
    -Wall
    -Wno-reorder
 extra_scripts = extra_script.py


joelkoz (commented):
I see this makes this consistent with the `examples/platformio.ini`, which is a good thing.  But, this failed to build?  Did this break the build somehow?

Also - just for my own knowledge, is this entry even necessary?  I am able to build SensESP without any kind of error.  Of course, I've never tried to deploy it, but you aren't supposed to deploy the library directly. What does this even do? I saw this come up when helping a user, but he was incorrectly using this `platformio.ini` to deploy his app.

joelkoz:
Error from the CI logs.  Not sure if the "failure" is a big deal or not:

```
Environment     Status    Duration
--------------  --------  ------------
d1_mini         SUCCESS   00:01:55.596
esp32dev        SUCCESS   00:01:56.763
esp-wrover-kit  FAILED    00:00:07.529
==================== 1 failed, 2 succeeded in 00:03:59.888 ====================
Error: Process completed with exit code 1.
```

ba58smith:
The old build flag (-Wl,-Teagle.flash.4m1m.ld) has been in the SensESP `platformio.ini` since the very beginning. I think it's necessary because the default uses 512k instead of 1 meg for whatever the second parameter is, and I'm guessing @mairas determined that 1 meg was needed?
It's always worked, and this is just the new way of doing it, so I'm not going to change it.

ba58smith:
I have no idea why it would have failed the CI check for the esp-rover-kit, as that's an ESP32. I think we can safely ignore the failure. (I had another PR fail the CI check, and the ONLY change in it was a change to a comment.)

mairas:
Right, this was the change you requested me to do for the CI ini file. I hadn't actually looked into this PR earlier (because of the red X next to it).

mairas:
Passed on a retry. Merging.

PR #293, opened 2020-12-04
Hysteresis transform can be instantiated many times

### LINKED ISSUE #250: Hysteresis should store its state in a member variable
Hysteresis transform defined in `src/transforms/hysteresis.h` stores its state in a static function variable. This probably breaks if the transform is instantiated more than once. The state should be stored in a member variable, instead.

mairas:
The previous version of `Hysteresis` used a static member variable to store the output state, potentially breaking the transform if it was instantiated more than once (as the different instances would store the same state). Now state is a regular member variable.

I had to change `LambdaTransform` to use `std::function` instead of old-fashioned function pointers. That likely incurred a small performance penalty when calling the lambda function but I think code readability is now slightly better.

Fixes #250.

### DIFF
--- a/src/transforms/hysteresis.h
+++ b/src/transforms/hysteresis.h
@@ -33,23 +33,24 @@ class Hysteresis : public LambdaTransform<IN, OUT, IN, IN, OUT, OUT> {
   Hysteresis(IN lower_threshold, IN upper_threshold, OUT low_output,
              OUT high_output, String config_path = "")
       : LambdaTransform<IN, OUT, IN, IN, OUT, OUT>(
-            &(Hysteresis::function), lower_threshold, upper_threshold,
-            low_output, high_output, hysteresis_param_info, config_path) {}
+            // the lambda function needs to be defined in this awkward
+            // location because it needs to be able to capture `this`
+            [this](IN input, IN lower_threshold, IN upper_threshold,
+                   OUT low_output, OUT high_output) {
+              if (input < lower_threshold) {
+                this->last_value_ = low_output;
+              } else if (upper_threshold <= input) {
+                this->last_value_ = high_output;
+              }
+              // If neither of the above conditions were met, input is between
+              // the lower and upper thresholds (in the hysteresis region)
+              return this->last_value_;
+            },
+            lower_threshold, upper_threshold, low_output, high_output,
+            hysteresis_param_info, config_path) {}
 
  private:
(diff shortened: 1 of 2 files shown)


joelkoz (approved):
Looks ok.  The conversion to std::function<> was not a _requirement_ to make the last value non-static was it?  Why is that?  Is that the only way to be able to use a closure and capture `this`?

mairas:
> Looks ok. The conversion to std::function<> was not a _requirement_ to make the last value non-static was it? Why is that? Is that the only way to be able to use a closure and capture `this`?

I couldn't find a way around it. Capturing `this` changes the lambda signature so that it's no longer compatible with the classic function pointer. Using `std::function` masquerades that signature change while still allowing passing regular function pointers to a LambdaTransform constructor. (Magic!)

I don't think the cost is huge, though. Most likely it's just an `operator()` passthrough call which the optimizer might do away with.

ba58smith:
I'd say "Looks good to me", but that would be a lie. It looks like the dog's breakfast to me! :-) I'm glad you guys understand this stuff.

@mairas, am I going to need to redo the LambdaConsumer in a similar fashion? It has:
```
 private:
  void (*function)(IN input);
```
which looks a lot like what you just replaced for Hysteresis:
`OUT (*function0)(IN input);`
became
`std::function<OUT (IN)> function0;`

I don't THINK so, because I don't think a static data member is involved, but I want to make sure.

mairas:
@ba58smith your code will still work fine. I haven't studied closely how `std::function` works but it's magic.

joelkoz:
> I haven't studied closely how `std::function` works but it's magic.

I agree!! The syntax is so much nicer - I can actually look at it and understand it without my brain hurting.

@mairas You had mentioned a possible performance hit? Do you know that to be true for sure? I kind of assumed it was a syntax recognized directly by the compiler and acted upon (vs. the compiler seeing it as just any other old template function) and that the compiler would optimize for it.  I'll admit however, that I have no basis for this opinion.  I've not studied it either.

mairas:
@joelkoz no, I haven't studied whether the performance hit actually exists. I was just assuming that the direct implementation would include overriding `operator()` and then assigning the arguments and forwarding the call, which of course is something that the compiler could optimize away. But this is all speculation on my part.

mairas:
Assuming this was already approved, merging.

PR #295, opened 2020-12-04
Remove EspSoftwareSerial dependency

### LINKED ISSUE #262: EspSoftwareSerial compiler failure for ESP8266
I opened the following Issue with the EspSoftwareSerial library:
*****************************
When trying to build a PlatformIO Project that uses a brand new version of EspSoftwareSerial (6.10.0), I get:

"pio/libdeps/d1_mini/EspSoftwareSerial/src/SoftwareSerial.h:132:9: error: 'int SoftwareSerial::availableForWrite()' marked override, but does not override",

but only for my ESP8266 Project. My ESP32 project builds correctly.

When I revert to version 6.9.0, PlatformIO for some reason uses version 6.8.5, but the Project builds correctly for ESP8266 and ESP32, with no other changes.

Everything was working until version 6.10.0 was released a few days ago.
*******************************
I got this reply to the above Issue: "Don't use the EspSoftwareSerial from the Arduino library repository with the ESP8266, it's already versioned and part of the ESP8266 Arduino core. For ESP32, do use the Arduino library repository, though. Even with the ESP32, be sure to match new EspSoftwareSerial releases to current, often "unreleased", Git master branch commits."

mairas replied: "OK, so it just needs the platform definition in library.json . And if I understood the last sentence, we should also peg the version."

mairas:
As discussed in #283, this PR removes the EspSoftwareSerial PR altogether. I refactored the gps_compass example to use SoftwareSerial (which is provided by the Arduino SDK) on ESP8266 and another HardwareSerial UART on ESP32. The resulting #ifdefs are a bit messy but at least the dependency could be removed for good.

Fixes #262.

### DIFF
--- a/examples/gps_compass.cpp
+++ b/examples/gps_compass.cpp
@@ -1,6 +1,9 @@
 #include <Arduino.h>
 
-#include "SoftwareSerial.h"
+#ifdef ESP8266
+#include <SoftwareSerial.h>
+#endif
+
 #include "sensesp_app.h"
 #include "wiring_helpers.h"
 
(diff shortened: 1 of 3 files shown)


joelkoz (approved):
Looks good to me!

joelkoz:
Should we hold off on merging this until after we can make the CI fail with the EspSoftwareSerial in there?

mairas:
> Should we hold off on merging this until after we can make the CI fail with the EspSoftwareSerial in there?

I get it now. It doesn't fail because it uses the SensESP platformio.ini file and that approach works fine. Instead, we should build the examples as if they were independent projects depending on the SensESP library. I need to tweak around a bit. Also, I noticed we're also building the esp32-wrover-kit environment in the CI, which is completely redundant and adds almost one-third to the build time. I'll work around that as well.

mairas:
OK, PR #295 now has a failing build due to EspSoftwareSerial. The failure should disappear once this PR has been merged. I'm merging this and crossing my fingers.

PR #296, opened 2020-12-04
Limit the number of examples to be built in CI

### LINKED ISSUE #288: Refactor CI to include building of ESP32
PR #283 uncovered the fact that the continuous builds we have in GitHub build for the esp8266 only and not the ESP32.  We have now experienced platform specific build failures when PRs were checked in that the CI failed to catch.

We need to refactor the CI scripts to include esp32 as a build target. However, so as not to double the build time, we should pair down the list of files used from the `examples` directory to ones that best represent both platforms as well as the most common use cases of SensESP.

mairas:
I began tweaking our `.github/workflows/test.yml` file to enable testing on esp32 as well but then realized that we _are_ building against that, too. Similarly to running `pio run` on the command line, it builds all default environments. It's not terribly visible in the CI details because the esp32 build is after the super-long esp8266 build log, but it is there nevertheless.

So instead of doing anything for that, I limited the number of examples to be built to make the build a bit faster. I removed examples that I thought didn't test any new features. Or at least that was the intention...

Fixes #288. (Actually it wasn't broken to begin with.)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -8,34 +8,25 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        example: [
-          examples/ads1x15_volt_meter.cpp,
-          examples/analog_input.cpp,
-          examples/bme280_example.cpp,
-          examples/bmp280_example.cpp,
-          examples/fuel_flow_meter.cpp,
-          examples/fuel_level_sensor/fuel_level_sensor_example.cpp,
-          examples/gps_compass.cpp,
-          examples/hysteresis.cpp,
-          examples/ina219_example.cpp,
-          examples/lambda_transform.cpp,
-          examples/milone_level_sensor/milone_level_sensor.cpp,
-          examples/onewire_temperature/onewire_temperature_example.cpp,
-          examples/relay_control.cpp,
-          examples/rpm_counter.cpp,
-          examples/sht31_example.cpp,
-          examples/temperature_sender.cpp,
-          examples/thermocouple_temperature_sensor/max31856_thermocouple_example.cpp,
-          examples/ultrasonic_level_sensor/ultrasonic_dvp_example.cpp]
+        example:
+          - examples/ads1x15_volt_meter.cpp
+          - examples/analog_input.cpp
+          - examples/bme280_example.cpp
+          - examples/hysteresis.cpp
+          - examples/lambda_transform.cpp
+          - examples/onewire_temperature/onewire_temperature_example.cpp
+          - examples/relay_control.cpp
+          - examples/rpm_counter.cpp
+          - examples/ultrasonic_level_sensor/ultrasonic_dvp_example.cpp
     steps:
-    - uses: actions/checkout@v1
-    - name: Set up Python
-      uses: actions/setup-python@v1
-    - name: Install dependencies
-      run: |
-        python -m pip install --upgrade pip
-        pip install platformio
-    - name: Run PlatformIO
-      run: platformio ci -v --lib . -c ./platformio.ini
-      env:
-        PLATFORMIO_CI_SRC: ${{ matrix.example }}
+      - uses: actions/checkout@v1
+      - name: Set up Python
+        uses: actions/setup-python@v1
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install platformio
+      - name: Run PlatformIO
+        run: platformio ci -v --lib . -c ./platformio.ini
+        env:
+          PLATFORMIO_CI_SRC: ${{ matrix.example }}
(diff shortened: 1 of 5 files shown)


mairas:
Now the CI builds install SensESP as a dependency like a separate project would. This approach exposes correctly the EspSoftwareSerial build issues shown in the failing checks. Those failures will disappear once PR #295 is merged.

.github/workflows/test.yml
```diff
@@ -8,34 +8,30 @@ jobs:
...
-          examples/sht31_example.cpp,
-          examples/temperature_sender.cpp,
-          examples/thermocouple_temperature_sensor/max31856_thermocouple_example.cpp,
-          examples/ultrasonic_level_sensor/ultrasonic_dvp_example.cpp]
+        example:
+          - examples/ads1x15_volt_meter.cpp
+          - examples/analog_input.cpp
+          - examples/bme280_example.cpp
+          - examples/hysteresis.cpp
+          - examples/ina219_example.cpp
```
ba58smith, line 16: This is SO similar to the BME280 example, I don't think it's testing anything new. I think it can be dropped from the tests.

ci/platformio-d1_mini.ini
```diff
@@ -0,0 +1,18 @@
...
+framework = arduino
+lib_ldf_mode = deep
+monitor_speed = 115200
+lib_deps = ${PROJDIR}
+
+[env:d1_mini]
+platform = espressif8266
+board = d1_mini
+build_flags =
+   -Wl,-Teagle.flash.4m1m.ld
```
ba58smith, line 14: The `-Wl,-Teagle.flash.4m1m.ld` build flag has been deprecated. Delete this line, then add this ABOVE `build_flags =`:
`board_build.ldscript = eagle.flash.4m1m.ld`
See PR #292

.github/workflows/test.yml
mairas: ina219? OK, will drop.

ci/platformio-d1_mini.ini
mairas: OK, will do. I think that's still present in the project platformio.ini, then.

joelkoz (commented):
I've been reading through the comments and changes, and now I think we have a little bit of a dilemma.

@ba58smith stated these two examples are similar:

`examples/bme280_example.cpp`
`examples/ina219_example.cpp`

so the second one was removed from the CI build.  If we make a change that requires a code update to both bme280 and ina219, we may miss that fact. That is, the change is checked in, the build breaks for `bme280_example`, we update `bme280_example`, the build passes, but we now have a second example that may need to be fixed but we don't know it.

By having a subset of the examples in the CI script, we run the risk of having some examples that don't compile.

That may be a small and acceptable risk, but I wanted to point it out.  Maybe we create a second workflow that is the entire set of examples, and have that run once per night? I'm not up to speed on GitHub workflows, but I am under the impression that running on check-in is but one of many options? I don't think having this second workflow in place is a requirement for this PR, but maybe we should create an issue for it for some future work?

.github/workflows/test.yml
```diff
@@ -8,34 +8,26 @@ jobs:
...
\ No newline at end of file
+      - uses: actions/checkout@v1
+      - name: Set up Python
+        uses: actions/setup-python@v1
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install platformio
+      - name: Run PlatformIO
+        run: platformio ci -v --lib . -c ./platformio.ini
```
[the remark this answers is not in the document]
mairas: Yes, that was the case previously. That's why we didn't catch all those EspSoftwareSerial issues. But check the updates in the PR; all that has changed now.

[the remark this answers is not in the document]
joelkoz: Ha!  I actually wrote that comment yesterday, but I guess I forgot to tell GitHub I was starting a review, so it held on to that comment!  In any event, glad it got resolved.

mairas:
> That may be a small and acceptable risk, but I wanted to point it out. Maybe we create a second workflow that is the entire set of examples, and have that run once per night? I'm not up to speed on GitHub workflows, but I am under the impression that running on check-in is but one of many options? I don't think having this second workflow in place is a requirement for this PR, but maybe we should create an issue for it for some future work?

It would indeed be possible to run more thorough tests at other times; however, with such schemes there's always the problem that when the build eventually fails, the code is already merged in and the PR submitter can't be so easily encouraged to fix the issue. Anyway, most of the sensor and transform issues cause the build to fail even when those classes aren't specifically used. Strongly and statically typed languages are nice in that regard.

ba58smith:
I suppose a nightly build of every example would give us the opportunity to catch problems before merging most of the time, so seems worth doing.

mairas:
> I suppose a nightly build of every example would give us the opportunity to catch problems before merging most of the time, so seems worth doing.

I created a new issues for this: #300

ba58smith (approved):
Looks good to me.

PR #297, opened 2020-12-04
Add Typecast transform

joelkoz:
Typecast is a template transform that allows you to specify a conversion from one type to another using a lambda function passed into the constructor.  Two implementations are also included: BoolToInt and IntToBool.

This should solve some of the problems we've been having between mismatched inputs and outputs.  I needed it for a project but I'm including it here because I thought I heard some discussions about the need for this.


ba58smith (approved):
Looks good to me.

Would IntToFloat be a good addition, or is that not necessary, because the compiler will do it automatically?

joelkoz:
> Would IntToFloat be a good addition, or is that not necessary, because the compiler will do it automatically?

That is a good question.  I actually had to test it, and it looks like the compiler fails to find a valid `connect_to()` match if an int is connected to a float.  So yes, we need that one to.  Very easy to add.

mairas:
This is something sorely needed, so a good addition. Nothing wrong with the implementation, but it looks pretty much like a duplicate of the zero-parameter LambdaTransform. Just wondering whether the simple typecast cases could be implemented with a fixed template function of `[](IN input) { return (OUT)input; } -> OUT` and the other cases with a LambdaTransform?

I like how the specific cases are defined by inheriting and defining the lambda in the constructor initializer. Clever.

joelkoz:
> Just wondering whether the simple typecast cases could be implemented with a fixed template function of `[](IN input) { return (OUT)input; } -> OUT` and the other cases with a LambdaTransform?

Whoa!  I never thought about doing a templated lambda function.  I started playing around with this, and I got the idea for a templated lambda function as a default parameter to Typecast, and it worked!

So now, we still have the option of passing in a custom lambda, but the default is to use a compiler derived typecast.  That means any class that is capable of using the c++ cast operator automatically has an implementation!  That also means that if we make up our own types in the future (for example, my ClickType class), we can define an opererator overload like `operator int()`, and the Typecast class can now be used unmodified.  Kinda cool!

The implementation of the "simple" cases (`IntToBool`, `IntToFloat`, etc.) now require only a simple typedef, and that is just for saving typing.  Simply using `new Typecast<int,  bool>()` as is works fine now with no additional work.

mairas:
Template lambda as a _default value_ for the constructor? That's... wicked. You nasty man. I like it. Merge at will.

joelkoz:
> Template lambda as a _default value_ for the constructor? That's... wicked. You nasty man. I like it. Merge at will.

lol.  I was pretty stunned myself. I had never thought of doing it before, and I wasn't sure it would even work until I tried it! No reason it SHOULDN'T work, but it just never occurred to me.  The technique opens up a lot of possibilities...

src/transforms/typecast.h
```diff
@@ -0,0 +1,53 @@
...
+ * @tparam The data type the typecast outputs after conversion
+ */
+template <typename IN, typename OUT>
+class Typecast : public LambdaTransform<IN, OUT>
+{
+
+public:
+    Typecast(std::function<OUT (IN)> cast =
+       [](IN input) -> OUT { return (OUT)input; } ) : LambdaTransform<IN, OUT>(cast) {}
+};
```
ba58smith, line 31: For my education: can you explain why the `(IN)` in `std::function<OUT (IN)>` is in parentheses?

PR #298, opened 2020-12-05
Sk output metadata support

joelkoz:
This PR allows SensESP to send metadata describing the Signal K paths it is sending along with the first delta update it reports to the server.  Metadata is described in the Signal K specification here: https://signalk.org/specification/1.5.0/doc/data_model_metadata.html

Using this feature requires the Signal K Node Server version 1.37.3 or greater.


mairas:
This is an important addition, and at a quick look the PR would seem to be OK. I'd still want to read through it more carefully tomorrow.

src/signalk/signalk_delta.cpp
```diff
@@ -4,8 +4,10 @@
 #include "ArduinoJson.h"
 #include "sensesp.h"
 
+#include "signalk/signalk_emitter.h"
+
 SKDelta::SKDelta(const String& hostname, unsigned int max_buffer_size)
-    : hostname{hostname}, max_buffer_size{max_buffer_size} {}
+    : hostname{hostname}, max_buffer_size{max_buffer_size}, meta_sent_{false} {}
```
mairas, line 10: About the `meta_sent_` handling: didn't the new spec say the metadata should be sent once per connection? If the server is rebooted or the websocket gets disconnected for any other reason, I think `meta_sent_` should be cleared.

src/signalk/signalk_emitter.h
```diff
@@ -17,7 +17,40 @@
...
+   * Metadata holds optional Signal K meta data that is associated with
+   * the sk_path the SKEmitter emits. If Metadata is defined, it will
+   * be sent to the Signal K server along with the first delta emitted.
+   * Note that this structure is a subset of the complete meta model.
+   * These are the most common properties. More complex strctures
+   * can be emitted by overriding add_metadata()
+   * @see https://signalk.org/specification/1.5.0/doc/data_model_metadata.html
+   * @see add_metadata()
+   */
+  struct Metadata {
```
mairas, line 33: What was your reasoning for defining this inside SKEmitter? Given that it's an object instantiated by the end user, it'd feel more natural for me to have it in the global namespace. (Of course, we could debate whether we should start using namespaces for real, but maybe another time...)

src/signalk/signalk_emitter.cpp
```diff
@@ -2,6 +2,45 @@
...
 SKEmitter::SKEmitter(String sk_path) : sk_path{sk_path} {
   sources.push_back(this);
 }
+
+void SKEmitter::add_metadata(JsonArray& meta) {
+
+   Metadata* my_meta = this->get_metadata();
+   if (my_meta != NULL) {
+
+      JsonObject json = meta.createNestedObject();
```
mairas, line 22: Minor detail: should Metadata know how to serialize itself as a Json object?

mairas (changes requested):
Overall very nice. Some minor comments. The `meta_sent_` lifecycle is the only thing that concerns me (assuming I got it right).

src/signalk/signalk_delta.cpp
joelkoz: I don't think the spec said anything official - they just didn't want it to be sent with each delta.  I agree however - if we lose the connection and re-connect, that could very well be because the server rebooted.  I'll expose this via a method, and then have ws_client modify it.  Unless ws_client can be subscribed to, in which case I'll do it the "right way."

mairas: Hmm, subscribing to it is indeed possible! I implemented the interface for status displays etc but the status led blinker uses it as well. And now that I remember, I'll create an issue to rename the interface classes to something more specific.

mairas: Correcting myself, LedBlinker et al are OK (in my opinion) but LedController should be StatusLedController or something like that.

src/signalk/signalk_emitter.h
joelkoz: I've always put data structures that are subordinate to classes inside those classes, especially when the name is generic.  On the other hand, when I went to actually use this in an app, SKEmitter seemed strange as its not a class an end user sees much if at all.  SKOutput would have been a more natural place.  And certainly it would be even easier if this were a global class. If so, I think it needs a prefix to act as a pseudo namespace, as simply "Metadata" is too generic I think.  Maybe `SKOutputMetadata`?  Or, we can stick with simply `Metadata`, as in practice, I don't think there is a huge risk of name collision.  The compromise is probably `SKMetadata`.  Ok with that?

mairas: SKMetadata or SKOutputMetadata, me thinks. It contains Signal K metadata, so I'd lean on SKMetadata.

src/signalk/signalk_emitter.cpp
joelkoz: The `Metadata` struct wasn't really intended to be the final word in serialization. As I mentioned in the comments, the `Metadata` struct is only a subset of the possible metadata that can be described.  I didn't think it was worth the effort to add support for things like Zones and alert methods as that stuff added complexity that I didn't think would be used much.  That meant we needed a way to extend it in the future. My goal was to have the "write the metadata" code be a `virtual` method that could be overwritten by any descendant of SKEmitter that needed the specialization, but there would be a simple implementation that would work in most cases. Now, I suppose we could instead make the default implementation be "ask Metadata to fill out the json", and any specialization would be done by overriding a `virtual` serialize method that is in `Metadata`.  The technique to extend the functionality would then be "create a descendant of Metadata and re-do the serialization class" vs. "make a custom version of SKOutput".  In theory that is going to add one more level of indirection as this method would just delegate to `Metadata`, but in practice, it's a negligible performance hit.  This approach makes even more sense if `Metadata` is made a global class.  Now that I've typed all this and seen the pieces laid out, I kind of like that approach better.  I'll do that.

src/signalk/signalk_delta.cpp
joelkoz: After studying the code, it looks like SKDelta is co-owned by `SensESPApp` and `WSClient`, but SKDelta has no awareness of either. To minimize the number of changes, it was much simpler to expose a "reset meta send" method on SKDelta, and just have WSClient call that whenever it connected to an SK server. There were other options of course, but this was the quickest and safest.

examples/analog_input.cpp
```diff
@@ -60,7 +60,7 @@ ReactESP app([]() {
   // Wire up the output of the analog input to the Linear transform,
   // and then output the results to the Signal K server.
```
ba58smith, line 61: Can you add to the comments immediately above? Something about what `new SKEmitter::Metadata("ratio")));` does.

ba58smith (approved):
Looks good to me! Just one small comment about a comment.

examples/analog_input.cpp
joelkoz: Sure.  FYI, "ratio" is the "units" field. It is the official name in the Signal K spec of a "ratio of one number to another" (aka a percentage represented as a number between 0.0 and 1.0)

mairas:
Looks good to me too. Could you clean the merge commits from the PR before merging? Something like
```
git checkout master
git pull upstream master
git checkout SKOutput-meta-data-support
git rebase master
# check the results before pushing
git log
git push -f origin SKOutput-meta-data-support
```
might clean them away.

The reason is that we're trying to keep the commit log clean to make it as readable and informative as possible.

PR #299, opened 2020-12-06
SmartSwitchController with examples

joelkoz:
This is the final bit of code that completes the original SmartSwitch PR.  It contains just the SmartSwitchController, as well as two example source files. All of the other changes have been merged into master.


src/controllers/smart_switch_controller.cpp
```diff
@@ -0,0 +1,46 @@
+#include "smart_switch_controller.h"
+#include "transforms/truth_text.h"
+
+void SmartSwitchController::enable() {
+    this->emit(is_on);
+}
```
ba58smith, line 6: Since `enable()` is called at the very end of main.cpp, does this mean that SmartSwitchController will always emit `false` at the very beginning of program execution? Won't that turn off a light that's on before the program starts?

examples/smart_switch_example.cpp
```diff
@@ -0,0 +1,126 @@
...
+
+
+  // Create a switch controller to handle the user press logic and 
+  // connect it to the load switch...
+  SmartSwitchController* controller = new SmartSwitchController();
+  controller->connect_to(load_switch);
+
+
+  // Connect a physical button that will feed manual click types into the controller...
+  DigitalInputValue* btn = new DigitalInputValue(PIN_BUTTON, INPUT, CHANGE, 100);
```
ba58smith, line 99: PR #285 changes DigitalInputValue to DigitalInputState. (Although because of a typedef, DigitalInputValue will still work.) If you have any other changes to make to this PR, you might want to include this, but I wouldn't if it's the only thing you need to change.

examples/smart_switch_example.cpp
```diff
@@ -0,0 +1,126 @@
...
+  // This allows any device on the SignalK network that can make
+  // such a request to also control the state of our switch.
+  auto* sk_listener = new SKStringPutRequestListener(sk_path);
+  sk_listener->connect_to(controller);
+
+
+  // Finally, specify that the load switch should report its status at regular
+  // intervals back to the SignalK server...
+  load_switch->connect_to(new RepeatReport<bool>(10000, config_path_repeat))
+             ->connect_to(new SKOutputBool(sk_path, config_path_sk_output));
```
ba58smith, line 121: I've never understood the reason for RepeatReport. Can you explain it in the comment above where it's used?

examples/virtual_switch_example.cpp
```diff
@@ -0,0 +1,120 @@
...
+#include "signalk/signalk_output.h"
+#include "signalk/signalk_put_request.h"
+#include "signalk/signalk_value_listener.h"
+#include "system/rgb_led.h"
+#include "transforms/debounce.h"
+#include "transforms/click_type.h"
+#include "transforms/repeat_report.h"
+#include "transforms/press_repeater.h"
+
+// This example implements a virtual switch that can remotely control the
```
mairas, line 18: The example is good but I think the file name is misleading. It's not a virtual switch because there's a real, physical switch input present. How about `remote_switch_example.cpp` instead? And I think both examples could go to a common subdirectory under `examples`, something like `smart_switch`.

mairas (approved):
Looks good to me. A minor suggestion; ignore it if you can't be bothered.

src/controllers/smart_switch_controller.cpp
joelkoz: Yes, it will.  But remember - in this case, the SmartSwitchController IS THE DEVICE. The state of the lights follows the state stored in the code.  I'm pretty sure if you reset a Wemos, the digital outputs are reset, in which case the relay is reset.  So this line just keeps everything in sync. Its there just to make sure the relay and the led status light and the status on the Signal K server are all in sync.
  That being said, I suppose it's possible that each state change emitted by the controller could also be saved to flash memory and read back in at startup, similar to our configuration options. Thus, if there is a power failure of any sort, the switch state would be restored to its last state.  That functionality may be best saved for an external class, however (one that feeds the input of SmartSwitchController, and any other transform that wants its state restored upon startup).

examples/smart_switch_example.cpp
joelkoz: Yeah - I was waiting for that PR to get merged.  I'll do that now.  It's best to keep the example code fresh if possible.

examples/virtual_switch_example.cpp
joelkoz: Good idea.  I will do that as I have some other minor changes to make and this is easy enough.

examples/smart_switch_example.cpp
joelkoz: How about this:

```
// Finally, connect the load switch to an SKOutput so it reports its state to the Signal K server.  Since the load
// switch only reports its state whenever it changes (and switches like light switches change infrequently), send
// it through a `RepeatReport` transform, which will cause the state to be reported to the server every 10
// seconds, regardless of whether or not it has changed.  That keeps the value on the server fresh and lets the
// server know the switch is still alive.
```

Issue #301, opened 2020-12-07
LedController should be renamed


mairas:
LedController is the controller class for the system status Led (ordinarily the onboard LED on GPIO 2). The class name is too specific and should be renamed to something like StatusLedController.

Similarly, VisualOutputController is too generic. StatusOutputController, maybe.

joelkoz:
I took a peek at these two classes.  My suggestions:

1. Rename `VisualOutputController` to `SystemStatusConsumer`
IMO, this class is really like an interface definition, and its primary purpose is to consume input regarding the network status. There might be a number of other purposes for doing this - not just creating a visual indicator. Given this new name, we may want to move it out of the `controller` directory and into the `system` directory

2. Rename `LedController` to `SystemStatusLed` as it represents a single led that outputs the system status.

3. We might benefit from breaking the implementation of this new`SystemStatusLed` into a base `SystemStatusController` that holds the implementation of the `set_input()` methods and defines the  `set_wifi_*()` and `set_ws_*()` methods as abstract (or better yet, give them default empty implementations).  Again, responding to the system status for things other than a visual indicator might be beneficial.  That interim class would be much more in line with the definition of a "controller" IMO, and deserves to be placed in the controller directory.

4. Refactor `SystemStatusController` to be a `ValueProducer<SystemStatusValue>`
To really leverage this work, the other possibility would be to make `SystemStatusController` also be a `ValueProducer` of events, similar to the way `ClickTypes` was implemented.  By doing that we'd have network events consumed by a SystemStatusController (a specialized version of `SystemStatusConsumer`), translated into a system event enumeration value then output, and then `SystemStatusLed` could be implemented as a simple `ValueConsumer<SystemStatusValue>` of those enumerations.

Finally if the `SystemStatusController` were a singleton (explicitly designed or implied by being a globally available member of `sensesp_app`), then its output could be connected to more than one entity that might be interested in these status values (like the SKDelta that needed to reset its meta flag when the system re-connected to the server), or any user code that needs to react to the server being present (or not).

mairas:
Good proposals. I'll try using them as a guideline and see what happens.

PR #302, opened 2020-12-08
System status controller refactoring

mairas:
This is pure refactoring (no functional changes whatsoever) of VisualOutputController and LedController classes according to suggestions presented in #301. Only steps 1-3 from @joelkoz's suggestions are implemented. I decided step four would deserve its own PR.


src/controllers/system_status_controller.h
```diff
@@ -0,0 +1,42 @@
...
+ * Classes inheriting from SystemStatusController should override the
+ * set_wifi_* and set_ws_* methods to take the relevant action when such
+ * an event occurs.
+ */
+class SystemStatusController : public SystemStatusConsumer {
+ protected:
+  virtual void set_wifi_no_ap() {}
+  virtual void set_wifi_disconnected() {}
+  virtual void set_wifi_connected() {}
+  virtual void set_wifimanager_active() {}
```
joelkoz, line 18: What does this status do? This method stands out as unique in its name from the other three (and actually, its pattern is different than the other 7 methods here).  Not that it matters much, but would `set_wifi_manageractive()` be a better name?

joelkoz (approved):
Looks good.  Only one minor comment (a question really, but something to consider)

src/controllers/system_status_controller.h
mairas: That status is set when WiFiManager is active. That is, when the device has its own active portal and waits for a client to connect and to set the wifi configuration. I don't like the name  much either but I think it's somewhat clearer than your proposal. Hm, how about `set_wifimanager_activated()`? It's more in line with the others. I'll do that and see whether you approve or not. :-)

src/controllers/system_status_controller.h
```diff
@@ -0,0 +1,42 @@
...
+ * Classes inheriting from SystemStatusController should override the
+ * set_wifi_* and set_ws_* methods to take the relevant action when such
+ * an event occurs.
+ */
+class SystemStatusController : public SystemStatusConsumer {
+ protected:
+  virtual void set_wifi_no_ap() {}
+  virtual void set_wifi_disconnected() {}
+  virtual void set_wifi_connected() {}
+  virtual void set_wifimanager_activated() {}
```
joelkoz, line 18: Ok, I understand now.  It's not really related to the wifi connectivity status, but it's instead whether the device is a host of its own network or a client.  This is VERY minor, but my suggestion is to put a blank line after the first three methods to set `set_wifimanager_activated()` apart from the other three.  I agree that it is named appropriately because it IS different than the first three.  Of course, the better approach would be to add a quick Doxygen doc above this one method.

This actually brings up another question: if the WIFIManager is active, is `set_wifi_connected()` is called when a client connects to the device?  As I was looking at this, I wasn't thinking about the "portal mode" at all, so I would have assumed (and would have written) code that responded to `set_wifi_connected()` as an event when the device connected as a client to our target server's wifi.  If the two "connected/disconnect" methods share this double meaning, how about adding a `bool` parameter that indicates if we are in host or client mode?

I don't want to slow down your other changes.  These suggested enhancements could be saved for another PR. But, if they're easy...

joelkoz (approved):
I left one more comment re: the "wifi lifecycle".  You can implement now, save for a future PR or ignore.  There is enough information "as is" to deduce where in the lifecycle we are, but the suggested parameters would eliminate the need for future code to track it themselves.

ba58smith (approved):
This one's over my head. I trust you guys.

src/controllers/system_status_controller.h
mairas: I'm not sure it would be easy to get information on the WiFiManager state. OK, a bit, yes: it's possible to tell if WifiManager is trying to connect to an AP or if it's entering the configuration mode (becomes an AP itself). But that's it. So, those two additional states could be added to the mix. However, I don't think it's possible to know whether a client connects to the WifiManager AP. The AP activation is a blocking function, so we really can't do anything during the that time, anyway.

This bit of code will be mangled by the next PR, so I'll leave the WifiManager state as is for the time being.

PR #307, opened 2020-12-09
DigitalInputChange: always output first value

ba58smith:
The output is compared to the previous output, and if it's not different, it's not emitted. This caused the intial value to not be output, if the first value was 0 (button goes to GND when pressed). Now the initial value will always be output.


joelkoz (approved):
Look good to me

ba58smith:
Merging.

PR #309, opened 2020-12-09
Fix #213 Access Token Lost on network problem

### LINKED ISSUE #213: Add more debug for "Bad access token detected"
There have been some reported issues via Slack where SensESP have got successful authorization from Signal K Server but fails when reconnecting. SensESP output "Bad access token detected". To figure this out more debug messages is probably needed.

joelkoz:
This fix is for the loss of device authentication when the SignalK server or network goes down AFTER the initial authentication check. SensESP would clear the token and request a new one in this scenario.  Fixes #213 

Source code provided by @JohnySeven

### DIFF
--- a/src/net/ws_client.cpp
+++ b/src/net/ws_client.cpp
@@ -81,7 +81,9 @@ void WSClient::connect_loop() {
 }
 
 void WSClient::on_disconnected() {
-  if (this->connection_state == kWSConnecting && server_detected) {
+  if (this->connection_state == kWSConnecting && 
+      server_detected && 
+      !token_test_success) {
     // Going from connecting directly to disconnect when we
     // know we have found and talked to the server usually means
     // the authentication token is bad.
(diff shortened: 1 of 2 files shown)


joelkoz:
Also note: this is not the complete set of changes from @JohnySeven. This PR relates only to the code specifically dealing with the Access Token problem.

joelkoz:
I have tested this with downed SignalK server and downed wifi AP. Both situations seem to recover fine now.

mairas:
The `token_test_success` variable doesn't seem to be saved over reboots, right? Isn't that part of the issue - you kill the power, and when the power is restored, the ESP boots faster than the SK server and token auth fails. (It's also possible I've completely off the track here -- I'm a bit hazy on the details...)

joelkoz:
No, `token_test_success` is not saved, but it doesn't have to be.  The key to this fix is here:

```
void WSClient::on_disconnected() {
  if (this->connection_state == kWSConnecting && 
      server_detected && 
      !token_test_success) {   // <-- This is new
    // Going from connecting directly to disconnect when we
    // know we have found and talked to the server usually means
    // the authentication token is bad.
    debugW("Bad access token detected. Setting token to null.");
    auth_token = NULL_AUTH_TOKEN;
    save_configuration();
  }
 ```

In the case you describe (the device boots and checks before the Signal K server is there to answer) is already handled because until the initial handshake with a Signal K server, the `server_detected` is `false`.  The current code handles that fine. 

The bug is actually when things are going great, and then the server disappears (it reboots, network goes down, etc.) but the device is still running.  In that situation, `server_detected` is `true`. But when the server is not there to answer, this `on_disconnected()` gets called.  Without the new `token_test_success`, the token gets cleared, as `server_detected` is now `true` (it doesn't get set to `false` until the end of this method).

The normal "rejected" path is the server answers (`server_detected == true`), but it drops the connection when it's handed a bad token. That case still works, as `token_test_success` never gets set to `true` in the first place.

**Comment by JohnySeven** [comment]
Great @joelkoz! Looking good to me!

**Comment by mairas** [comment]
Thanks for the explanation! Looks good to me!

File: src/net/ws_client.cpp [hunk 1]
```diff
@@ -254,49 +256,69 @@ bool WSClient::get_mdns_service(String& server_address, uint16_t& server_port) {
...
+  debugI("WSClient connect attempt (state=%d)", connection_state.get());
+
   if (connection_state != kWSDisconnected) {
     return;
   }
-  debugD("Initiating connection");
 
-  connection_state = kWSAuthorizing;
+  if (WiFi.isConnected()) {
+    debugD("Initiating connection");
```
ba58smith, line 266: This is a tad confusing: if `WiFi.isConnected`, then why "Intitiate connection"? How about one or two more words: "Initiating connection to SK Server" (or whatever, specifically, it's referring to).

src/net/ws_client.cpp
```diff
@@ -254,49 +256,69 @@ bool WSClient::get_mdns_service(String& server_address, uint16_t& server_port) {
...
+    if (this->server_address.isEmpty()) {
+      if (!get_mdns_service(server_address, server_port)) {
+        debugE("No Signal K server found in network when using mDNS service!");
+      } else {
+        debugI("Signal K server has been found at address %s:%d by mDNS.",
+              server_address.c_str(), server_port);
+      }
+    }
+
+    if (!server_address.isEmpty() && server_port > 0) {
```
ba58smith, line 282: Line 273 has `this->` before `server_address`, but line 282 doesn't, nor do several other `server_address` references in this section (and elsewhere, for that matter). Is that intentional? Is it important?
I think this illustrates the value of putting the trailing `_` on member variable names: `server_address_` means `this->server_address_`, and `server_address` is something else.

ba58smith (approved):
Logic looks good. Just two comments on clarity, which I would really like to see, but aren't critical.

src/net/ws_client.cpp
joelkoz: Sure.  This comment was actually in there already as is, but I can change it while I'm at it.

src/net/ws_client.cpp
joelkoz: Agree it's inconsistent.  This PR is a minor change to some very old code that has been there since the very beginning. I wouldn't be surprised to learn some fragments were borrowed from the old SigSens project.

using `this->` in front of member variables is technically redundant - it's like putting parenthesis around the multiplication part of a formula when precedence dictates that operation anyway. So the direct answer to the question of "is it important" is "no."  For the sake of clarity, however, I usually _do_ use the `this->` prefix in my personal code on other projects to distinguish members of classes, but then again, I do _not_ use the "_" suffix. Either case generates the same object code by the compiler.

I can't argue that we shouldn't refactor some of the older code base to match the current coding style, but I think that should be a PR unto itself. When working with old code, I tend to think leave everything be, including formatting, so it's easier to see what the real changes are.  Code style updates should be separate PRs.  That allows one or the other to be reverted independently. This is particularly true for critical base code used by every project (like this, which manages the network connection to the server).

mairas: Yup, quite a bit of the original `ws_client` implementation came from SigKSens...

joelkoz: Actually, upon closer inspection, in this particular method, there is a local variable named `server_address`, so in this instance, the `this->` part is critical. It is used to distinguish between the member variable `server_name` vs. the local variable `server_name`.  You are correct that the "_" suffix would have really helped clarify that.

ba58smith: I'm reading this on my phone so can't see all the code, but the crux of my comment is that the code you just added may be inconsistent. With "this" in one case and without it in the other. While it would be nice to fix all of the ambiguity with "_", I agree that shouldn't be in this PR. But at least be consistent with "this" in the new code you're adding, for the sake of the person who eventually tackles the "_".

ba58smith: Oh... You may not have added that code. It just looks like it because of the way GitHub is showing the diff.

joelkoz: @ba58smith - I don't want you to think I am dismissing your concerns, but its important to note that very little new code exists here.  If you are looking at the diff, be sure to use the "ignore white space" feature.  There was a large section of code that was indented two spaces to be wrapped inside an `if` statement.  All of the code you are referring to is pre-existing code, simply shifted over two spaces. Ignoring white space in the diff reveals just the added `if` statement, as well as the new setting of the `token_test_success`.

I'll create an issue so we don't forget this file needs some refactoring.

PR #310, opened 2020-12-10
Refactor SystemStatusController into ValueProducer

### LINKED ISSUE #301: LedController should be renamed

mairas:
`SystemStatusController` has been so far a `ValueConsumer` for `WSConnectionState` and `WifiState` and the use model was to inherit it to define actual behavior. This PR refactors `SystemStatusController` into a ValueProducer, permitting users to simply consume the `SystemStatus` updates instead.

The change has also been implemented in SensESPStatusDisplay.

Fixes #301.


mairas:
Oh, I think I still need to update the member variable naming...

src/controllers/system_status_controller.h
```diff
@@ -1,42 +1,40 @@
...
+                               public ValueConsumer<WSConnectionState>,
+                               public ValueProducer<SystemStatus> {
  public:
   SystemStatusController() {}
 
   /// ValueConsumer interface for ValueConsumer<WifiState> (Networking object
   /// state updates)
-  virtual void set_input(WifiState new_value,
-                         uint8_t input_channel = 0) override;
+  virtual void set_input(WifiState new_value, uint8_t input_channel = 0);
```
joelkoz, line 33: Just curious - any reason why you removed the `override` keyword? Declaring a class as a `ValueConsumer<T>` does bring in a `set_input(T input)`.

src/net/networking.h
```diff
@@ -11,11 +11,11 @@
 #include "system/observablevalue.h"
 #include "system/valueproducer.h"
 
-enum WifiState {
+enum class WifiState {
```
joelkoz, line 14: Another curiousity question: did adding `class` help resolve the ambiguity problem you were having w/ `enum` vs. `int`? Is this a pattern we should use in the future?

joelkoz (approved):
Looks good!  I only had two questions, but they don't affect merge approval.

src/controllers/system_status_controller.h
mairas: After doing the main refactoring I figured the SystemStatusConsumer base class was no longer useful and I removed it. Any classes should either observe SystemStatusController or the Network or WSClient classes directly. The method declarations were there, so there was nothing to override any more.

src/net/networking.h
mairas: It wasn't related to that. Fixing the diamond inheritance pattern solved the issue. The enum classes are still based on integral types but they don't litter the global namespace as badly because you have to refer to them via the enum type name (WifiState::kConnected instead of just kConnected). Not just my call but I prefer being explicit. 😊

joelkoz: Oh nice!  I totally agree - if you have an standard C enumeration with five values, you end up with six global symbols (the enumeration itself, and the five values). I never realized you could simply add "class" in front and hide them. The other advantage of requiring the namespace is you don't have to rely on some kind of prefix to ensure you don't end up with a name clash using common names

src/controllers/system_status_controller.h
joelkoz: Is that true? I know they are no longer from SystemStatusConsumer, but they _are_ from ValueConsumer. This class now inherits directly from three templated versions of ValueConsumer.  The ValueConsumer template has a `set_input()` method declared that you inherit. Those are the ones you are now overriding, right?
It's a minor point, but I just wanted to make sure we are on the same page.

mairas: No, you're of course right. I must've gotten an error after the last `ValueConsumer<int>` interface got removed and wrongly interpreted that `override` was no longer allowed. For the remaining two consumers it's exactly like you said. Fixing.

mairas:
Restored the mistakenly removed override qualifiers Joel commented on. I'll reformat the files and merge.

PR #312, opened 2020-12-11
Estimate the required delta serialization buffer size based on data

mairas:
The delta output serialization buffer was fixed to 1024 characters. That wouldn't be enough if there're many paths or metadata to transmit.

This PR adds some crummy heuristics for estimating the required buffer size based on the data to be transmitted. An added benefit is somewhat smaller buffers for simple programs.

### DIFF
--- a/src/signalk/signalk_delta.cpp
+++ b/src/signalk/signalk_delta.cpp
@@ -3,7 +3,6 @@
 #include "Arduino.h"
 #include "ArduinoJson.h"
 #include "sensesp.h"
-
 #include "signalk/signalk_emitter.h"
 
 SKDelta::SKDelta(const String& hostname, unsigned int max_buffer_size)
@@ -18,8 +17,62 @@ void SKDelta::append(const String val) {
 
 bool SKDelta::data_available() { return buffer.size() > 0; }
 
+unsigned int SKDelta::get_doc_size_estimate() {
+  int buf_size = buffer.size();
+  int estimate =
+      2 * JSON_OBJECT_SIZE(1) +         // source and one update
+      JSON_ARRAY_SIZE(1) +              // one update
+      JSON_ARRAY_SIZE(buf_size) +       // buf_size values in the update
+      buf_size * JSON_OBJECT_SIZE(2) +  // two key-value pairs in each object
+      200;  // some extra headroom to hide embarrassing bugs
+
+  for (auto item : buffer) {
+    // also reserve space for the pre-rendered strings
+    estimate += item.length() + 1;
+  }
+  return estimate;  
+}
+
+unsigned int SKDelta::get_metadata_size_estimate() {
+  int num_metadata = SKEmitter::get_sources().size();
+  int estimate = JSON_ARRAY_SIZE(num_metadata);
+
+  int num_fields;
+  auto update_estimate = [&](String& field) {
+    if (!field.isEmpty()) {
+      num_fields++;
+      estimate += field.length() + 1;
+    }
+  };
+
+  for (auto const& source : SKEmitter::get_sources()) {
+    num_fields = 0;
+    auto metadata = source->get_metadata();
+    if (metadata == NULL) {
+      continue;
+    }
+    update_estimate(metadata->units_);
+    update_estimate(metadata->display_name_);
+    update_estimate(metadata->description_);
+    update_estimate(metadata->short_name_);
+    
+    estimate += JSON_OBJECT_SIZE(num_fields);
+  }
+  return estimate;
+}
+
 void SKDelta::get_delta(String& output) {
-  DynamicJsonDocument jsonDoc(1024);
+  // estimate the size of the serialized json string
+
+  unsigned int doc_size_estimate = get_doc_size_estimate();
+
+  debugD("doc size estimate: %d", doc_size_estimate);
+
+  if (!meta_sent_) {
+    doc_size_estimate += JSON_OBJECT_SIZE(1) + get_metadata_size_estimate();
+  }
+
+  DynamicJsonDocument jsonDoc(doc_size_estimate);
 
   // JsonObject delta = jsonDoc.as<JsonObject>();
   JsonArray updates = jsonDoc.createNestedArray("updates");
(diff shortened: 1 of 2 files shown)


ba58smith (approved):
Looks good.
Related: I just did a quick search of all of SensESP for DynamicJsonDocument, and there are a LOT of places where we use the default of 1024. (I even found a "FIXME" comment saying to look at all of them in the program and set them to the correct size.) Is that worth doing? Save 100 here, 200 there, 500 somewhere else, and it'll start to add up.

mairas:
Oh man, I repeated the search and yes, that's a lot of results. Mostly it
doesn't really matter because the buffers are only temporary: the memory
gets released as soon as the function is finished. It seems like most uses
are only related to serializing a small fixed-size document, but whenever
there's a previously unknown number of elements, it might be better to fix
them. We've had two similar bugs recently, so it might be better to be a
bit proactive about it.

KEGustafsson:
Verified and working. Now Json string is not anymore cut due to 1k limit.

ba58smith:
@mairas, I'll make a PR and assign it to myself to start searching the code for those 1024's that are fixed (and see if they can be reduced significantly), but more important, the ones that are variable (and make them vary as necessary).

src/signalk/signalk_delta.cpp
```diff
@@ -19,7 +19,18 @@ void SKDelta::append(const String val) {
...
 
 void SKDelta::get_delta(String& output) {
-  DynamicJsonDocument jsonDoc(1024);
+
+  // estimate the size of the serialized json string
+
+  // 70 required for initial boilerplate, 200 as extra buffer
+  int doc_size_estimate = 270;  
+  // pessimistic estimate 80 chars per path
+  doc_size_estimate += 80 * buffer.size();
```
joelkoz, line 28: This seems a little tight, don't you think?  I just looked at my latest device, and the sk path alone is 43 characters.  Granted, the 200 extra buffer probably takes care of this, but wouldn't a better estimate be the sum of the length of all the strings in the `buffer` list?  That would actually be extremely accurate, wouldn't it?

joelkoz (approved):
I would be OK as is, but I did comment on one quick change that may not require us to return to this code for a while.  Your call if you want to do it now or later.

src/signalk/signalk_delta.cpp
mairas: The json objects in `buffer` aren't serialized yet. It'd be possible to get the path strings but since the length of the actual value is highly dependent on the data type, I'm not sure that would improve the result too much. I'll just bump the estimate to 100 characters. I'd assume most of the paths to be quite a bit shorter than your example, so that hopefully would be quite enough. Does that sound reasonable?

ba58smith: Sounds reasonable to me. The longest path in any of my code is electrical.alternators.12V.voltage - 34 characters.

joelkoz: @mairas Are you sure about that?  Buffer is a list of String (`std::list<String> buffer;`).  If you look at the code of `sensesp_app.cpp`, line 148 does this for each `SKEmitter`:

```
      sk_source->attach([sk_source, this]() {
        this->sk_delta_->append(sk_source->as_signalk());
      });
```

and the signature of `SKDelta`'s append is `SKDelta::append(const String val)`, which makes sense - the only way to push something on a list of `String` is to convert it to a String.  So, by my reading, whenever an SKEmitter emits its value, the returned json is converted to a string and pushed onto `buffer` in an already serialized manner.

mairas: No, you're right of course; I misread the code. I didn't realize that ArduinoJson's `serialized()` allows you to insert preformatted chunks into the output. Since that's the case, your proposal is by far the best one. I'll update the PR later today.

mairas:
Ready for re-review.

ba58smith (approved):
I'll let you two decide on this one. I can barely follow the conversation, let alone provide any meaningful review of the code.

src/signalk/signalk_delta.cpp
```diff
@@ -19,7 +18,29 @@ void SKDelta::append(const String val) {
...
+      200;  // some extra headroom to hide embarrassing bugs
+
+  for (auto item : buffer) {
+    // also reserve space for the pre-rendered strings
+    doc_size_estimate += item.length() + 1;
+  }
+
+  if (!meta_sent_) {
+    // FIXME: how much do the metadata actually take
+    int num_metadata = SKEmitter::get_sources().size();
```
ba58smith, line 38: If you fixed this, remove the comment. If you didn't, either fix it, or make an Issue to go back and fix it later.

mairas: Ooo, that's strict! :-D Such a good thing that I rewrote that part and removed the comment...

mairas:
Ready for re-re-review.

src/signalk/signalk_delta.cpp
```diff
@@ -18,8 +17,62 @@ void SKDelta::append(const String val) {
...
+  }
+  return estimate;  
+}
+
+unsigned int SKDelta::get_metadata_size_estimate() {
+  int num_metadata = SKEmitter::get_sources().size();
+  int estimate = JSON_ARRAY_SIZE(num_metadata);
+
+  int num_fields;
+  auto update_estimate = [&](String& field) {
```
joelkoz, line 41: Nice use of the reference capture to make a nested function. I'll have to use this technique in the future!

joelkoz (approved):
Looks good to me!

src/signalk/signalk_delta.cpp
mairas: I couldn't decide whether this was super-hacky or a good way to reduce code repetition!

src/signalk/signalk_delta.cpp
ba58smith: Ha! That came across much less nice than I intended!

mairas: It was a fair comment in this case! Occasionally I might leave FIXME comments for less crucial things, and in those cases an issue might be a bit heavy-handed. But what you said would've been the right action to take!

Issue #316, opened 2020-12-12
Compiler warning about IntegratorT member initialization sequence


ba58smith:
```
Compiling .pio/build/esp32dev/lib272/SensESP/wiring_helpers.cpp.o
In file included from .pio/libdeps/esp32dev/SensESP/src/wiring_helpers.cpp:13:0:
.pio/libdeps/esp32dev/SensESP/src/transforms/integrator.h: In instantiation of 'IntegratorT<C, P>::IntegratorT(P, P, String) [with C = float; P = float]':
.pio/libdeps/esp32dev/SensESP/src/wiring_helpers.cpp:59:43:   required from here
.pio/libdeps/esp32dev/SensESP/src/transforms/integrator.h:71:5: warning: 'IntegratorT<float, float>::k' will be initialized after [-Wreorder]
   P k;
     ^
.pio/libdeps/esp32dev/SensESP/src/transforms/integrator.h:70:13: warning:   'float IntegratorT<float, float>::value' [-Wreorder]
   P value = 0;
             ^
.pio/libdeps/esp32dev/SensESP/src/transforms/integrator.h:33:3: warning:   when initialized here [-Wreorder]
   IntegratorT(P k = 1, P value = 0, String config_path = "")
```
I found the problem and will create the PR.

ba58smith:
Addresses Issue #317.

ba58smith:
Closing with PR #317.

PR #317, opened 2020-12-12
Fix compiler warning in Integrator

ba58smith:
Solves the compiler warning in Issue #316.

### DIFF
--- a/src/transforms/integrator.h
+++ b/src/transforms/integrator.h
@@ -67,8 +67,8 @@ class IntegratorT : public Transform<C, P> {
   virtual String get_config_schema() override { return FPSTR(INTEGRATOR_SCHEMA); }
 
  private:
-  P value = 0;
   P k;
+  P value = 0;
 };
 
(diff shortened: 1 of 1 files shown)


joelkoz (approved):
I THINK I follow this. ;)

Looks good.

ba58smith:
It's weird that it throws a compiler warning. In one case - a PR that someone else was working on - I think a simple change in the order of the member varibles - was the difference between it compiling and not!

Merging.

joelkoz:
If you look at our `platformio.ini`, we have this line:
```
build_unflags = -Werror=reorder
```

The "-Werror=reorder" means "turn the reorder warning into an error", but since its in the `build_unflags" section, that means "remove it from the build flags set."

If a user creates their own `platformio.ini` file and doesn't put that line in it, their version will not compile, as I think the default for this option is to consider this an error.

ba58smith:
Makes sense, but I never understood that `-Werror=reorder` until you just explained it. Good to know, as it'll surely come up again with another end user.

PR #318, opened 2020-12-12
New example: chain counter with LambdaConsumer

ba58smith:


examples/chain_counter.cpp
```diff
@@ -0,0 +1,121 @@
...
+int read_delay = 10;
+String read_delay_config_path = "/button_watcher/read_delay";
+auto* button_watcher = new DigitalInputChange(BUTTON_PIN, INPUT, CHANGE, read_delay, read_delay_config_path); 
+
+
+/* * Create a DebounceInt to make sure we get a nice, clean signal from the button.
+ * Set the debounce delay period to 15 ms, which can be configured at debounce_config_path
+ * in the Config UI.
+*/
+int debounce_delay = 15;
```
joelkoz, line 89: Is this the dead horse you were talking about?  Shouldn't this really be _at minimum_ two times the read_delay? I would think this example would make more sense with the debounce delay being 30.  That is 3 times the read delay.  I'm pretty sure that in this specific example, as configured, the debounce will have zero effect.

examples/chain_counter.cpp
```diff
@@ -0,0 +1,121 @@
...
+auto reset_function = [accumulator](int input) {
+  if (input == 1) {
+    accumulator->reset();  // Resets the output to 0.0
+  }
+};
+
+/* Create the LambdaConsumer that calls reset_function, Because DigitalInputChange
+ * outputs an int, the version of LambdaConsumer we need is LambdaConsumer<int>. 
+*/
+auto* button_consumer = new LambdaConsumer<int>(reset_function);
```
joelkoz, line 109: Question for the group...  I understand that by making `reset_function` a stand alone named lambda, and then passing it in as a parameter makes it explicitly clear what is going on.  On the other hand, since these examples are used as design patterns by many users, should we use the "anonymous function" syntax instead (i.e. place the lambda directly as a parameter to the constructor of this `LambdaConsumer`?  That is, should we promote the use of anonymous lambdas as parameters as a "best practice", or just leave this up to the individual on how to use?

The only downside of this explicit approach is that its more typing, and it requires use of a local identifier.

One compromise would be to put the "anonymous lambda function" version in a comment: 
```
// As an alternative, you can define the lambda function directly in the constructor of the
// LambdaConsumer, like this. This approach is much more concise, but is harder to
// understand by many.  Bla bla bla...
```

joelkoz (approved):
Looks good to me.  I will not be offended if you ignore my suggestions.

examples/chain_counter.cpp
ba58smith: 1. Program starts and nothing is happening with the button. Its state is LOW.
2. The button is pushed, which is an interrupt CHANGE, so the pin is read and `triggered_` is set to `true;`.
3. Between 1 and 10 ms later, the `app.onRepeat(10)` will see that `triggered- = true` and will output the value of the most recent read of the pin. If the button is bouncy, there might have been several reads of the pin, and if it's still bouncing, the pin could be HIGH or LOW. Let's say it's LOW...
4. Debounce will get the LOW and set up `onDelay(15), and if it gets no more input for 15 ms, it will emit the LOW. 
5. If the LambdaConsumer gets the LOW in 15 ms, it will ignore it - it's looking for a HIGH. But it won't emit the LOW - see steps 6 and 7.
6. The button is still pushed and bouncing, but it finally settles in at HIGH, and when the next `app.onRepeat(10)` rolls around, it will be sent to Debounce.
7. Debounce will cancel the DelayReaction it set up 10 ms ago, and set up a new one, because `value_sent_` is `false`. (Nothing has been sent yet). In 15 ms, it will emit the HIGH, unless another LOW comes in first. But another LOW won't come in, because the button has stopped bouncing, and there will be no more input to Debounce until the button is released (say, 500 ms later, which will be dealt with as a whole new "transaction" by then.) So the HIGH will be emitted, and the LambdaConsumer will see it, and will fire off the `reset_function`.

Assume the first output, in step 3, was HIGH:
4. Debounce gets the HIGH and sets up a 15 ms `onDelay` before it does anything.
5. Same as before: the LambdaConsumer won't get anything.
6. The button finally settles in at HIGH, and `triggered_` is set to `true`, and when the next `app.onRepeat(10)` rolls around, it won't send the HIGH, because `output != last_output_` if FALSE - it was HIGH last time, and it's HIGH this time.
7. Debounce won't get any input, so it won't cancel the pending DelayReaction, and 5 ms later, it will emit the first HIGH it received - the one it queued up the first time. The LambdaConsumer will see the HIGH, and fire off the `reset_function`.

So, in any situation in which the total bouncy time of the button is less than 20 ms, I think we're golden. Since the average bounce time of a button is only 6.2 ms, we should be good. To be really safe, I could make the 10 for DigitalInputChange and the 15 for Debounce be 20 and 25, but is that necessary? I guess it probably doesn't matter to whatever is consuming the output of Debounce, so maybe I should. But I don't think, in any scenario, the Debounce delay needs to be more than 5 ms longer than the DigitalInputChange delay.

I've written this out in such detail to make sure I understand it... and I think the above is right. Right?

examples/chain_counter.cpp
ba58smith: I doubt that very many end users would understand the "anonymous function" syntax. I sure as heck didn't until I spent a LOT of time working on the LambdaConsumer transform, and writing this example. I kinda like the idea of putting the anonymous syntax in a comment, though. 

@mairas ?

examples/chain_counter.cpp
joelkoz: Yes, I think you are right.  I keep forgetting that many of the bounces are happening behind the scenes, and Debounce will reset itself.

FWIW, I wrote it out also and intentionally tried to come up with a scenario where it breaks. I wrote out four scenarios, and on the fourth one, I ALMOST got one. It looked like we were going to have a problem, but I think the algorithm won in the end. For breadcrumbs, here are the rules of the code, and my "almost" scenario relative to the initial button state change:

```
Rules:
A. DigitalInputChange only sends if its latest sample is different from the last
B. Debounce only queues a send event when value is different from its "stable value"
C. Debounce ALWAYS cancels any queued event when it receives ANY value
D. Debounce's "stable value" gets reset only when a queued event is finally sent

Scenario:
Button bounces for 21 milliseconds

Result:
1. Press and hold HIGH. Stable signal for both classes is "HIGH"

2. Let go. Button goes LOW and starts bouncing at time "n"

3. By chance, the DigitialInputChange repeat loop detects LOW, and 
    since it is different from its last (stable) change, it sends the LOW immediately

4. Debounce gets the LOW. Since it is different from it's "stable", it queues up sending 
     LOW "n+15" from now

5. DIC is still bouncing, but happens to be LOW at "n+10". It is not different since the 
     last send at "n", so nothing sent.

6. Debounce timer expires at "n+15", it sends LOW, and sets its stable to LOW

7. DIC still bouncing, and sees a HIGH at "n+20". HIGH is different from "n+10", 
    so it sends HIGH

8. Debounce receives HIGH. Nothing to cancel, but since it is different from "stable", 
     it queues up HIGH for "n+20+15"

9. DIC settles at LOW. At "n+30", LOW is different than "n+20", so it sends LOW

10. Debounce cancels send. Since LOW is not different than stable, it does nothing.

11. No further change from DIC, and send event from 6 ends up being correct
```

examples/chain_counter.cpp
mairas: I think as an example Butch's approach works better because it's more stepwise and fine-grained commenting is easier. If the developer is comfortable with the code and the concepts, they can quite easily do the anonymous function. I'd probably have done that myself without even thinking too much but I think this approach is clearer.

examples/chain_counter.cpp
```diff
@@ -0,0 +1,121 @@
...
+ * counter_config_path.
+ */
+uint counter_read_delay = 1000;
+String counter_config_path = "/chain_counter/read_delay";
+auto *chain_counter = new DigitalInputCounter(COUNTER_PIN, INPUT_PULLUP, RISING,
+                          counter_read_delay, counter_config_path);
+
+/* An IntegratorT<int, float> called "accumulator" adds up all the counts it
+* receives (which are ints) and multiplies each count by gypsy_circum, which is the
+* amount of chain, in feet, that is moved by each revolution of the windlass. (Since
```
mairas, line 55: According to SK conventions, internally we probably should deal with SI units and then do conversions in the UI. So, I believe this should be in meters.

examples/chain_counter.cpp
```diff
@@ -0,0 +1,121 @@
...
+auto* accumulator = new IntegratorT<int, float>(gypsy_circum, 0.0, accum_config_path);
+
+/* 
+ * chain_counter is connected to accumulator, which is connected to an SKOutputNumber,
+ * which sends the final result to the indicated path on the Signal K server. (Note that
+ * each data type has its own version of SKOutput: SKOutputNumber for floats, SKOutputInt,
+ * SKOutputBool, and SKOutputString.)
+ */
+chain_counter->connect_to(accumulator) 
+             ->connect_to(new SKOutputNumber("chain_counter.path"));
```
mairas, line 70: I believe the path should be `navigation.anchor.something`, maybe `navigation.anchor.chainLength` would do?

mairas (commented):
A great example! A couple of small comments.

examples/chain_counter.cpp
joelkoz: Good point. Maybe this example would benefit from metadata being defined to specify the units. SK specs say all numeric values should have metadata, and now that we can do it...

ba58smith: That should have been obvious (meters). I'll fix it.

@joelkoz, I'll look at the metadata addition. If there's already a path in the spec for the amount of anchor rode deployed, I won't - but to be honest, I didn't even look. :-(

examples/chain_counter.cpp
ba58smith: I got lazy and didn't even look for it in the spec. I will, and if it's not there, I'll add the metadata for whatever I use in the example.

examples/chain_counter.cpp
mairas: I didn't see one but I might've missed it. `navigation.anchor` was dealing with deployed anchor but didn't find the rode length there.

examples/chain_counter.cpp
ba58smith: I added the anonymous approach to the comments.

examples/chain_counter.cpp
ba58smith: I changed it to `navigation.anchor.rodeDeployed` and added the metadata.

examples/chain_counter.cpp
ba58smith: I made it `navigation.anchor.rodeDeployed`, to be more generic than `navigation.anchor.chainDeployed`.

ba58smith:
I believe this is now ready for final review. All review comments have been addressed.

mairas:
Thumbs up!

PR #323, opened 2020-12-15
Add metadata in chain_counter example

ba58smith:


examples/chain_counter.cpp
```diff
@@ -62,15 +62,30 @@ float gypsy_circum = 0.32;
...
  * each data type has its own version of SKOutput: SKOutputNumber for floats, SKOutputInt,
  * SKOutputBool, and SKOutputString.)
  */
-chain_counter->connect_to(accumulator) 
-             ->connect_to(new SKOutputNumber("navigation.anchor.rodeDeployed"));
+  String sk_path = "navigation.anchor.rodeDeployed";
+  String sk_path_config_path = "/rodeDeployed/sk";
+  
+  chain_counter->connect_to(accumulator)
+               ->connect_to(new SKOutputNumber(sk_path, sk_path_config_path, metadata));
```
joelkoz, line 88: This gives me an idea: we could change SKOutputNumber from a `typedef` to an actual class that has multiple constructors.  One of those versions could take a `String` as a third parameter, which would be `units`, and the implementation would be to wrap that with a `SKMetadata` for units.  The current "3 parameter version", which takes `SKMetadata` as a third parameter but defaults to NULL for backward compatibility, could generate a `debugW()` warning message if not metadata is present.  Basically, the spec says that all numeric data should have at least a `units` specified. This would help introduce users to the conecpt.

joelkoz (approved):
Looks good.  One comment not related specifically to this PR, but to spark a conversation.

examples/chain_counter.cpp
joelkoz: Another idea for SensESP v2.0: we add `sk_path` as a member of `SKMetadata`, and have a version of `SKOutput` that takes that as its _ONLY_ parameter.  We then start moving toward specifying `sk_path` not with a String, but with `SKMetadata`.  We could even deprecate all versions of constructors that take `sk_path` as a `String`.  What this would do is strongly type the `sk_path` as a complex Signal K definition, and not just any old `String`

mairas: I'm all for stronger typing, and one way or another, providing the units could be made mandatory. How many types of units might we have and is it likely there would be new ones not yet present in the SK specification? What I'm thinking is could the units be expressed as an enum?

Would that enhanced `SKMetadata` have any other uses than conveying the parameters to SKOutput?

joelkoz: I was thinking about using an enumeration in the initial implementation of the `SKMetadata` units also, but decided against it for two reasons:

1. It would take up memory, as we would have to store the string equivalents of ALL of the unit types so we can render them correctly when sending the signal k
2. Not all of the units someone may possibly want are defined. For example, I think for time strings, I've seen a `units` field mentioning the RFC for ISO time strings in the wild. I am assuming that means other units for string (like "ON/OFF" or something like that) may someday have some non SI value.

In the end, keeping the units field as a `String` gave us maximum flexibility and kept our memory footprint down.

As far as other units - I am thinking "possibly?".  For example, the `SKListener` class which listens for values could possibly receive initial metadata.  By specifying a class instance in the `SKListener` constructor and holding a reference to it for example, we'd have a container to hold the values the server reported. Either that, or we could create a new `SKMetaListener` which would be a `ValueProducer<SKMetadata>`

ba58smith: I like the idea of moving the path into SKMetadata, but consider what that would do to the ability to configure the path in the Config UI. If you're going to allow the path to be configurable, you'd need a second String parameter - the config_path for the path. (But as I noted in your PR with the new SKOutputNumeric - I personally don't ever WANT a config_path for my sk_path - it's just not something I'm going to change on the fly.)

ba58smith:
Merging.

PR #324, opened 2020-12-15
Easy units and meta warnings for numeric SK outputs

joelkoz:
This PR adds a warning when outputting numeric values without metadata.  Also, an additional (optional) constructor parameter has been added to allow a quick specification of the "units" field by using a String as the third parameter.

This was from an idea sparked from discussions in #323

### DIFF
--- a/examples/ads1x15_volt_meter.cpp
+++ b/examples/ads1x15_volt_meter.cpp
@@ -82,7 +82,7 @@ ReactESP app([]() {
 
   tank_level->connect_to(new Linear(multiplier, offset, "/tank_level/linear"))
       ->connect_to(new SKOutputNumber("tanks.freshWater.currentLevel",
-                                      "/tank_level/skPath"));
+                                      "/tank_level/skPath", "ratio"));
 
   sensesp_app->enable();
 });
--- a/src/signalk/signalk_output.cpp
+++ b/src/signalk/signalk_output.cpp
@@ -1,0 +1,13 @@
+#include "signalk_output.h"
+
+template <typename T>
+SKOutputNumeric<T>::SKOutputNumeric(String sk_path, String config_path, SKMetadata* meta) :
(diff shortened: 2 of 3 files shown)


mairas (approved):
Simplifies things. I like it.

ba58smith (approved):
Looks good to me. Approved.

On a related note: I don't ever make my SK path strings configurable (i.e., I don't provide a config_path for them, because I know the path is the right one, and I'm not going to change it). How about another constructor with just (String sk_path, SKMetadata* meta), so I don't have to make up a config_path I'm never going to use? Of course, that can't work with your new (String sk_path, String sk_path_config_path, String units), because you'd have two constructors with (String, String). Maybe you have a thought? Or maybe it's not worth messing with.

joelkoz:
> On a related note: I don't ever make my SK path strings configurable (i.e., I don't provide a config_path for them, because I know the path is the right one, and I'm not going to change it). How about another constructor with just (String sk_path, SKMetadata* meta), so I don't have to make up a config_path I'm never going to use? Of course, that can't work with your new (String sk_path, String sk_path_config_path, String units), because you'd have two constructors with (String, String). Maybe you have a thought? Or maybe it's not worth messing with.

Yes, that would be easy to do.  Actually would probably be good to add to `SKOutput` in general. You are correct, however, that it can't be done with SKOutputNumeric's "Use a String as a shortcut parameter" as that would be ambiguous.

joelkoz:
@ba58smith `SKMetadata` was actually designed with the parameters to the constructor to be in "Most likely to be used" order.  So, even though I couldn't create a two parameter shortcut that omits the config path, you could now do:

`SKOutputNumeric(sk_path, new SKMetadata("K"))`

though that is actually 12 more characters than

`SKOutputNumeric(sk_path, "", "")`

:-P

ba58smith:
All looks good to me! Thx for the extra constructors - I'll start using them.

PR #328, opened 2020-12-16
SKMetadata timeout is a float

joelkoz:
This field should be a `float`.  Our default value of -1 means the timeout is not specified (so it won't be explicitly sent, and will use the server's default).


ba58smith:
Be sure to search the examples for the String version.

joelkoz:
I did. Timeout field was not used in any of the examples

mairas:
LGTM. I'll merge right away to get my PR rebased on top of this.

PR #332, opened 2020-12-17
Various files: Fix or add Doxygen comments; minor clean-up

ba58smith:
Address Doxygen issues noted when Version 1.0.0 was released.

Other MINOR fixes in code noticed while fixing Doxygen.

This addresses these issues in several files. I still have a lot more to go, but I don't want to have too many in one PR, so this is it for now. I'll keep going in other PR's when I have time.


joelkoz (approved):
Looks good. Nothing out of the ordinary

PR #333, opened 2020-12-17
Doxygen, fix data member names, remove "value" from Config UI for many Sensors

ba58smith:
Going through all the Sensor files to make sure they:
- Have Doxygen-style comments
- Have data member names with trailing "_"
- Don't have "value" as a read-only entry in the Config UI

~~This is enough for one PR. I'll keep going in other PR's.~~
Oh, what the heck. I'll do another one or two in this PR.

OK, NOW this is ready for review. Lots of commits, all of them very small and "focused" - should be easy to review.

### DIFF
--- a/src/sensors/bme280.cpp
+++ b/src/sensors/bme280.cpp
@@ -1,13 +1,12 @@
 #include "bme280.h"
 
 #include "sensesp.h"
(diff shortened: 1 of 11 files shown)


joelkoz (approved):
Looks good to me

PR #335, opened 2020-12-19
Part 2: Doxygen; data member names; remove "value" from Config UI; (for many Sensors) 

ba58smith:
Continuing the cleanup:
- Sensor base class
- SHT31
- SystemInfo
- Ultrasonic

Ready to review. (Just 3 more sensors to go!)

### DIFF
--- a/src/sensors/sht31.h
+++ b/src/sensors/sht31.h
@@ -24,9 +24,6 @@ class SHT31 : public Sensor {
  public:
   SHT31(uint8_t addr = 0x44);
   Adafruit_SHT31* adafruit_sht31;
-
- private:
-  uint8_t addr;
 };
 
 
--- a/src/sensors/sht31.h
+++ b/src/sensors/sht31.h
@@ -22,7 +22,7 @@
 
 class SHT31 : public Sensor {
  public:
-  SHT31(uint8_t addr = 0x44, String config_path = "");
+  SHT31(uint8_t addr = 0x44);
   Adafruit_SHT31* adafruit_sht31;
 
  private:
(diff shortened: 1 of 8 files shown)


src/sensors/sht31.h
```diff
@@ -6,35 +6,49 @@
...
+**/
 
 class SHT31 : public Sensor {
  public:
-  SHT31(uint8_t addr = 0x44, String config_path = "");
-  Adafruit_SHT31* adafruit_sht31;
-
- private:
-  uint8_t addr;
+  SHT31(uint8_t addr = 0x44);
```
joelkoz, line 25: Are you 100% sure there is no use case where someone would want to create a descendant class of SHT31?  If they did, and if they wanted to have THAT class be configurable, there is no way to pass the configuration path down to chain to the base classes.

joelkoz (commented):
One comment about removing of config_path

joelkoz (approved):
Actually - I studied this code a little more, and I now understand that this design pattern is used elsewhere, and the class I had a question about is not really a configurable SendESP "thing" used directly, but is instead used by the other "Value" class.  It threw me that it inherited from "Sensor" yet it's not really meant to be observed directly.

ba58smith:
Yeah, the constructor is really just a kind of wrapper for the Adafruit_SHT31 class, so it can be used like many other sensors. Actually, most of them have nothing configurable, so maybe none of them need to inherit from Sensor, do they? Should I go back and make them not descend from anything? (In a separate PR, of course.)

src/sensors/sht31.h
ba58smith: See your other comment about this - you resolved it there.

PR #336, opened 2020-12-19
Prevent duplicate PUT requests

joelkoz:
Fixes the strange timing problem when a PUT request is issued by a remote switch, but the the physical switch controller sends its current status before it answers the PUT request. This caused the remote switch to issue a second PUT request as it appeared the state had changed yet again.

This fix will also prevent future issues with the server being overwhelmed with requests as this only lets any one PUT request instance to make a single request at a time.

### DIFF
--- a/src/controllers/smart_switch_controller.h
+++ b/src/controllers/smart_switch_controller.h
@@ -34,6 +34,14 @@ class SmartSwitchController : public BooleanTransform,
                               public ValueConsumer<String> {
 
      public:
+       /**
+        * The constructor
+        * @param auto_initialize If TRUE, the controller will emit an
+        *  initial "off" status when enabled. This is generally the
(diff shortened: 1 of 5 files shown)


src/controllers/smart_switch_controller.h
```diff
@@ -34,13 +34,22 @@ class SmartSwitchController : public BooleanTransform,
                               public ValueConsumer<String> {
 
      public:
+       /**
```
ba58smith, line 37: In the big comment above - the @brief tag will cause everything to the next blank line to be included in the "brief" comment. If you want only the first sentence to be the "brief", put a blank line after it.

ba58smith (approved):
One comment about a comment. Code changes look good.

mairas (approved):
Looks good!

PR #343, opened 2020-12-24
Use an IP address for all server addresses in /examples

ba58smith:
Recently had a new user who didn't recognize `sk-server.local` as an alias for the SK Server's IP address, and I spent a couple hours helping him track it down. I'm assuming most users will recognize an IP address as something they need to change, like their wifi SSID and password.

### DIFF
--- a/examples/hysteresis.cpp
+++ b/examples/hysteresis.cpp
@@ -10,7 +10,7 @@ ReactESP app([]() {
   SetupSerialDebug(115200);
 
   sensesp_app = new SensESPApp("sensesp-hysteresis-example", "My WiFi SSID",
-                               "my_wifi_password", "skserver.lan", 80);
+                               "my_wifi_password", "192.168.10.3", 80);
 
   const char* sk_path = "environment.indoor.illuminance";
   const char* analog_in_config_path = "/indoor_illuminance/analog_in";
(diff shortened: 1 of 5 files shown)


mairas (approved):

JohnySeven:
I'm maybe late for show, but what if we add something like this into samples?

```
//#define SERVER_ADDRESS "xx.xx.xx.xx"
#define SERVER_PORT 3000
//#define WIFI_SSID "your_wifi"
//#define WIFI_PASSWORD "your_password
#ifndef SERVER_ADDRESS
#error Please define server address - uncomment line with SERVER_ADDRESS and specify your server address
#endif
#ifndef WIFI_SSID
#error Please define Wifi SSID - uncomment line with WIFI_SSID and specify your WIFI SSID
#endif
#ifndef WIFI_PASSWORD
#error Please define Wifi password - uncomment line with WIFI_PASSWORD and specify your WIFI password
#endif
```
And add defines into builder parameters?

ba58smith:
> I'm maybe late for show, but what if we add something like this into samples?
> 
> ```
> //#define SERVER_ADDRESS "xx.xx.xx.xx"
> #define SERVER_PORT 3000
> //#define WIFI_SSID "your_wifi"
> //#define WIFI_PASSWORD "your_password
> 
> And add defines into builder parameters?

Overkill? And it would make it more complicated to explain how to NOT provide those parameters,  so that WifiManager and mDNS can do their jobs.

PR #344, opened 2020-12-24
Clarify which SensESP to use in platformio.ini

ba58smith:
Add `https://github.com/SignalK/SensESP` to the `lib_deps = ` section of `examples/platformio.ini`, as an option to `SignalK/SensESP`, because the latter doesn't have any of the PR's merged since the last release.

This is the result of spending a lot of time helping figure out why new users are getting errors - because they're using new examples from `master`, with code from `latest` (the current "Relase" version).

Hopefully, this will cut down on support issues, or at least make it easier to tell someone how to solve the "using new examples" problem.

### DIFF
--- a/examples/platformio.ini
+++ b/examples/platformio.ini
@@ -16,7 +16,7 @@
 [platformio]
 ;set default_envs to whichever board(s) you use. Build/Run/etc processes those envs
 default_envs = 
-   ;esp32dev
+;   esp32dev
    d1_mini
 ;   esp-wrover-kit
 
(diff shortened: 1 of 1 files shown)


examples/platformio.ini
```diff
@@ -26,8 +26,16 @@ framework = arduino
 lib_ldf_mode = deep
 monitor_speed = 115200
 lib_deps =
+; Comment out one of the following two paths that point to the
+; SensESP library code.
+
+; This one is the "Release" version - the branch called "latest".
   SignalK/SensESP
 
+; This one has all merged PR's, but is not yet an official "release" version.
```
mairas, line 35: It really isn't a good practice to encourage end users to use a development version of any software. If you insist on adding this option, could you please add a comment similar to the following: "This version is unstable and under continuous development and will break without warning. Use it only if you want to participate in the development or need to test some yet unreleased feature."

ba58smith: Done.

ba58smith: And to be clear, I'm not encouraging end users to use this version - but it IS the version that they see in the repo. People end up trying the examples in `master`, then they have trouble, and the solution is often for them to use all the code in `master` rather than `latest`.

PR #345, opened 2020-12-24
Use elapsedMillis for SystemHz

ba58smith:


mairas (approved):
The practical difference is small, but why not. I like `elapsedMillis` too. :-)

Not in scope of this PR, but the timer could also be changed to microseconds. That would slightly (although probably not significantly, by any means) improve the accuracy of SystemHz.

ba58smith:
@mairas , I did it because I like consistency. Since ElapsedMillis is now available in SensESP, it seems like we should use it, as it's cleaner, and less prone to someone doing the math wrong when doing it "the old way".

PR #347, opened 2020-12-25
MAX31856 enhancements

ba58smith:
- Change the conversion mode to ONESHOT.
- Add a second constructor, per @mairas request.
- Remove potential endless loop with begin().
- Add a constructor without the obsolete data ready pin.
- Update the example.

### DIFF
--- a/src/sensors/max31856_thermocouple.cpp
+++ b/src/sensors/max31856_thermocouple.cpp
@@ -14,6 +14,7 @@ MAX31856Thermocouple::MAX31856Thermocouple(int8_t cs_pin, int8_t mosi_pin,
   load_configuration();      
   max31856_ = new Adafruit_MAX31856(cs_pin, mosi_pin, miso_pin, clk_pin);
   if (!max31856_->begin()) {
+    sensor_detected_ = false;
     debugW("No MAX31856 detected: check wiring.");
   }
   max31856_->setThermocoupleType(tc_type);
@@ -31,6 +32,7 @@ MAX31856Thermocouple::MAX31856Thermocouple(Adafruit_MAX31856* max31856,
 
 void MAX31856Thermocouple::enable() {
   // Must be at least 500 to allow time for temperature "conversion".
+  if (sensor_detected_) {
   if (read_delay_ < 500) {
     read_delay_ = 500;
   }
@@ -41,6 +43,8 @@ void MAX31856Thermocouple::enable() {
       this->emit(temp);
       });
   });
+  }
+  else debugE("MAX31856 not enabled: no sensor detected.");
 }
 
 void MAX31856Thermocouple::get_configuration(JsonObject& root) {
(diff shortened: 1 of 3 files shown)


src/sensors/max31856_thermocouple.cpp
```diff
@@ -11,22 +11,35 @@ MAX31856Thermocouple::MAX31856Thermocouple(int8_t cs_pin, int8_t mosi_pin,
     : NumericSensor(config_path),
       data_ready_pin_{drdy_pin},
       read_delay_{read_delay} {
-  load_configuration();
+  load_configuration();      
   max31856_ = new Adafruit_MAX31856(cs_pin, mosi_pin, miso_pin, clk_pin);
   if (!max31856_->begin()) {
-    while (1) delay(10);
+    debugW("No MAX31856 detected: check wiring.");
```
mairas, line 17: It seems to me that if the physical sensor is not detected, the measurement loop in `enable()` is started anyway. I wonder what `readThermocoupleTemperature()` returns if the device is not present? My preference would be to only start the repeat loop if the `begin()` call was successful.

mairas (commented):
Very nice. One small comment/request.

src/sensors/max31856_thermocouple.cpp
ba58smith: Done. Look OK?

ba58smith:
@BjarneBitscrambler, I thought about your suggestion to allow the sensor to "recover", but I couldn't think of any scenario in which it would actually do what it's intended to. The actual MAX31856 is almost certainly going to be soldered in place, or at least solidly plugged into a socket, and should be protected from accidental physical disturbances that could unseat it. If it comes unseated, it will have to be physically reseated, and at that point, it would make sense to just push the RESET button on the ESP, which starts the code all over again. IOW, I couldn't think of a situation in which `begin()` would fail that can be fixed easily enough to try to make the code handle it.

BjarneBitscrambler:
@ba58smith - thinking it through as you describe, it does sound unlikely a mechanical/physical failure would occur that could be corrected by just re-initializing. So your present approach seems appropriate.

At the risk of letting more worms loose, I noticed in the MAX31856 spec sheet that it can detect various faults, like broken thermocouple wires. Any thoughts on detecting that and outputting a `null` instead of a number?  I'll totally agree if you say it goes quite a bit beyond basic sensor functionality...

ba58smith:
@BjarneBitscrambler, I looked at the Datasheet, and see the open circuit and over/under voltage detection. They both look doable, but I'm going to say either or both of those could be done as an enhancement, in a new PR, later.

BTW, after a discussion w/ @mairas, it was determined that, at least for now, invalid sensor output should simply stop, rather than sending `null`, with an appropriate debugE message indicating why output is not happening. That's what I've done in this PR.

src/sensors/max31856_thermocouple.cpp
mairas: Yup!

mairas:
Looks good to me now!

PR #348, opened 2020-12-26
FIX: Many "sensor" classes should not inherit from Sensor

ba58smith:
This goes back to the first sensor code that I wrote. I mistakenly made the class that represents the physical sensor inherit from `Sensor`. It doesn't need to, because it doesn't have anything that's configurable, and it doesn't actually emit anything.
- ADS1x15
- BME280
- BMP280
- SHT31
- 9DOF

### DIFF
--- a/src/sensors/ads1x15.cpp
+++ b/src/sensors/ads1x15.cpp
@@ -3,8 +3,7 @@
 #include "sensesp.h"
 
 template <class T_Ada_1x15>
-ADS1x15<T_Ada_1x15>::ADS1x15(uint8_t addr, adsGain_t gain)
-    : Sensor(), gain_{gain} {
+ADS1x15<T_Ada_1x15>::ADS1x15(uint8_t addr, adsGain_t gain) : gain_{gain} {
   ads_ = new T_Ada_1x15(addr);
   ads_->begin();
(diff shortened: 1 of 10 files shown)


BjarneBitscrambler:
I had been looking at that same issue relating to the orientation sensor, and I concluded that it might be useful to have configurable properties so one can switch things like the I2C pins using the web interface. Is that a reasonable justification for keeping the Sensor inheritance, or is there a better approach?

mairas:
> I had been looking at that same issue relating to the orientation sensor, and I concluded that it might be useful to have configurable properties so one can switch things like the I2C pins using the web interface. Is that a reasonable justification for keeping the Sensor inheritance, or is there a better approach?

If you only want to be able to configure something, you should only inherit `Configurable`.

mairas (approved):

ba58smith:
@BjarneBitscrambler - I think that if you change the I2C pins at run-time, the changes won't take effect until you restart the ESP. Aside from that, I think it should work. (Making it inherit from `Configurable`, rather than `Sensor`, as @mairas said.)

ba58smith:
@BjarneBitscrambler - if you want to do that, please wait until after I merge this PR before you submit a PR that does it.

PR #353, opened 2021-01-01
Add instructions to disable SSL in the README

ba58smith:
To prevent more "support calls" from someone enabling SSL when they install Signal K Server.


ba58smith:
Minor changes, and only to the README. Merging.

PR #354, opened 2021-01-03
Fix outdated comment in Transform.h

ba58smith:
@param `id` was replaced by `config_path` a long time ago.


ba58smith:
Trivial change, and only to a comment. Merging.

PR #357, opened 2021-01-07
Add pictures

norbert-walter:
I add several pictures for a better understanding


ba58smith:
Since these are from the Fritzing file (which at least some people can't open), I'll add them.

PR #359, opened 2021-01-13
Allow compilation when DEBUG_DISABLED defined

### LINKED ISSUE #358: Fixes to allow RemoteDebug messages be turned off
### Problem
We can't turn off the  `debugI()`, `debugE()`, etc messages because of a missing `#ifndef ... #endif` guard in `sensesp_app.cpp`

### Details
The RemoteDebug library (used by SensESP to provide the `debugI()`, `debugE()`, etc message functions) has the ability to compile with the library's functions turned off. This is useful for conserving RAM.  From `RemoteDebug.cpp`:
```
///// Debug disable for compile to production/release
///// as nothing of RemotedDebug is compiled, zero overhead :-)
#define DEBUG_DISABLED true // Uncomment if the IDE did not recognize, to force it

#ifndef DEBUG_DISABLED
[...]
```
`sensesp_app.cpp` tests `DEBUG_DISABLED` when creating the `RemoteDebug` object, but just following that test there is a missing test in `SetupSerialDebug()`  From `sensesp_app.cpp`:
```
[... to line 22]
#ifndef DEBUG_DISABLED
RemoteDebug Debug;
#endif

void SetupSerialDebug(uint32_t baudrate) {
  Serial.begin(baudrate);

  // A small delay and one debugI() are required so that
  // the serial output displays everything
  delay(100);
  Debug.setSerialEnabled(true);                   <<<<<< Missing #ifndef...#endif
  delay(100);
  debugI("\nSerial debug enabled");
}
```
Correcting this uncovers a second problem. `RemoteDebug.h` placed its `#ifndef DEBUG_DISABLED` around several `#include` statements. When these are no longer included, then compilation fails because `sensors/system_info.cpp` no longer has access to the `WiFi` object.  The fix for that is to insert the following block in `sensors/system_info.cpp`:
```
#if defined(ESP8266)
  #include <ESP8266WiFi.h>
#elif defined(ESP32)
  #include <WiFi.h>
#endif
``` 

### Fix
I'll open a PR shortly.

BjarneBitscrambler:
Basically, we are unable to turn off the debug*() informative prints due to two small compilation errors. Corrected by including one #ifndef guard and one #include statement.  See #358 for details.

This resolves #358.

### DIFF
--- a/src/sensesp_app.cpp
+++ b/src/sensesp_app.cpp
@@ -29,9 +29,11 @@ void SetupSerialDebug(uint32_t baudrate) {
 
   // A small delay and one debugI() are required so that
   // the serial output displays everything
+#ifndef DEBUG_DISABLED
   delay(100);
   Debug.setSerialEnabled(true);
   delay(100);
+#endif 
   debugI("\nSerial debug enabled");
 }
 
--- a/src/sensors/system_info.cpp
+++ b/src/sensors/system_info.cpp
@@ -1,6 +1,11 @@
 #include "system_info.h"
 
 #include "Arduino.h"
+#if defined(ESP8266)
+  #include <ESP8266WiFi.h>
+#elif defined(ESP32)
+  #include <WiFi.h>
+#endif
 #include "sensesp.h"
 
 void SystemHz::tick() { tick_count_++; }


BjarneBitscrambler:
### Testing Performed
The changes compile and run on both ESP8266 (d1_mini) and ESP32 (WROVER).  

There are approx 8900 more bytes reported by freemem when running the smartswitch example on the d1_mini with DEBUG_DISABLED  (freemem = 22680 vs. 13768). And, with DEBUG_DISABLED, all is quiet on the serial port when running SensESP as expected.

src/sensesp_app.cpp
```diff
@@ -29,9 +29,11 @@ void SetupSerialDebug(uint32_t baudrate) {
 
   // A small delay and one debugI() are required so that
   // the serial output displays everything
+#ifndef DEBUG_DISABLED
   delay(100);
   Debug.setSerialEnabled(true);
   delay(100);
+#endif 
   debugI("\nSerial debug enabled");
```
ba58smith, line 37: Shouldn't the debugI be above the #endif?

ba58smith (changes requested):

src/sensors/system_info.cpp
```diff
@@ -1,6 +1,11 @@
 #include "system_info.h"
 
 #include "Arduino.h"
+#if defined(ESP8266)
+  #include <ESP8266WiFi.h>
+#elif defined(ESP32)
+  #include <WiFi.h>
+#endif
 #include "sensesp.h"
```
ba58smith, line 10: I'm not at my computer so I can't verify this: your change here fixes the situation if you disable debug. Does it cause a problem if you DON'T disable debug?

ba58smith (changes requested):

src/sensors/system_info.cpp
BjarneBitscrambler: None that I've noticed.  Everything compiles, as evidenced by the auto-build checks (which I believe would be using the normal debug-enabled configuration). 
When I turn off debug msgs in RemoteDebug.*, I still see the expected info in the sensor's web interface, but the terminal window has zero traffic as it should. When debug msgs aren't disabled, the expected Signal K messages display in the terminal window.

src/sensesp_app.cpp
BjarneBitscrambler: It could be, but it doesn't make any difference. The RemoteDebug library turns **all** debug_() calls into nothing (i.e. zero compiled code) when DEBUG_DISABLED is defined. So there's no need in any of the code to put #ifdef...#endif guards around the info messages.  That's why there's such a large savings in the code size.

joelkoz (approved):
These look OK to me.

src/sensesp_app.cpp
ba58smith: I hadn't thought of that (that if debug is disabled, the debug message in question won't print). Just looks odd. But no need to change it.

ba58smith (approved):

src/sensesp_app.cpp
BjarneBitscrambler: Great - thanks for reviewing!

ba58smith:
Closes #358 .

PR #360, opened 2021-01-14
fix: output SHT31 humidity as ratio, not percent

joabakk:
Spec compliance


mairas:
Thanks! Backwards incompatible, but adherence to the spec overrides that. Merging.

PR #361, opened 2021-01-14
Remove the delta size estimate debugging message

mairas:
When working with the Json doc size estimates, it seems I left behind an unnecessary and spammy debugging message.


ba58smith:
Thank you!

PR #363, opened 2021-01-19
ENH: SmartSwitchController double click to sync sk paths

joelkoz:
This PR enhances SmartSwitchController to allow the "double click" button input to sync the light's status to one or more additional sk paths via an SKBooleanPutRequest().  For example, a "double click" on one of the boat's interior lights can cause ALL the interior lights to turn on.

### DIFF
--- a/src/controllers/smart_switch_controller.cpp
+++ b/src/controllers/smart_switch_controller.cpp
@@ -1,13 +1,18 @@
 #include "smart_switch_controller.h"
 #include "transforms/truth_text.h"
 
-SmartSwitchController::SmartSwitchController(bool auto_initialize, String config_path, std::set<SyncPath>* defaults) :
+SmartSwitchController::SmartSwitchController(bool auto_initialize, String config_path, const char* sk_sync_paths[]) :
    BooleanTransform(config_path),
    auto_initialize_{auto_initialize} {
 
-  if (defaults != NULL) {
-    sync_paths.clear();
-    sync_paths = *defaults;
+  if (sk_sync_paths != NULL) {
+      sync_paths.clear();
+      int i = 0;
+      while (strlen(sk_sync_paths[i]) > 0) {
+          SyncPath path(sk_sync_paths[i]);
+          sync_paths.insert(path);
+          i++;
+      } // while
   }
 
   load_configuration();
--- a/src/controllers/smart_switch_controller.cpp
+++ b/src/controllers/smart_switch_controller.cpp
@@ -1,6 +1,19 @@
 #include "smart_switch_controller.h"
 #include "transforms/truth_text.h"
 
+SmartSwitchController::SmartSwitchController(bool auto_initialize, String config_path, std::set<SyncPath>* defaults) :
+   BooleanTransform(config_path),
+   auto_initialize_{auto_initialize} {
+
+  if (defaults != NULL) {
+    sync_paths.clear();
+    sync_paths = *defaults;
+  }
+
+  load_configuration();
+}
+
+
 void SmartSwitchController::enable() {
     if (auto_initialize_) {
        this->emit(is_on);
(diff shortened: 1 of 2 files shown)


ba58smith (approved):
Once again, I don't feel qualified to review this code - I can barely follow it. :-) You'll have to rely on @mairas.

src/controllers/smart_switch_controller.h
```diff
@@ -1,31 +1,39 @@
...
  * If the input is a ClickType, or a String that's NOT a "valid truth type", 
  * SmartSwitchController's output will simply toggle the output back and forth 
- * between "on" and "off" with each input. The one exception to this is if
- * the input is a ClickType with the value ClickTypes::UltraLongSingleClick
- * which will reboot the MCU.
+ * between "on" and "off" with each input. There are two exceptions to this if
+ * the input is a ClickType:
+ * <ol>
+ *   <li>A value of ClickTypes::DoubleClick will toggle the output as well as
+ *       send PUT requests to syncrhonize zero or more paths that are configured to
```
mairas, line 28: Nit: typo: syncrh...

src/controllers/smart_switch_controller.cpp
```diff
@@ -46,3 +68,46 @@ void SmartSwitchController::set_input(String new_value, uint8_t input_channel) {
     }
     this->emit(is_on);
 }
+
+
+void SmartSwitchController::get_configuration(JsonObject& root) {
+  JsonArray jPaths = root.createNestedArray("sync_paths");
```
mairas, line 74: Nice example of storing a data structure in configuration. Editable, too!

src/controllers/smart_switch_controller.h
```diff
@@ -34,22 +42,46 @@ class SmartSwitchController : public BooleanTransform,
                               public ValueConsumer<String> {
 
      public:
+       class SyncPath {
```
mairas, line 45: I think we already discussed nested classes (which aren't really part of my active vocabulary, for a reason or another) but I forgot what the conclusion was, so I hope I'm not repeating myself: since SyncPath is used as part of the public interface, I have a slight preference for it being a top-level class of its own. Ignore if you like it as is.

mairas (commented):
Neat. I'm wondering that given that this is a bit more complex use case, would an example be a good addition?

src/controllers/smart_switch_controller.h
joelkoz: This code technique mimics exactly CurveInterpolator, which is the transform I wrote many moons ago and is already part of the code set. It too has a nested class. In both of these examples, the class is used exclusively to hold complex configuration data, and is thus so tightly integrated into the class I think it makes sense to keep them nested. If the class served any value outside of this class, I think it would make sense to un-nest it.  My first inclination was actually to make it protected or private if not for the fact that I wanted to leave open the option for the programmer to hard code the paths in their app and pass it in to the constructor.  Actually, now that I said that, it makes more sense to make the constructor accept an array of Strings vs. the SyncPath class, and then make SyncPath protected.  It is not really intended for the developer to have access to the PUT request object nor work with the class directly.

BTW - CurveInterpolator is a better example of having an array of objects stored as configuration and editable the the Config UI. In the case of SyncPath, the configuration really only stores a String (and the config is an array of Strings).  The second member of SyncPath stores a runtime pointer to the PUT request object.  CurveInterpolator, on the other hand, actually stores an array of objects as its configuration.

src/controllers/smart_switch_controller.cpp
joelkoz: > Neat. I'm wondering that given that this is a bit more complex use case, would an example be a good addition?

I have a working example (sort of) I am doing for the Signal K blog post I am working on.  The code doesn't actually use this feature, but my deployment takes advantage of the configuration (which is part of the blog post).

src/controllers/smart_switch_controller.h
mairas: Fair enough, thanks for sharing the rationale!

joelkoz: Change made.  FYI, I had to keep the nested class public because `std:set<>`, which is used to store the set of SyncPath, needs to be able to "see" the `operator<()` method (it is a required definition to be able to add a class to a set.)

Issue #287, opened 2020-11-30
Consider changing class Transform to have Configurable and Enable be virtual


joelkoz:
In an attempt to make some of the other previous Transforms like Debounce use templates, an issue was discovered:  by putting the implementation of everything in the .cpp file, the compiler will not be capable of generating a Debounce<> class for any future data types (e.g. `Debounce<SomeFutureDataType>`, as it will be blind to the implementation. On the other hand, if you put the entire implementation in the header file (esp. the configuration stuff) the compiler will generate a TON of redundant code.  

The compromise is to make an interim base class and put the common code that is NOT type specific in and put that common code in the .cpp.   Then we'd put the "type specific code" in the header file.  This technique was used in SKPutRequest().  The problem with that code however is SKPutRequest() is a `ValueConsumer` and not a full blown `Transform`.  It is not possible to introduce `Transform` into the inheritance hierarchy once Configurable has already been used (as was done for SKPutRequestBase) UNLESS we define class `TransformBase`'s inheritance to be `virtual Configurable` and `virtual Enable`.

It is true that SKPutRequest() could have been defined as a `ValueConsumer<>` and also `ValueProducer<>` (which are the key components of Transform) and MOST of the functionality would be preserved.  However, the "chained connect_to()" calls would not work, as those only work with Transform (at the moment).  An alternative to changing `TransformBase` is to simply find a way to make "chained connect_to()" work with any class that is a ValueConsumer and ValueProducer, and not require Transform at all.

ba58smith:
Note for @joelkoz, from a post on Slack: https://stackoverflow.com/questions/46226021/c-function-parameter-that-extends-multiple-classes

PR #365, opened 2021-01-25
Debounce cleanups

mairas:
I did a bit of reworking on the networking logic and removed the "reboot on wifi disconnect" from the Networking class and made it a consumer of SystemState in SensESPApp. For this, I required Debounce to only trigger the reset if the state had been stable for 3 minutes.

First, I noticed that due to the Debounce template implementation, it was impossibly to instantiate Debounce<SystemStatus>. I did a quick benchmark to verify the claim in Issue #287 that "if you put the entire implementation in the header file (esp. the configuration stuff) the compiler will generate a TON of redundant code." (The issue sounded a bit peculiar because basically all of STL is, well, templates.)

I have a random SensESP test program that takes some changing AnalogInput, applies hysteresis and rounding and then Debounces the boolean and integer values. Compiled with the original implementation, the RAM footprint was 51.0 KB and flash footprint 1015.4 KB. I then refactored the code into a standard headers-only template class. The RAM footprint remained at 51.0 KB while the flash footprint dropped to 1011.0 KB. So, that is the first major change in this PR.

The second change relates to the Debouncing logic. I am debouncing the system state which gets regular WifiDisconnected updates (the same value all over again). The original logic reset the timer for every update, even if they repeated the same value. That didn't quite make sense to me - it would be impossible to debounce any polled values with this logic. I shuffled the code a bit around but the only change in the end was that repeated values are considered stable instead of resetting the timer every time.


src/transforms/debounce.h
```diff
@@ -42,26 +42,31 @@ class DebounceTemplate : public SymmetricTransform<T> {
...
-        this->value_sent_ = true;
+        this->debounced_value_ = input;
         this->emit(input);
       });
+      value_received_ = true;
     }
   }
 
  private:
+  bool require_change_;
```
ba58smith, line 66: It doesn't appear that `bool require_change_;` is used anywhere.

ba58smith (approved):
Looks good to me.

src/transforms/debounce.h
mairas: Ah! An accidental leftover. Will remove.

mairas: Fixed.

mairas:
Since this PR includes a change in the behavior logic, if @joelkoz has any doubt, it'd be great if you could test it before merging so that you're not left in the dark (pun absolutely intended!).

PR #366, opened 2021-01-25
WSClient cleanups

mairas:
Two minor cleanups: first, token validity is now tested on a URL that doesn't return a ton of data. Second, a bit of logic cleanup that should have no functional changes.

### DIFF
--- a/src/net/ws_client.cpp
+++ b/src/net/ws_client.cpp
@@ -257,60 +257,60 @@ void WSClient::connect() {
     return;
   }
 
-  if (WiFi.isConnected()) {
-    debugD("Initiating websocket connection with server...");
-
-    connection_state = WSConnectionState::kWSAuthorizing;
-
-    String server_address = this->server_address;
-    uint16_t server_port = this->server_port;
-
-    if (this->server_address.isEmpty()) {
-      if (!get_mdns_service(server_address, server_port)) {
-        debugE("No Signal K server found in network when using mDNS service!");
-      } else {
-        debugI("Signal K server has been found at address %s:%d by mDNS.",
-               server_address.c_str(), server_port);
-      }
-    }
-
-    if (!server_address.isEmpty() && server_port > 0) {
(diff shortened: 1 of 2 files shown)


src/net/ws_client.cpp
```diff
@@ -320,7 +320,7 @@ void WSClient::test_token(const String server_address,
   HTTPClient http;
 
   String url = String("http://") + server_address + ":" + server_port +
-               "/signalk/v1/api/";
+               "/signalk/";
```
ba58smith, line 323: THANK YOU!

ba58smith (approved):
This one's over my head. Sorry.

mairas:
> This one's over my head. Sorry.

It absolutely isn't! :-D The diff just looks messy but all I did was change:
```
if (foo) {
  a_lot_of_code_here;
} else {
 a_bit_of_code_here;
}
```

to 

```
if (!foo) {
  a_bit_of_code_here;
  return;
}
a_lot_of_code_here;
```

src/net/ws_client.cpp
```diff
@@ -257,60 +257,60 @@ void WSClient::connect() {
...
-    if (this->server_address.isEmpty()) {
-      if (!get_mdns_service(server_address, server_port)) {
-        debugE("No Signal K server found in network when using mDNS service!");
-      } else {
-        debugI("Signal K server has been found at address %s:%d by mDNS.",
-               server_address.c_str(), server_port);
-      }
-    }
+  String server_address = this->server_address;
+  uint16_t server_port = this->server_port;
```
ba58smith, line 272: This is an exellent example of where adding the trailing underscore to member variables can simplify things:
```
String server_address = server_address_;
  uint16_t server_port = server_port_;
  ```
  Voila! No more confusion about which server_address or sever_port we're talking about!
  
  Just an observation. :-)

mairas: Yeah, I've actually grown to like the underscores. I can add a cleanup commit to the PR to rename the remaining offending variables in WSClient...

PR #367, opened 2021-01-25
Non blocking wifi

mairas:
This PR implements quite a few changes in the networking code. WiFi connection logic is no longer blocking: we don't need to wait for the wifi to get connected before entering the main loop. This allows easier simultaneous NMEA 2000 and other protocol use.

The wifi disconnection watchdog that reboots the device after disconnection is now a LambdaConsumer of SystemStatus in SensESPApp instead of being hard-coded in the Networking class. Sure, the behavior is arguably still hard-coded but will be easier to modify in the future. This requires PR #365 to work.

Wifi state changes are now captured and propagated using event handlers. Due to underlying API differences, the implementation is platform-specific (different for ESP8266 and ESP32).

Wifi settings were not reset properly at least on ESP32. Now they're nuked to invalid values on reset to force the device to forget the old values.

Wifi settings are no longer saved on SPIFFS by SensESP. The settings are in any case saved to flash memory by WifiManager, so having them in multiple places just made things more difficult.

### DIFF
--- a/src/controllers/system_status_controller.cpp
+++ b/src/controllers/system_status_controller.cpp
@@ -6,16 +6,16 @@ void SystemStatusController::set_input(WifiState new_value,
   // this would be a simple array dereferencing
   switch (new_value) {
     case WifiState::kWifiNoAP:
-      this->emit(SystemStatus::kWifiNoAP);
+      this->update_state(SystemStatus::kWifiNoAP);
       break;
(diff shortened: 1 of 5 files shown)


mairas:
One of the tests fails until PR #365 has been merged.

ba58smith:
I've reviewed #365, and it looks OK to me. But that's all I can add to this PR - it's way over my head.

PR #370, opened 2021-02-06
Change min interval for sending deltas to 5 ms from 100 ms

### LINKED ISSUE #369: Increase max rate of SK updates
Currently, the max rate at which SK updates are sent to the server is 10 Hz.  This value is set in `net/ws_client.cpp`:
```
void WSClient::enable() {
  app.onDelay(0, [this]() { this->connect(); });
  app.onRepeat(20, [this]() { this->loop(); });
  app.onRepeat(100, [this]() { this->send_delta(); });          <<< throttled to max 10 Hz
  app.onRepeat(10000, [this]() { this->connect_loop(); });
}
```
If a fast sensor (e.g. GPS or compass) generates updates at a higher rate, they are stored and sent out in batches at 100 ms intervals. Here, for example, is a typical delta sent out at 100 ms intervals when the attitude and heading reports intervals are set to 50 ms:
```
(D) (get_delta) delta: {"updates":[{"source":{"label":"SensESP_D1"},"values":[
     {"path":"navigation.attitude","value":{"yaw":2.853591,"pitch":3.072033,"roll":1.368445}},
     {"path":"navigation.headingCompass","value":2.853591},
     {"path":"navigation.headingMagnetic","value":2.940857},
     {"path":"navigation.attitude","value":{"yaw":2.853448,"pitch":3.071935,"roll":1.368391}},
     {"path":"navigation.headingCompass","value":2.853448},
     {"path":"navigation.headingMagnetic","value":2.940714}
]}]}
```
Note the duplicate keys in the delta, due to queued updates.

While 10 Hz is a reasonable update rate for most parameters, responsiveness by controllers such as autopilots and radar plotters can benefit by faster rates. Bear in mind that there are multiple delay points in the signal chain - the cumulative latency between the instant that a parameter is generated and when it arrives at its destination could approach 150 ms or more.

The ESP8266 and ESP32 hardware supports faster update rates. I've been testing both platforms with the change to `app.onRepeat(1, [this]() { this->send_delta(); });`  (i.e. SensESP checks every 1 ms whether a delta is available, and sends it if so) and it has been functioning well. There is even a slight benefit to memory usage (related issue #239) in that deltas that are sent out when available rather than queuing them up will take less RAM.

I propose **changing the interval between `send_delta()` calls to, say, 5 ms.**  This would reduce the latency contribution due to SensESP into the same order of magnitude as that due to other delays, like WiFi router routing and the Signal K server running on a raspberry Pi.

Any reasons not to do this?  I can raise a PR if it sounds reasonable.

BjarneBitscrambler:
Current minimum interval between SK delta messages is 100 ms. This change improves the maximum possible rate of deltas by decreasing the interval between calls to `send_delata()` to 5ms.

This  resolves #369

### DIFF
--- a/src/net/ws_client.cpp
+++ b/src/net/ws_client.cpp
@@ -71,7 +71,7 @@ WSClient::WSClient(String config_path, SKDelta* sk_delta, String server_address,
 void WSClient::enable() {
   app.onDelay(0, [this]() { this->connect(); });
   app.onRepeat(20, [this]() { this->loop(); });
-  app.onRepeat(100, [this]() { this->send_delta(); });
+  app.onRepeat(5, [this]() { this->send_delta(); });
   app.onRepeat(10000, [this]() { this->connect_loop(); });
 }
 


mairas:
Thanks! Merging.

PR #371, opened 2021-02-07
AnalogInput: update docs for this and related classes

ba58smith:
Make it clear that AnalogVoltage is probably not needed in most cases.


ba58smith:
Since this is all just changes to comments, I'll go ahead and merge.

PR #373, opened 2021-02-11
Multiple sensor example

ba58smith:
Several people have asked on Slack about using more than one sensor in a single main.cpp. This should address that.

### DIFF
--- a/examples/multiple_sensors_example.cpp
+++ b/examples/multiple_sensors_example.cpp
@@ -1,0 +1,72 @@
+#include "sensors/analog_input.h"
+
+#include <Arduino.h>
+
+#include "sensesp_app.h"
+#include "signalk/signalk_output.h"
+#include "transforms/linear.h"
+
+// SensESP builds upon the ReactESP framework. Every ReactESP application
+// defines an "app" object vs defining a "main()" method.
+ReactESP app([]() {
+
+// Some initialization boilerplate when in debug mode...
+#ifndef SERIAL_DEBUG_DISABLED
+  SetupSerialDebug(115200);
+#endif
+
+  // Create the global SensESPApp() object.
+  sensesp_app = new SensESPApp();
+
+  // The "Signal K path" identifies this sensor to the Signal K server. Leaving
+  // this blank would indicate this particular sensor (or transform) does not
+  // broadcast Signal K data.
+  // To find valid Signal K Paths that fits your need you look at this link:
+  // https://signalk.org/specification/1.4.0/doc/vesselsBranch.html
+  const char* sk_path = "environment.indoor.illuminance";
+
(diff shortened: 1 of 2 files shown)


examples/multiple_sensors_example.cpp
```diff
@@ -0,0 +1,72 @@
...
+  const char* analog_in_config_path = "/indoor_illuminance/analog_in";
+  const char* linear_config_path = "/indoor_illuminance/linear";
+
+  // Create a sensor that is the source of our data, that will be read every 500
+  // ms. It's a light sensor that's connected to the ESP's AnalogIn pin. The
+  // AnalogIn pin on ESP8266 is always A0, but ESP32 has many pins that can be
+  // used for AnalogIn, and they're expressed here as the XX in GPIOXX.
+  // When it's dark, the sensor's output (as read by analogRead()) is 120, and
+  // when it's bright, the output is 850, for a range of 730.
+  uint8_t pin = A0;
```
mairas: To make the example build with ESP32 as is, could you have an ifdef to define the pin? Such as
```
#if defined(ESP8266)
#define ANALOG_INPUT_PIN A0
#elif defined(ESP32)
#define ANALOG_INPUT_PIN 12
#endif
```
after the includes?

examples/multiple_sensors_example.cpp
```diff
@@ -0,0 +1,72 @@
...
+  const float offset = -16.44;
+
+  // Wire up the output of the analog input to the Linear transform,
+  // and then output the results to the Signal K server. As part of
+  // that output, send some metadata to indicate that the "units"
+  // of the value we are sending is a "ratio" - which is the official
+  // unit type for a percentage represented as a float between 0.0 and 1.0
+  // See https://github.com/SignalK/specification/blob/master/schemas/definitions.json#L87
+  // for more details.
+  analog_input->connect_to(new Linear(multiplier, offset, linear_config_path))
```
mairas: I don't actually see multiple sensors in this example?

mairas (commented):
Sorry, I've been so buried with the SH-ESP32 work that I didn't notice this PR at all. The major comment is that I don't actually see multiple sensors in this example.

examples/multiple_sensors_example.cpp
ba58smith: You don't? There are three: temperature and humidity at the top, then the DigitalInutCounter at the bottom. You may have been thrown off by the IFDEF's in the middle, with two sensors above it. I can move those all to the top of the file if you like?

mairas: Nevermind! I was a doofus and was only staring at one of the two commits! It's perfect as is.

mairas:
As said, I was only staring at one of the two commits. It's great as is! Merging.

PR #383, opened 2021-03-07
Don't pin down the ReactESP version number

mairas:
I hadn't noticed that ReactESP version number was actually pinned down in library.json. This PR removes the version restriction.


mairas:
Merging.

PR #384, opened 2021-03-07
Transforms for dew point temperature and air density

GHBLoos:
Dew Point Temperature added for BME 280 sensor.


mairas:
Hey, kudos for your first PR! Dew point calculation is definitely a good feature to have. Since it is a function of temperature and humidity (and pressure), it would also be usable for other hardware sensors (and architecturally more apt) if it was a transform instead of part of the BME280 sensor.

Also, while at it, shouldn't it be using the pressure-aware formulations such as the Arden Buck equation? https://en.wikipedia.org/wiki/Dew_point#Calculating_the_dew_point

GHBLoos:
I can improve the code using the Arden Buck equation. But as I read it, you can use humidity _or_ pressure for the equation. So you don't need both.
A transform may be a good solution. Would it be possible to use temperature and humidity (or pressure) from two different sensors? I am just not good enough in coding to figure that out myself.

src/transforms/dew_point.cpp
```diff
@@ -0,0 +1,34 @@
+#include "dew_point.h"
+
+// dew point
+
+DewPoint::DewPoint() : NumericTransform() {}
+
+void DewPoint::set_input(float input, uint8_t inputChannel) {
+  inputs[inputChannel] = input;
+  received |= 1 << inputChannel;
+  if (received == 0b11) {
```
ba58smith, line 10: I'd like to see a comment briefly explaining what `if (received == 0b11)` is doing.

src/transforms/dew_point.cpp
```diff
@@ -0,0 +1,34 @@
...
+    received = 0;
+
+          float TC= inputs[0]-273.15;     // temperature in Celsius
+          float RH= inputs[1];            // relative humidity
+
```
ba58smith, line 15: spacing, for consistency and readability. As previously noted, we've spent a lot of time making the SensESP code follow the Google Style Guide, and be consistent.

above, in the example: (bme_temperature, bme_humidity)

TC = inputs
RH = inputs
and below:
TC < 0
gamma = log(RH * exp((b - (TC / d)) * (TC / (c + TC))));
(c * gamma) / (b - gamma);

ba58smith (approved):
Nice job! Just a few housekeeping comments.

src/transforms/dew_point.cpp
GHBLoos: I copied this from difference.cpp. To be honest, I don't really understand why it is like that.

src/sensors/bme280.cpp
```diff
@@ -46,7 +47,20 @@ void BME280Value::enable() {
...
+
+          float h= bme280_->adafruit_bme280_->readHumidity()/100; //humidity in percent, so divide by 100
+          float t= bme280_->adafruit_bme280_->readTemperature();
+
+          float alfa=(a*t)/(b+t)+ log(h);
+
+          float dpt = (b*alfa)/(a-alfa); 
+        
+          output = dpt + 273.15; // Kelvin is Celsius + 273.15
+       } else {
```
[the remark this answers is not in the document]
GHBLoos: Ok, working on that. Thanks for pointing me to the Google Style Guide. That helps.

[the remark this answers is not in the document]
GHBLoos: Looking through that guide:
`enum BME280ValType { temperature, pressure, humidity, dewPointTemperature };`
Shouldn't that be like
`enum BME280ValType { kTemperature, kPressure, kHumidity, kDewPointTemperature };`
(https://google.github.io/styleguide/cppguide.html#Enumerator_Names)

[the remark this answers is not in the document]
ba58smith: You're correct. We don't follow the Guide to the letter,  but we do try to be very consistent with the parts we do follow. If you want to use the leading k in this Transform, I don't think anyone will object,  and maybe you'll start a trend!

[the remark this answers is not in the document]
mairas: Regarding code formatting: if you're using Visual Studio Code you can get your code automatically formatted (spaces, indentation, line widths, whatnot) according to the project standards by applying the command "Format Document" (either via Cmd-P or Ctrl-P, or by shortcut Alt-Shift-F - not sure if the latter is like that on every platform, though).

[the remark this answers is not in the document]
GHBLoos: Learned something again today! Thanks. That saves a lot of work.

GHBLoos:
Sorry for the mess.

GHBLoos:
I think this is completed now. I am not sure how to finish this process. Do I have to do anything? Or are you going to decide to merge it or not. Let me know if I need to do something.

examples/bme280_example.cpp
```diff
@@ -52,5 +54,22 @@ ReactESP app([]() {
...
+
+  // Use the transform dewPoint to calculate the dewpoint based upon the temperature and humidity.
+  auto* dew_point  = 
+      new DewPoint();
+
+  dew_point->connect_from(bme_temperature, bme_humidity)
+          ->connect_to(new SKOutputNumber("environment.outside.dewPointTemperature"));
+
+  // Use the transform airDensity to calculate the air density of humid air ased
+  // upon the temperature, humidity and pressure.
```
ba58smith, line 66: Typo: "ased" should be "based".

examples/bme280_example.cpp
```diff
@@ -52,5 +54,22 @@ ReactESP app([]() {
 
   bme_humidity->connect_to(new SKOutputNumber("environment.outside.humidity"));
 
+
+  // Use the transform dewPoint to calculate the dewpoint based upon the temperature and humidity.
+  auto* dew_point  = 
+      new DewPoint();
```
ba58smith, line 60: Did the PlatformIO auto-format break the above into two lines? Looks weird. Of course, it works either way, but seems like it should be a single line.

examples/bme280_example.cpp
```diff
@@ -52,5 +54,22 @@ ReactESP app([]() {
...
+
+  dew_point->connect_from(bme_temperature, bme_humidity)
+          ->connect_to(new SKOutputNumber("environment.outside.dewPointTemperature"));
+
+  // Use the transform airDensity to calculate the air density of humid air ased
+  // upon the temperature, humidity and pressure.
+  auto* airDensity = 
+      new AirDensity();
```
ba58smith, line 68: Same comment as above about this being two lines instead of one.

ba58smith (approved):
One typo, two very minor formatting questions. Other than that, it looks good to me. Nice job!

I would like @mairas to do the final review and merge.

examples/bme280_example.cpp
GHBLoos: I am not sure, but I put it on a single line now.

mairas:
This looks great! Good work! Git commit history could be reworked a bit but I don't have the heart to torture you with it so I'll just squash and merge. :-)

PR #386, opened 2021-03-20
Pin ADS1x15 version due to incompatible changes in 2.0.0

mairas:


mairas:
The approach was agreed upon in advance and now the builds pass. Merging.

PR #389, opened 2021-03-20
Make ADS1x15 work with new version of Adafruit library

ba58smith:


mairas:
Thanks, and sorry for the delay!

PR #391, opened 2021-03-23
Heat_index transform added

GHBLoos:
Heat index transform added.
Example of transform added in BME280_example (and I changed a key from `inside.engineroom` to `outside`)

### DIFF
--- a/src/transforms/heat_index.cpp
+++ b/src/transforms/heat_index.cpp
@@ -1,0 +1,68 @@
+#include "heat_index.h"
+
+// heat index temperature
+
+HeatIndexTemperature::HeatIndexTemperature() : NumericTransform() {}
+
+void HeatIndexTemperature::set_input(float input, uint8_t inputChannel) {
+  inputs[inputChannel] = input;
+  received |= 1 << inputChannel;
+  if (received ==
+      0b11) {  // for 2 channels, use 0b11. For 3 channels, use b111 and so on.
+    received = 0;  // recalculates after all values are updated. Remove if a
+                   // recalculation is required after an update of any value.
+
+    // The following equation approximate the heat index.
+    // For more info on the calculation see
+    // https://en.wikipedia.org/wiki/Dew_point#Calculating_the_dew_point
+
+    float temp_celsius =
+        inputs[0] - 273.15;  // dry-bulb temperature (in degrees Celsius)
+    float relative_humidity = inputs[1];
+
+    // Coefficients for temperature in Celsius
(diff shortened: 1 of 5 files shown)


src/transforms/heat_index.cpp
```diff
@@ -0,0 +1,68 @@
...
+  }
+}
+
+// heat index effect (warning levels)
+
+HeatIndexEffect::HeatIndexEffect() : Transform<float, String>() {}
+
+void HeatIndexEffect::set_input(float input, uint8_t inputChannel) {
+  float heat_index_temperature = input;
+  String heat_index_effect = "";
```
ba58smith, line 56: One very minor suggestion: make it `float heat_index_temp_kelvin = input + 273.15`, then all of your comparisons can be like `if (heat_index_temp_kelvin > 54)`. Obviously it works as you have it - this would just be a tiny bit cleaner.

ba58smith (approved):
Except for one teeny-tiny suggestion, it looks good to me. I assume you've tested it?

GHBLoos:
I have one thing that bothers me in the calculation. I think that below 40% humidity and below 26C temperature, there is no heat index temperature. That would mean that if one of these is below that treshold, it's just the (dry bulb) temperature. But Internet is not very clear about that. 
At this point I think I should add an if-then statement for the humidity and temperature treshold.
I will search some more info on internet about this.

ba58smith:
It seems like an awful lot of work to be able to display a number that's supposed to tell someone how the weather feels. :-) But since it's a Transform, no one will use it who doesn't want that info.
@mairas, you want to give this your blessing before merging?

GHBLoos:
I used different algorithms. That also solves the problem of temperatures and humidity below 27C and 40% humidity. The algorithm can be found [here](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3801457/figure/f3/). According to [this paper](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3801457/) from 2013 it is the best algorithm.

[This online calculator](https://www.wpc.ncep.noaa.gov/html/heatindex.shtml) can be used to check the calculation, which would be great if someone could test it that is in an environment that has high temperatures and humidity (unfortunately where I live it is not that warm).

src/transforms/heat_index.cpp
ba58smith: Sorry, I had the math wrong!

mairas:
Looks good to me! Merging.

PR #393, opened 2021-03-27
Add instructions to notify platformio about the update

mairas:


mairas:
Depending on which way [the discussion at PlatformIO community forum](https://community.platformio.org/t/publishing-a-new-version-of-signalk-sensesp/20277/5) goes, this step might not be needed.

PR #396, opened 2021-03-29
Bump master branch version number to 1.0.4-alpha

mairas:
To better differentiate between the stable releases and the development branch, this PR bumps the development branch version number to 1.0.4-alpha. This should not imply that the next release version is 1.0.4 -- it could be that, or 1.1.0, or 2.00, depending on the changes done during the development cycle.

### DIFF
--- a/Doxyfile
+++ b/Doxyfile
@@ -38,7 +38,7 @@ PROJECT_NAME           = SensESP
 # could be handy for archiving the generated documentation or if some version
 # control system is used.
 
-PROJECT_NUMBER         = 1.0.3
+PROJECT_NUMBER         = 1.0.4-alpha
 
 # Using the PROJECT_BRIEF tag one can provide an optional one line description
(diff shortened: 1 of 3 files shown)


ba58smith:
Do you anticipate any issues with the PlatformIO version-recognition code having a problem with `-alpha`? I don't remember ever seeing any libraries with alpha characters in the version - only numeric.

mairas:
> Do you anticipate any issues with the PlatformIO version-recognition code having a problem with `-alpha`? I don't remember ever seeing any libraries with alpha characters in the version - only numeric.

A dash is a standard [semantic versioning](semver.org) separator for pre-release versions and PlatformIO claims to support semantic versioning, so I'm pretty sure it works fine.

But to be clear, that version would never be released; it'd just be shown if people are using the `https://github.com/SignalK/SensESP.git` dependency.

mairas:
Barring no comments, merging.

PR #398, opened 2021-04-04
bme280.cpp decimal to hex

GHBLoos:
Debug output of SensorID was in decimal, but should be in hexadecimal.

was `debugI("SensorID is: 0x%d", adafruit_bme280_->sensorID());`  
is now `debugI("SensorID is: 0x%T", adafruit_bme280_->sensorID());`

`%d` -> `%T`


ba58smith:
I can't test this, as I don't have one of these sensors. But if you tell me you've tested it and it works, I'll merge the PR.

Thanks!

GHBLoos:
I did test it and it works. You can easily test it yourself, because this part of the code runs when a sensor is not found.

PR #399, opened 2021-04-04
1-Wire error detection and library change

mairas:
DallasTemperature sensor implementation now correctly detects the read error codes. Under the hood, the 1-Wire library has been changed from the stale `PaulStoffregen/OneWire` to `pstolarz/OneWireNg`. This seems to fix all the read errors on ESP32 and works unchanged on ESP8266 too.

### DIFF
--- a/src/sensors/onewire_temperature.cpp
+++ b/src/sensors/onewire_temperature.cpp
@@ -1,7 +1,6 @@
 #include "onewire_temperature.h"
 
 #include <DallasTemperature.h>
-#include <OneWire.h>
 
 #include <algorithm>
 
--- a/src/sensors/onewire_temperature.cpp
+++ b/src/sensors/onewire_temperature.cpp
@@ -139,8 +139,17 @@ void OneWireTemperature::update() {
 }
 
 void OneWireTemperature::read_value() {
+  float tempC = dts_->sensors_->getTempC(address_.data());
+  // we're on purpose ignoring the "conversion not ready" value (+85°C)
+  // because the update method always waits for 750 ms before reading the value
+  if (tempC == DEVICE_DISCONNECTED_C) {
+    char ow_addr_str[24];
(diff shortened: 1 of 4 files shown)


src/sensors/onewire_temperature.cpp
```diff
@@ -1,7 +1,7 @@
 #include "onewire_temperature.h"
 
+#include <OneWireNg_Wrapper.h>
```
pstolarz, line 3: No need to include this header here. `DallasTemperature.h` includes all required headers depending on the library config (OneWire vs OneWireNg usage).

src/sensors/onewire_temperature.h
```diff
@@ -3,8 +3,8 @@
 
 #include <set>
 
+#include <OneWireNg_Wrapper.h>
 #include <DallasTemperature.h>
```
pstolarz, line 7: Same as above

src/sensors/onewire_temperature.cpp
mairas: Ok. I actually thought as much, but for some reason it didn't work at the first attempt. Gotta try again!

But... How did you find this PR?! 😂

pstolarz: If someone uses my lib I tend to get informed really fast ;) I hope it won't make me ashamed and will work as expected.
Terveisiä Puolasta!

mairas: What I had tried was to just keep importing "OneWire.h" with your DallasTemperature fork. That of course wouldn't work. Hm, maybe I'll start creating issues in the OneWireNg repo... ;-)

src/sensors/onewire_temperature.h
mairas: Resolved above.

pstolarz (approved):

PR #412, opened 2021-04-16
Remove final tag in order to enable overriding

jcaprano:
Not making the set_input method final will allow for overriding in subclasses to implement other functionality such as outputting the discrete value of the closest sample instead of interpolating.


mairas:
Thanks! Merging.

jcaprano:
Thanks for merging! Looking forward to making more valuable contributions :)

PR #413, opened 2021-04-16
Make samples member protected

jcaprano:
Not 100% sure if this is ok architecture wise: In order to override get_input in a subclass, I also need to be able to access samples since I still need to iterate over the samples in the implementation in my subclass. Please let me know if there is a more elegant way of doing this!


mairas:
Just out of interest, why did you close this PR after all?

jcaprano:
didn't mean to! actually I renamed the branch and that might have accidentally closed it..

PR #414, opened 2021-04-16
Changed samples from private to protected

jcaprano:
Not 100% sure if this is ok architecture wise: In order to override get_input in a subclass, I also need to be able to access samples since I still need to iterate over the samples in the implementation in my subclass. Please let me know if there is a more elegant way of doing this!

### DIFF
--- a/src/transforms/curveinterpolator.h
+++ b/src/transforms/curveinterpolator.h
@@ -56,7 +56,7 @@ class CurveInterpolator : public NumericTransform {
     add_sample(new_sample);
   }
 
- private:
+ protected:
   std::set<Sample> samples;
 };
 


jcaprano:
Here it is again, I think renaming the branch with an open PR wasn't the smartest idea.. I have explained a bit of the background of this to @Butch on Slack 👍

mairas:
The PR looks good.

I read the description on Slack. The idea seemed good; I'm just wondering if you could get sufficiently similar results by adding a sufficiently long median filter after the input.

Anyway, merging. Thanks!

jcaprano:
Thank you! That's a good point, should work too :)

PR #417, opened 2021-05-05
WiFi.localIP() returns broadcast address 255.255.255.255 when using workaround

frewie:
When using this workaround, WiFi.localIP() returns 255.255.255.255 after a successful wifi connect and the signal k server can't be found with mdns.  
I tested that for the ESP8266 v2 with platform version Espressif 8266 v2.6.3

see here: https://github.com/espressif/arduino-esp32/issues/4732

### DIFF
--- a/src/net/networking.cpp
+++ b/src/net/networking.cpp
@@ -73,7 +73,7 @@ void Networking::setup_saved_ssid() {
 
   // this is a workaround for setting the DHCP hostname, suggested in
   // https://github.com/espressif/arduino-esp32/issues/2537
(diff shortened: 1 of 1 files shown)


mairas:
Aww, cool! Thanks!

Instead of commenting out, could you simply remove the offending line (and the comments above)? No benefit in leaving broken code lying around! :-)

ba58smith:
Instead of waiting for that little bit of cleanup, I'm going to merge this PR, and then make another PR for the cleanup. This bug is biting people on a daily basis.

PR #418, opened 2021-05-07
Add Git installation to README

ba58smith:
To eliminate support "calls" when all of the libraries don't get downloaded.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ NOTE: SensESP does not currently support SSL. If you enabled that when you insta
 Once the SK Server is installed and running, go to the Dashboard (enter `localhost:3000` into the Raspberry Pi's browser to start it), select Server - Settings from the left side menu, and make sure the "mdns" option is ON.
 
 SensESP is a library and will be installed automatically as a dependency when defined as such in the project's
-`platformio.ini` file. Instructions below.
+`platformio.ini` file. Instructions below. However, in order for PlatformIO to install all of the libraries that SensESP uses, you must have Git installed. If you do already, you should update it to the latest version. If you don't, install it now: https://git-scm.com/book/en/v2/Getting-Started-Installing-Git
 
 You need to have PlatformIO installed. First, download and install [Visual Studio Code](https://code.visualstudio.com/).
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -18,7 +18,7 @@ NOTE: SensESP does not currently support SSL. If you enabled that when you insta
 Once the SK Server is installed and running, go to the Dashboard (enter `localhost:3000` into the Raspberry Pi's browser to start it), select Server - Settings from the left side menu, and make sure the "mdns" option is ON.
 
 SensESP is a library and will be installed automatically as a dependency when defined as such in the project's
-`platformio.ini` file. Instructions below.
+`platformio.ini` file. Instructions below. However, in order for PlatformIO to install all of the libraries that SensESP uses, you must have Git installed. If you do already, you should update it to the latest version. If you don't, install it now: https://git-scm.com/book/en/v2/Getting-Started-Installing-Git
```
mairas, line 21: "you should update it to the latest version":

What features does the latest (2.31.1) Git version provide as opposed to the one included in your Linux distribution or Macos Xcode command line tools or Homebrew?

AFAIK, Git has extremely high backwards compatibility and manually updating to a version not provided by your OS tools just increases the risk of messing things up.

ba58smith: So you're OK with giving instructions to install the latest version of Git if you don't have it innstalled, but not OK with suggesting they update to the latest version? Isn't it just considered good practice to always be working with the latest released version of software, especially such widely-used software as Git? I'm not trying to argue, I just don't quite understand the objection.

mairas: My point is that as long as your Git version is less than, say, ten years old, in our use case it makes absolutely no difference which exact Git version you have. It's a bit like suggesting that you should be using the latest version of the bash shell - sure, upgrading is possible, but is complicated to do because you're messing with your operating system standard tools, and adds a lot of complexity and risks breaking things. And it very rarely solves any problems, unless you are trying to use some esoteric or cutting edge features.

Also, if anyone is having trouble getting their project built, I wouldn't want to give them any kind of misconception that the Git version they're using might have any impact on the problem they're having. Because it doesn't, and if they start updating their Git version, it's just creating more trouble for them.

PR #420, opened 2021-05-08
Prevent Configure <hostname> AP from showing

ba58smith:
After wifi is configured, the "Configure <hostname>" AP would still appear in a list of available AP's. This solves that.


mairas:
I could not replicate the original issue on ESP32. If this issue appears on ESP8266, I would like to have `#ifdef ESP8266`'s around the new line of code to prevent any possible regression on ESP32.

mairas:
Thanks!

PR #444, opened 2021-10-25
Implement SensESPBaseApp

mairas:
A SensESPBaseApp is implemented, allowing to construct a minimal SensESP app without all the standard scaffolding (networking, WS client, etc).


mairas:
Modified according to discussion. `SensESPBaseApp` is now (more or less) an abstract base class, and and concrete classes should be inherited from it. `SensESPMinimalApp` is the minimal class implementation that should be used for building lightweight apps.

mairas:
Not so fast... A SensESPBaseAppBuilder ABC is needed.

mairas:
OK, it's good now.

PR #474, opened 2021-11-27
ADS1115 output should be signed

### LINKED ISSUE #473: ADS1115 sensor code does not work when one of the ADC inputs is grounded.
The ADS1115 ADC outputs a 2's complement 16 bit binary integer on conversion with +MAX FS as 7FFF h and -MAX FS as 8000 h. If the ADC input AINx is grounded, the ADC can still output a small negative value because of noise or internal offsets. The ADS1x15 sensor code does not handle this case, treating the ADC output as a unsigned integer so that the grounded ADC voltage output appears as nearly 2x the MAX FS range.

mairas:
One-character bugfixes are the best.

I didn't have suitable hardware at hand so I didn't actually test the change. Let's summon @RBerliner for that!

Fixes #473.

### DIFF
--- a/src/sensors/ads1x15.h
+++ b/src/sensors/ads1x15.h
@@ -67,7 +67,7 @@ class ADS1x15RawValue : public NumericSensor {
   T_ads_1x15* ads1x15_;
   ADS1x15Channel_t channel_;
   uint read_delay_;
-  uint16_t raw_value_ = 0;
+  int16_t raw_value_ = 0;
 
  private:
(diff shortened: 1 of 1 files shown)


mairas:
Verified in the issue #473 discussion. Merging.

dpazz:
Please where can be sourced the ads1115 code. I'm in the development of an interface to old VDO instruments (analog 110 mm but I like them) and i have to make a precise ADC for the sin/cos output coming from the sensor.

PR #475, opened 2021-11-27
ReactESP v2 changes

mairas:
I released ReactESP v2 earlier today. This PR implements the required changes due to the updated API.

### DIFF
--- a/examples/chain_counter.cpp
+++ b/examples/chain_counter.cpp
@@ -2,11 +2,11 @@
 
 #include "sensesp_app.h"
 #include "sensesp_app_builder.h"
-#include "system/lambda_consumer.h"
 #include "sensors/digital_input.h"
-#include "transforms/integrator.h"
-#include "transforms/debounce.h"
 #include "signalk/signalk_output.h"
+#include "system/lambda_consumer.h"
+#include "transforms/debounce.h"
+#include "transforms/integrator.h"
 
 using namespace sensesp;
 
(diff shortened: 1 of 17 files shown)


examples/chain_counter.cpp
```diff
@@ -41,115 +42,120 @@ ReactESP app([]() {
...
   metadata->description_ = "Anchor Rode Deployed";
   metadata->display_name_ = "Rode Deployed";
   metadata->short_name_ = "Rode Out";
 
-/**
- * chain_counter is connected to accumulator, which is connected to an SKOutputNumber,
- * which sends the final result to the indicated path on the Signal K server. (Note that
- * each data type has its own version of SKOutput: SKOutputNumber for floats, SKOutputInt,
- * SKOutputBool, and SKOutputString.)
- */
```
ba58smith, line 84: Didn't you eliminate the specific types of SKOutput, by making it a template type?

mairas: It's a template type, but for convenience, those specific types are still defined at the end of `signalk_output.h`:

```
typedef SKOutputNumeric<float> SKOutputFloat;
typedef SKOutputNumeric<int> SKOutputInt;
typedef SKOutput<bool> SKOutputBool;
typedef SKOutput<String> SKOutputString;
```

If it was just up to me, I would indeed have a mild preference for ditching those typedefs and just using the template syntax, but I've accepted the notion that there are SensESP users who're not comfortable with the C++ template class syntax and it's better to have the typedefs around. :-)

PR #479, opened 2021-12-04
Make OTA authentication mandatory

mairas:
Over-the-air updates are now enabled only if a password has been provided. This improves safety (a lot) and saves memory if OTA is not needed.


mairas:
Doh. Should've been against `v2-dev`

PR #486, opened 2021-12-09
Fix analog_input example: make the output a ratio, not a percentage

ba58smith:
Also rewrite the comments to better explain how to calculate the Linear multiplier and offset.


mairas:
Looks good! Thanks!

Issue #484, opened 2021-12-08
analog_input.cpp example outputs percentage instead of ratio


ba58smith:
When originally created, this example output its value as a percentage, from 0 to 100.

When the ability to add metadata to the SKOutput constructor was added, the example had `ratio` added as its `unit` in the metadata:
`->connect_to(new SKOutputNumber(sk_path, "", new SKMetadata("ratio")));`

So I think the `linear` transform's `multiplier` value needs to be changed from `0.137` to `0.00137`. @mairas - if this seems right to you, I'll make the change and the PR.

ba58smith:
Fixed in `master` with PR #468.

PR #487, opened 2021-12-10
Make analog_input example output a ratio instead of a percentage

ba58smith:
Also update the comments to better document the calculation of the Linear multiplier and offset.


ba58smith:
Addresses Issue #484.

mairas:
Ditto.

PR #492, opened 2021-12-14
Remove set_test_auth_on_each_connect builder member function

mairas:
This was an optimization that, when turned off, would make the websocket connection logic bypass testing the authorization token on reconnects. Since this is very much a corner case and made the already hairy connection logic even more complex, I think it's better to remove it.


ba58smith:
I sure don't need it!

mairas:
Tested by having a test program connect to the server and then shutting down and restarting the server process. Would seem to work fine.

PR #493, opened 2021-12-15
Implement a RepeatSensor template class

mairas:
RepeatSensor can be used to interface any generic arduino-esp32 compatible sensor library with SensESP. You initialize an upstream sensor interface object and set it up in `setup()`. Next, you define a callback function which will read and return a new value from the sensor interface. Finally, you instantiate `RepeatingSensor` and define the return type, repeat interval, and provide the callback function. The created `RepeatingSensor` object is a normal producer that can be connected to transforms and output objects as usual.

### DIFF
--- a/examples/async_repeat_sensor.cpp
+++ b/examples/async_repeat_sensor.cpp
@@ -1,0 +1,75 @@
+/**
+ * @file async_repeat_sensor.cpp
+ * @brief Example of an asynchronous RepeatSensor that reads a digital input.
+ *
+ * RepeatSensor is a helper class that can be used to wrap any generic Arduino
+ * sensor library into a SensESP sensor. This file demonstrates its use with
+ * the built-in GPIO digital input, but in an asynchronous manner.
+ *
+ */
+
+#include "sensesp_app_builder.h"
+#include "signalk/signalk_output.h"
+#include "transforms/linear.h"
+
+using namespace sensesp;
+
+reactesp::ReactESP app;
+
+// The setup function performs one-time application initialization.
+void setup() {
+// Some initialization boilerplate when in debug mode...
+#ifndef SERIAL_DEBUG_DISABLED
+  SetupSerialDebug(115200);
+#endif
+
+  // Create the global SensESPApp() object.
+  SensESPAppBuilder builder;
+  sensesp_app = builder.get_app();
+
+  const uint8_t kDigitalInputPin = 15;
+
+  // Do here any sensor initialization you need. For a more complex sensor,
+  // this might include initializing the hardware, setting the I2C pins, etc.
+  pinMode(kDigitalInputPin, INPUT_PULLUP);
+
+  // GPIO pin that we'll be using for the analog input.
+
+  // define an asynchronous callback function that reads a digital input pin
+  // after a delay of 1000 ms.
+  auto analog_read_callback = [](RepeatSensor<bool>* sensor) {
+    debugI("Pretend to trigger an asynchronous measurement operation here.");
+    app.onDelay(
+        1000, [sensor]() { sensor->emit(digitalRead(kDigitalInputPin)); });
+  };
+
+  // Let's read the sensor every 2000 ms.
+  unsigned int read_interval = 2000;
+
+  // Create a RepeatSensor with float output that reads the analog input.
+  auto* digital =
+      new RepeatSensor<bool>(read_interval, analog_read_callback);
+
+  // The "Signal K path" identifies this sensor to the Signal K server. Leaving
+  // this blank would indicate this particular sensor (or transform) does not
+  // broadcast Signal K data.
+  // To find valid Signal K Paths that fits your need you look at this link:
+  // https://signalk.org/specification/1.4.0/doc/vesselsBranch.html
+  const char* sk_path = "environment.bool.pin15";
+
+  // Connect the output of the analog input to the Linear transform,
+  // and then output the results to the Signal K server. As part of
+  // that output, send some metadata to indicate that the "units"
+  // to be used to display this value is "ratio". Also specify that
+  // the display name for this value, to be used by any Signal K
+  // consumer that displays it, is "Indoor light".
+  digital->connect_to(
+      new SKOutputFloat(sk_path, ""));
+
+  // Start the SensESP application running
+  sensesp_app->start();
+}
+
+// The loop function is called in an endless loop during program execution.
+// It simply calls `app.tick()` which will then execute all reactions as needed.
+void loop() { app.tick(); }
--- a/examples/repeat_sensor_analog_input.cpp
+++ b/examples/repeat_sensor_analog_input.cpp
@@ -1,0 +1,77 @@
+/**
+ * @file repeat_sensor_analog_input.cpp
+ * @brief Example of a RepeatSensor that reads an analog input.
+ *
+ * RepeatSensor is a helper class that can be used to wrap any generic Arduino
+ * sensor library into a SensESP sensor. This file demonstrates its use with
+ * the built-in analog input sensor.
+ *
+ * Note that for analog input, it's actually better to use the AnalogInput
+ * sensor class as demonstrated by the analog_input.cpp example file. The
+ * AnalogInput class compensates for the ESP32 ADC nonlinearities. The approach
+ * presented here, however, can be used with any sensor library compatible with
+ * arduino-esp32.
+ *
+ */
+
+#include "sensesp_app_builder.h"
+#include "signalk/signalk_output.h"
+#include "transforms/linear.h"
+
+using namespace sensesp;
+
+reactesp::ReactESP app;
+
+// GPIO pin that we'll be using for the analog input.
+const uint8_t kAnalogInputPin = 36;
+
+// define a callback function that reads the analog input
+float analog_read_callback() { return analogRead(kAnalogInputPin) / 4096.0; }
+
+// The setup function performs one-time application initialization.
+void setup() {
+// Some initialization boilerplate when in debug mode...
+#ifndef SERIAL_DEBUG_DISABLED
+  SetupSerialDebug(115200);
+#endif
+
+  // Create the global SensESPApp() object.
+  SensESPAppBuilder builder;
+  sensesp_app = builder.get_app();
+
+  // Do here any sensor initialization you need. For a more complex sensor,
+  // this might include initializing the hardware, setting the I2C pins, etc.
+  // In this case, we'll just set the analog input attenuation to 2.5 dB,
+  // corresponding to maximum input voltage of 1.1 V.
+  analogSetAttenuation(ADC_2_5db);
+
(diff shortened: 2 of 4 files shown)


examples/repeat_sensor_analog_input.cpp
```diff
@@ -0,0 +1,77 @@
...
+
+using namespace sensesp;
+
+reactesp::ReactESP app;
+
+// GPIO pin that we'll be using for the analog input.
+const uint8_t kAnalogInputPin = 36;
+
+// define a callback function that reads the analog input
+float analog_read_callback() { return analogRead(kAnalogInputPin) / 4096.0; }
```
ba58smith, line 29: Could the callback function and kAnalogInputPin be declared and defined inside the setup() function? Or would that violate something? It just seems like creating these two things is something that's part of "setting up" the rest of the program. I'm asking more for my own education than anything else.

mairas: > Could the callback function and kAnalogInputPin be declared and defined inside the setup() function? Or would that violate something?

Well, yes, with lambdas... Other than that, nested functions are not allowed in C++. But given that I've heard some people complain so fervently against the use of lambdas, I caved in and wrote the example with regular functions...

examples/repeat_sensor_analog_input.cpp
```diff
@@ -0,0 +1,77 @@
...
+ *
+ * RepeatSensor is a helper class that can be used to wrap any generic Arduino
+ * sensor library into a SensESP sensor. This file demonstrates its use with
+ * the built-in analog input sensor.
+ *
+ * Note that for analog input, it's actually better to use the AnalogInput
+ * sensor class as demonstrated by the analog_input.cpp example file. The
+ * AnalogInput class compensates for the ESP32 ADC nonlinearities. The approach
+ * presented here, however, can be used with any sensor library compatible with
+ * arduino-esp32.
```
ba58smith, line 13: I think the example would be a whole lot more educational if it used an actual external sensor, like an SHT31. That would illustrate all the key concepts better than the AnalogInput, I think: what to `#include` to use an external library; the specific steps necessary to instantiate and then "start" the sensor; and then, rather importantly, how to use this approach to get both the temperature and the humidity from the sensor.

It would also be an excellent template for all those other sensors that you want to get rid of: the BME's and BMP's and INA's and ADS's. You'd kill a lot of birds with just this one stone.

Not to mention the fact that doing this with the AnalogInput is not something anyone would ever do, so it is - and it feels - totally contrived.

I'm guessing you'd prefer not to do that because then you'd have to include the Adafruit library for that sensor in the core SensESP code. But do you? Could this example somehow be excluded from the test script that tests all of the examples? I actually have an SHT31 that I could test it with, so we'd know for sure it's working. But maybe you don't want to make that kind of exception for other reasons.

Just a thought.

src/sensors/sensor.h
```diff
@@ -51,6 +51,35 @@ typedef SensorT<float> FloatSensor;
...
+  }
+
+ protected:
+  unsigned int repeat_interval_ms_;
+  std::function<T()> callback_;
+};
+
 }  // namespace sensesp
 
 #endif
```
ba58smith, line 85: This is brilliant, by the way. Too bad we didn't have this discussion before I started banging out all those cookie-cutter Sensors.

examples/repeat_sensor_analog_input.cpp
ba58smith: :-) Snarkiness translates perfectly from Finnish to English. :-)

But you answered my question, thank you - no nested functions in C++.

src/sensors/sensor.h
```diff
@@ -51,6 +51,35 @@ typedef SensorT<float> FloatSensor;
...
+   * RepeatSensor is a sensor that calls a callback function at given intervals
+   * and produces the value returned by the callback function. It can be used
+   * to wrap any generic Arduino sensor library into a SensESP sensor.
+   *
+   * @param repeat_interval_ms The repeating interval, in milliseconds.
+   * @param callback A callback function that returns the value the sensor will produce.
+   * @tparam T The type of the value returned by the callback function.
+   */
+  RepeatSensor<T>(unsigned int repeat_interval_ms, std::function<T()> callback)
+      : SensorT<T>(""),
```
ba58smith, line 69: Would you see any benefit of making the callback be the first parameter to the constructor, so that the repeat_interval can be the second one, and therefore have a default value? That would make the use of it in main.cpp even simpler than it already is (which is VERY simple), if the user is OK with the default read_interval (say, 500ms):
 `auto* analog_input =  new RepeatSensor<float>(analog_read_callback);`

Not all that important, obviously - just a thought.

examples/repeat_sensor_analog_input.cpp
mairas: It's a good idea. There'd need to be a custom platformio.ini file for the purpose to actually allow building the example.

But, I'd _also_ want to have the analog input version there, the reason being that I'd like it to be included in the GitHub test build. If something breaks, there's a better chance it'll get caught that way.

Also, one way to do it is to make an example GitHub repo that features such an external sensor. Then anyone could clone the repo or download it as a zip file and run the project without copying the files in place etc.

src/sensors/sensor.h
mairas: Well, I hadn't thought of it back then, either!

src/sensors/sensor.h
mairas: I wouldn't do that: if the callback is defined as an in-place lambda, any arguments that come after the lambda get pushed really far away and become difficult to notice. I think for that reason it's customary to have any callbacks as the last argument.

mairas:
While skimming through the different HW sensors in SensESP v1, I noticed the MAX31856 thermocouple. It uses asynchronous reading: first the reading is triggered with one function call, then, after a small period of time, the value is read and emitted. This wouldn't be possible with the simplistic approach initially implemented.

So, I added support for an alternative callback format, which receives a pointer to the sensor and then manually calls Sensor::emit() when ready. There's also a new example to showcase (and test) that behavior.

examples/repeat_sensor_analog_input.cpp
ba58smith: I like the idea of a separate GitHub repo for it, with the example there. The documentation can refer to it, as can the AnalogInput example that you already wrote (in a comment - after the new repo exists).

examples/async_repeat_sensor.cpp
```diff
@@ -0,0 +1,75 @@
...
+  SensESPAppBuilder builder;
+  sensesp_app = builder.get_app();
+
+  const uint8_t kDigitalInputPin = 15;
+
+  // Do here any sensor initialization you need. For a more complex sensor,
+  // this might include initializing the hardware, setting the I2C pins, etc.
+  pinMode(kDigitalInputPin, INPUT_PULLUP);
+
+  // GPIO pin that we'll be using for the analog input.
```
ba58smith, line 36: Shouldn't this say "digital" input, and shouldn't it be above line 30?

examples/async_repeat_sensor.cpp
```diff
@@ -0,0 +1,75 @@
...
+
+  // define an asynchronous callback function that reads a digital input pin
+  // after a delay of 1000 ms.
+  auto analog_read_callback = [](RepeatSensor<bool>* sensor) {
```
ba58smith, line 40: Shouldn't this be "digital_read_callback"?

examples/async_repeat_sensor.cpp
```diff
@@ -0,0 +1,75 @@
...
+    app.onDelay(
+        1000, [sensor]() { sensor->emit(digitalRead(kDigitalInputPin)); });
+  };
+
+  // Let's read the sensor every 2000 ms.
+  unsigned int read_interval = 2000;
+
+  // Create a RepeatSensor with float output that reads the analog input.
+  auto* digital =
+      new RepeatSensor<bool>(read_interval, analog_read_callback);
```
ba58smith, line 51: Ditto - "digital_read_callback"?

examples/async_repeat_sensor.cpp
```diff
@@ -0,0 +1,75 @@
...
+  // To find valid Signal K Paths that fits your need you look at this link:
+  // https://signalk.org/specification/1.4.0/doc/vesselsBranch.html
+  const char* sk_path = "environment.bool.pin15";
+
+  // Connect the output of the analog input to the Linear transform,
+  // and then output the results to the Signal K server. As part of
+  // that output, send some metadata to indicate that the "units"
+  // to be used to display this value is "ratio". Also specify that
+  // the display name for this value, to be used by any Signal K
+  // consumer that displays it, is "Indoor light".
```
ba58smith, line 65: This whole comment is wrong for this example.

examples/repeat_sensor_analog_input.cpp
```diff
@@ -0,0 +1,77 @@
...
+ *
+ */
+
+#include "sensesp_app_builder.h"
+#include "signalk/signalk_output.h"
+#include "transforms/linear.h"
```
ba58smith, line 19: A Linear transform is not used in this example.

mairas:
Dammit! Seems like I merged prematurely. I'll make a new PR for the fixes.

examples/repeat_sensor_analog_input.cpp
```diff
@@ -0,0 +1,77 @@
...
+      new RepeatSensor<float>(read_interval, analog_read_callback);
+
+  // The "Signal K path" identifies this sensor to the Signal K server. Leaving
+  // this blank would indicate this particular sensor (or transform) does not
+  // broadcast Signal K data.
+  // To find valid Signal K Paths that fits your need you look at this link:
+  // https://signalk.org/specification/1.4.0/doc/vesselsBranch.html
+  const char* sk_path = "environment.indoor.illuminance";
+
+  // Connect the output of the analog input to the Linear transform,
```
ba58smith, line 62: There is no Linear in this example.

examples/async_repeat_sensor.cpp
```diff
@@ -0,0 +1,75 @@
...
+  pinMode(kDigitalInputPin, INPUT_PULLUP);
+
+  // GPIO pin that we'll be using for the analog input.
+
+  // define an asynchronous callback function that reads a digital input pin
+  // after a delay of 1000 ms.
+  auto analog_read_callback = [](RepeatSensor<bool>* sensor) {
+    debugI("Pretend to trigger an asynchronous measurement operation here.");
+    app.onDelay(
+        1000, [sensor]() { sensor->emit(digitalRead(kDigitalInputPin)); });
```
ba58smith, line 43: Can you give me an example of a true "asynchronous measurement operation" that I can use in an example in a repo in the  Add-ons?

mairas: Check the MAX31856 thermocouple implementation in SensESP 1.x.

PR #499, opened 2021-12-18
Create an intermediate "sensesp" directory in the source tree

mairas:
This is a somewhat big change and at the last moment, but I felt it's better to do it now.

When working on the add-ons, I noticed that the SensESP sensor and transform include paths were clunky and a bit unobvious in the project using them.

For example,
```
#include "sensors/sensor.h"
```
does not indicate where that header is included from.

So, all `src` subdirectories have now moved under a `sensesp` intermediate directory, resulting in includes like:
```
#include "sensesp/sensors/sensor.h"
```

The top-level includes with `sensesp` in their name are still in place:
```
#include "sensesp_app_builder.h"
```

This change, of course, impacts all projects using SensESP, but I believe the change improves the clarity of the include statements.


ba58smith:
I like it - the more clear, the better. But I have reviewed only a couple of the files, just to make sure I knew what you did. Other than that - I'll trust that you got them correct!

PR #505, opened 2021-12-28
WIP: Create a minimal SK example

ba58smith:
@mairas , please comment on this `main.cpp`. If it's OK, then I'll write the accompanying GHPages page to explain it all, as the most basic example that sends data to SK.

I will then add to this example, and write a second tutorial page for it, and keep doing that until we have a decent example that illustrates all of the main concepts.


ba58smith:
I realized this needs to be in a separate repo in the SensESP Organization. Closing.

PR #510, opened 2022-01-07
add DigitalInputDebounceCounter counter

smr547:
Certain devices (e.g. reed switches) are very noisy when they actuate,
generating many interrupts per "event". In these cases, it is necessary to
ignore interrupts that occur very shortly after an initial interrupt.

This class is implemented as a slight specialization of DigitalInputCounter,
but the implementation is kept separate so that DigitalInputCounter may avoid
dynamic dispatch overhead (it needs to be able to run at high speed).

### DIFF
--- a/src/sensesp/sensors/digital_input.cpp
+++ b/src/sensesp/sensors/digital_input.cpp
@@ -1,5 +1,6 @@
 #include "digital_input.h"
 
+#include <elapsedMillis.h>
 #include <FunctionalInterrupt.h>
 
 #include "sensesp.h"
@@ -51,6 +52,47 @@ bool DigitalInputCounter::set_configuration(const JsonObject& config) {
     }
   }
   read_delay_ = config["read_delay"];
+  return true;
+}
+
+void DigitalInputDebounceCounter::handleInterrupt() {
+  if (since_last_event_ > ignore_interval_ms_) {
+    counter_++;
+    since_last_event_ = 0;
+  }
+}
+
+void DigitalInputDebounceCounter::get_configuration(JsonObject& root) {
+  root["read_delay"] = read_delay_;
+  root["ignore_interval"] = ignore_interval_ms_;
+}
+
+static const char DEBOUNCE_SCHEMA[] PROGMEM = R"###({
+    "type": "object",
+    "properties": {
+        "read_delay": { "title": "Read delay", "type": "number", "description": "The time, in milliseconds, between each read of the input" },
+        "ignore_interval": { "title": "Ignore interval", "type": "number", "description": "The time, in milliseconds, to ignore events after a recorded event" }
+    }
+  })###";
+
+String DigitalInputDebounceCounter::get_config_schema() {
+  return FPSTR(DEBOUNCE_SCHEMA);
+}
+
+bool DigitalInputDebounceCounter::set_configuration(const JsonObject& config) {
+  String expected[] = {"read_delay", "ignore_interval"};
+  for (auto str : expected) {
+    if (!config.containsKey(str)) {
+      debugE(
+          "Cannot set DigitalInputDebounceConfiguration configuration: missing "
+          "json field %s",
+          str.c_str());
+      return false;
+    }
+  }
+
+  read_delay_ = config["read_delay"];
+  ignore_interval_ms_ = config["ignore_interval"];
   return true;
 }
 
--- a/src/sensesp/sensors/digital_input.h
+++ b/src/sensesp/sensors/digital_input.h
@@ -3,6 +3,8 @@
 
 #include "sensor.h"
 
+#include <elapsedMillis.h>
+
 namespace sensesp {
 
 /**
@@ -88,15 +90,11 @@ class DigitalInputCounter : public DigitalInput, public SensorT<int> {
  public:
   DigitalInputCounter(uint8_t pin, int pin_mode, int interrupt_type,
                       unsigned int read_delay, String config_path = "")
-      : DigitalInput{pin, pin_mode},
-        SensorT<int>(config_path),
-        interrupt_type_{interrupt_type},
-        read_delay_{read_delay} {
-    load_configuration();
-  }
+      : DigitalInputCounter(pin, pin_mode, interrupt_type, read_delay,
+                            config_path, [this]() { this->counter_++; }) {}
 
   void start() override final {
-    ReactESP::app->onInterrupt(pin_, interrupt_type_, [this]() { this->counter_++; });
+    ReactESP::app->onInterrupt(pin_, interrupt_type_, interrupt_handler_);
 
     ReactESP::app->onRepeat(read_delay_, [this]() {
       noInterrupts();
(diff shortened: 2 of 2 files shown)


src/sensesp/sensors/digital_input.h
```diff
@@ -115,6 +116,70 @@ class DigitalInputCounter : public DigitalInput, public SensorT<int> {
...
+      interrupts();
+      notify();
+    });
+  }
+
+ private:
+  int interrupt_type_;
+  unsigned int read_delay_;
+  unsigned int ignore_interval_ms_;
+  unsigned long last_event_ms_ = 0;
```
ba58smith, line 176: This is VERY trivial, but you don't need the `ms_` at the end of these two new members. `read_delay_` is in ms, and it's not named `read_delay_ms_`. In the implementation above, `now` is in ms, but it's not called `now_ms`. For consistency, how about you make these `ignore_interval_` and `last_event_`?

src/sensesp/sensors/digital_input.cpp
```diff
@@ -54,6 +54,41 @@ bool DigitalInputCounter::set_configuration(const JsonObject& config) {
...
+        read_delay_ = config["read_delay"];
+        changed = true;
+      }
+      if (str == "ignore_interval") {
+        ignore_interval_ms_ = config["ignore_interval"];
+        changed = true;
+      }
+    }
+  }
+  return changed;
```
ba58smith, line 89: Naming the variable "changed" confused me - I think it really represents "found", doesn't it? I don't see anything that would tell you whether or not anything changed - just whether or not it was found.

Second, I don't think the logic is correct. "changed" starts out as false, then if you find "read_delay", changed becomes true. Then, if you DON'T find "ignore_interval", changed is still true, but shouldn't it be false at that point?

src/sensesp/sensors/digital_input.h
smr547: Renaming now to now_ms seems better. The ms prefix is useful, otherwise it's not clear on inspection what those variables represent.

src/sensesp/sensors/digital_input.cpp
smr547: There's nothing in the superclass documentation that says what the return value means, so I took a guess that it meant "this call to set_configuration changed _something_". The superclass documentation could be better in this regard.

smr547: I changed the implementation to be in line with other set_configuration impls with multiple values (e.g. voltage divider transform), hope that looks better now.

smr547:
Thanks for the review. If you still feel it's better to drop the `_ms_` suffix and be consistent with the other counter, please let me know.

I'm really happy with this debounce counter, our rain gauge is solid as a rock nw. :)

src/sensesp/sensors/digital_input.h
ba58smith: OK.

src/sensesp/sensors/digital_input.cpp
ba58smith: Looks good. Thx.

ba58smith:
I have specifically requested @mairas review, and will leave it to him to merge it.

Thanks for the contribution, @smr547!

src/sensesp/sensors/digital_input.h
```diff
@@ -115,6 +116,70 @@ class DigitalInputCounter : public DigitalInput, public SensorT<int> {
...
+    load_configuration();
+  }
+
+  void start() override final {
+    ReactESP::app->onInterrupt(pin_, interrupt_type_, [this]() {
+      unsigned long now_ms;
+
+      if ((now_ms = millis()) - last_event_ms_ > ignore_interval_ms_) {
+        this->counter_++;
+        last_event_ms_ = now;
```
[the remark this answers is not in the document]
mairas: Also, does the `last_event_ms_` part compile? I think I've always had to use `this` explicitly within lambdas: `this->last_event_ms_`...

mairas (commented):
Overall looks good - thanks for the PR! There are a few comments I'd like to get replies for. I have a hunch the code is not compilable as is...

mairas:
> Thanks for the review. If you still feel it's better to drop the `_ms_` suffix and be consistent with the other counter, please let me know and I'll do that.

I think there is actually some value in having those suffixes; given that the time values can be in ms or µs depending on the context, it's probably better to be explicit.

src/sensesp/sensors/digital_input.h
```diff
@@ -115,6 +116,70 @@ class DigitalInputCounter : public DigitalInput, public SensorT<int> {
...
+        read_delay_{read_delay},
+        ignore_interval_ms_{ignore_interval_ms} {
+    load_configuration();
+  }
+
+  void start() override final {
+    ReactESP::app->onInterrupt(pin_, interrupt_type_, [this]() {
+      unsigned long now_ms;
+
+      if ((now_ms = millis()) - last_event_ms_ > ignore_interval_ms_) {
```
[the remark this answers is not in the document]
smr547: Thank you, I didn't know about this one. Done!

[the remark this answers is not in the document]
smr547: Replaced with elapsedMillis.

src/sensesp/sensors/digital_input.h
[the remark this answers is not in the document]
smr547: All fixed now, I think.

src/sensesp/sensors/digital_input.h
```diff
@@ -115,6 +116,70 @@ class DigitalInputCounter : public DigitalInput, public SensorT<int> {
...
+ * @param interrupt_type Will be RISING, FALLING, or CHANGE
+ *
+ * @param read_delay_ms How often you want to read the pin, in ms
+ *
+ * @param ignore_interval_ms Ignore events within this interval after a recorded
+ * event.
+ *
+ * @param config_path The path to configuring read_delay in the Config UI
+ */
+class DigitalInputDebounceCounter : public DigitalInput, public SensorT<int> {
```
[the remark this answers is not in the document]
smr547: Done, I replaced the sibling class with a subclass. The superclass DigitalInputCounter now has a field interrupt_handler_, provided in a protected constructor. Please take a look and see if this is OK.

smr547:
Thanks for the review. I've run the CI script locally and it appears to compile, but it's still untested on actual hardware.

src/sensesp/sensors/digital_input.h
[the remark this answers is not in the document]
mairas: I was thinking of just removing `final` from `void start() override final` and then overriding `start()` in the subclass, but I think your approach may have resulted in less code duplication!

mairas:
Looks good to me. Thank you!

Because of the changes in existing code, I'll make a quick test on real hardware before merging to make sure nothing has been broken.

mairas:
Testing the PR led to quite a long debugging session. Turns out that in some previous `DigitalInputCounter` refactoring, I had accidentally dropped off initialization of `interrupt_type_` in the constructor. I've fixed that in PR #521.

Furthermore, to my slight schadenfreude, you also, in a similar fashion, forgot to initialize `interrupt_handler_` in the new protected constructor. I've rebased this PR and amended the commit with the fix.

Thanks again for the new feature!

Merging.

PR #511, opened 2022-01-07
Add debounce counter v1

smr547:
This change is similar to https://github.com/SignalK/SensESP/pull/510, just back-ported to the V1 API. I've tested this one and it works quite nicely.


smr547:
@mairas I've updated this to be similar to the v2-dev debounce counter and also to use elapsedMillis. I've tested both DigitalInputCounter and DigitalInputDebounce counter on actual hardware and both still work nicely.

mairas:
I'll take your word for it!

PR #517, opened 2022-01-11
Add GNGLL parser, fix parsing of south and west coordinates

sjamesr:
GNGLL sentences are emitted when the fix comes from a combination of GPS
and GLONASS. Before this, the fix would be ignored.

SensESP's NMEA parser incorrectly treats southerly latitudes and
westerly longitudes as positive degrees, this change also fixes that.

### DIFF
--- a/src/sensors/gps.cpp
+++ b/src/sensors/gps.cpp
@@ -9,6 +9,7 @@ GPSInput::GPSInput(Stream* rx_stream)
   rx_stream_ = rx_stream;
 
   nmea_parser_.add_sentence_parser(new GPGGASentenceParser(&nmea_data_));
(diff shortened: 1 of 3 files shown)


mairas (approved):
Thank you for the contribution! All of the NMEA parser codebase is a few years old and I've been itching to get back to it. Probably could use a bit of dustup...

We're currently busy finishing SensESP v2 development, and this is part of functionality that will be dropped out. The intention is to keep the core limited in size and then have all extra stuff (like this) as add-on libraries that you can then include in the project. Haven't done the add-on repo, yet, though.

Digressing a bit more - it might be even worth considering if the parser could be split into a repo of its own with no dependencies to SensESP. That might be complicated wrt. the ObservableValues etc, though.

Anyhoo - this PR looks great!

PR #518, opened 2022-01-11
Fix SKOutputPosition emitting "null"

sjamesr:
Previously, SKOutputPosition seemed to be emitting the string "null".
This patch fixes it, at least for me. Maybe I was doing something wrong before. :)


mairas:
Thanks for the fix! ArduinoJson got updated from 5 to 6 a while back, and it is quite possible no-one has tested or used this code since then. Glad that you fixed it.

mairas:
... aaand, merging.

PR #537, opened 2022-01-31
Hash the configuration file names

### LINKED ISSUE #536: Hash configuration path filenames
In the current implementation, the configuration path of a `Configurable` is used as-is as a SPIFFS filename. This becomes a problem if the configuration path is long because the SPIFFS filename has a path length limitation of 32 characters.

An early development version of the SH-ESP32 Engine Hat firmware has a following path:

```
/analog/A/fuelTank/currentLevel/interpolator
```

The path is 45 character long, far exceeding the limit.

This could be circumvented by hashing the config path with a suitable hash function and using the hash as the filename. MD5, for example, creates a 128-bit hash, which is exactly 32 bytes when converted to ASCII. MD5 as a cryptographic hash has been broken ages ago, but it doesn't matter in this case. Or, we could use a SHA-256 and truncate it to the required length.

Implementation-wise, it should be sufficient to apply the filename hashes in `Configurable::load_configuration` and `Configurable::save_configuration`.

One consequence of this change would be that there would be no limitations for the configuration path format or length.

From an API point of view, this would be a non-breaking change. Individual devices would lose their saved configuration.

mairas:
SPIFFS has a maximum path length of 32 characters. To circumvent this, configuration file paths are hashed with SHA-1 and then base64 encoded.

~This change resets existing device configurations.~ There are checks in place to make it backwards-compatible.

Fixes #536.

### DIFF
--- a/src/sensesp/system/configurable.cpp
+++ b/src/sensesp/system/configurable.cpp
@@ -1,8 +1,8 @@
 #include "configurable.h"
 
-#include "sensesp.h"
-
 #include "SPIFFS.h"
+#include "sensesp.h"
+#include "sensesp/system/hash.h"
 
 namespace sensesp {
 
@@ -39,13 +39,18 @@ void Configurable::load_configuration() {
            config_path.c_str());
     return;
   }
-  if (!SPIFFS.exists(config_path)) {
-    debugI("Not loading configuration: file does not exist: %s",
-           config_path.c_str());
+  String hash_path = String("/") + Base64Sha1(config_path);
+
+  if (!SPIFFS.exists(hash_path)) {
+    debugI("Not loading configuration for path %s: file does not exist: %s",
+           config_path.c_str(), hash_path.c_str());
     return;
   }
 
-  File f = SPIFFS.open(config_path, "r");
+  debugD("Loading configuration for path %s from file %s", config_path.c_str(),
+         hash_path.c_str());
+
+  File f = SPIFFS.open(hash_path, "r");
   DynamicJsonDocument jsonDoc(
       1024);  // TODO: set the size of ALL DynamicJsonDocuments throughout the
               // project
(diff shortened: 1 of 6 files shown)


src/sensesp/system/configurable.cpp
```diff
@@ -39,13 +39,18 @@ void Configurable::load_configuration() {
...
   }
-  if (!SPIFFS.exists(config_path)) {
-    debugI("Not loading configuration: file does not exist: %s",
-           config_path.c_str());
+  String hash_path = String("/") + Base64Sha1(config_path);
+
+  if (!SPIFFS.exists(hash_path)) {
+    debugI("Not loading configuration for path %s: file does not exist: %s",
+           config_path.c_str(), hash_path.c_str());
     return;
```
pshotton, line 47: Wouldn't it be a good idea for backward compatibility to check if the unhashed filename exists - then existing configurations will be preserved? If load succeeds might want to set a flag so that when saved with hashed name the old file is removed.

mairas: It's a good idea. How about first checking for the existence of the hashed filename, and if it doesn't exist, then trying the unhashed one (if it's less than 32 chars long)? Wouldn't need to set any flags.

pshotton: Flag was just to ensure old (non-hashed) file was deleted to save SPIFF space, but I guess you could code it to write out new and delete old in same block.

mairas: My initial reaction would be that the config files are so tiny that it doesn't matter. But given that the SPIFFS size in the partitioning scheme used in SensESP is minuscule, maybe the old files should be deleted, after all.

mairas: Now the unhashed filename is used as a fallback when loading the configuration, and during saving the unhashed filename is deleted if it exists. I've tested that a device retains its configuration even after these changes have been applied.

mairas:
@pshotton maybe you could do a sanity check before I merge?

Coincidentally, this change also allows for easier switching from SPIFFS to LittleFS, if we ever decide to do that.

pshotton (approved):
Looks good

mairas:
Thanks!

PR #538, opened 2022-02-01
Curve interpolator UI improvements

mairas:
`CurveInterpolator` web UI was a bit clunky to use. I changed the editor into table format and added options to provide display names for the input and output sample values. With these tweaks, this is how it looks like:

Screenshot 2022-02-01 at 16 10 45


JohnySeven:
Nice!

mairas:
Did a bit of cleanup on `CurveInterpolator` while at it. Now it's easier to use the class directly without subclassing it. And the titles are set with setters instead of constructor arguments.

PR #557, opened 2022-02-13
Collect UIOutput objects to SensESPApp

mairas:
UIOutput objects were scattered around unrelated classes. They're now all collected to `SensESPApp`.

Additionally, the `ui_outputs` instance map was a global variable. That was changed into a static member variable of UIOutputBase.


mairas:
Meh, this was supposed to be a stacked PR.

PR #579, opened 2022-03-08
WiFi connection logic improvements

mairas:
SensESP WiFi connection logic was a bit fragile and unnecessarily complex. Different connecting code was used depending on whether the WiFI AP was hard-coded or configured via WiFi Manager, and also reconnecting in case of lost WiFi was unreliable. This PR attempts to address those issues.

Major changes:
- WiFi Manager configuration portal is only enabled if no previous configuration has been saved. Once SSID and password have been entered and verified to work, they will be used ever after (or at least until the device is reset).
- If the WiFi AP is not available or the connection is lost, the device will keep attempting to reconnect until a connection has been established. WiFi Manager never gets launched, and these reconnection attempts are non-blocking (so the device is still operating normally, albeit without network).
- WiFi Manager operates still in a blocking fashion, but since it's never launched during normal operation, this is less harmful than before.

### DIFF
--- a/library.json
+++ b/library.json
@@ -40,7 +40,8 @@
     },
     {
       "owner": "alanswx",
-      "name": "ESPAsyncWiFiManager"
+      "name": "ESPAsyncWiFiManager",
+      "version": "^0.31"
     },
     {
       "owner": "bblanchon",
--- a/src/sensesp/controllers/system_status_controller.cpp
+++ b/src/sensesp/controllers/system_status_controller.cpp
@@ -2,21 +2,21 @@
 
 namespace sensesp {
 
-void SystemStatusController::set_input(WifiState new_value,
+void SystemStatusController::set_input(WiFiState new_value,
                                        uint8_t input_channel) {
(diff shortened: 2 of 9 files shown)


ckcrs:
Thanks a lot... Just testing and curios about "device is still operating normally, albeit without network" part.
Should I also expect to see normal data output from esp32 at my usb serial connection?
Sancan
PS: I can see below on by usb serial, till it reconnects to my restarting RPi, then data reporting starts:
"(wifi_station_disconnected)(C1) Disconnected from wifi.
(connect)(C0) WiFi is disconnected. SignalK client connection will connect when WiFi is connected.

mairas:
> Thanks a lot... Just testing and curios about "device is still operating normally, albeit without network" part.
> Should I also expect to see normal data output from esp32 at my usb serial connection?

If you're referring to the Signal K deltas that are printed out when the data is transmitted to the SK server, then no, that data is not printed out as long as the data isn't being transmitted.

PR #581, opened 2022-03-23
Fix the unbalanced namespace brackets when debugging is disabled

mairas:
As reported on Slack by RBO.


mairas:
Reported on Slack to work fine.

PR #588, opened 2022-04-25
Fixes required by platformio/platformio-espressif32 v.4.1.0

mairas:
PlatformIO's espressif32 framework got updated from 3.5.0 to 4.1.0. The major change was that the Arduino ESP32 core got updated from 1.0.6 to 2.0.2. The change broke some APIs and exposed some bugs in SensESP. This PR fixes those.

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -36,7 +36,7 @@ lib_deps =
 
 [espressif32_base]
 ;this section has config items common to all ESP32 boards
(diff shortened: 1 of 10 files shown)


mairas:
Additionally, it seems I had broken support for hardcoded WiFi SSIDs and passwords. That's hopefully fixed in the PR as well.

PR #551, opened 2022-02-08
SensESP's WebUI upgrade vol. 2

JohnySeven:
So this is my 2nd pull request with few enhancements in WebUI area.

This PR add new features and enhances few things in WebUI:
1. Status page was little messy so after talk with @mairas we're introducing UIOutput supports Groups! There few default groups or you can create your own. Here is example:
```
 auto customUIOutput = new UIOutput<int>("UserInfo", "My cool group");

  app.onRepeat(1000, [customUIOutput]()
  {
    int value = customUIOutput->get()+1;
    customUIOutput->set(value);
  });
```
2. Renamed "Advanced" page to "Control" page and I've made few UI enhancements to introduce HTTP commands.
3. HTTPCommand - new class that allows SensESP lib user to add custom commands on "Control" page, user can provide name, button label and define if WebUI should ask user to run command. Then in main.cpp user can add any command easily, here is example:
```
  sensesp_app->get_http_server()->add_command("On", "Turn on", false)
  ->attach([]()
  {
    pwm_value = on_value;
  });
```
4. In last WebUI PR there was house logo on page, this PR brings SensESP logo!
5. I've added new virtual method into `Configurable` class `virtual String get_config_description()` for returning configurable description that will be visible in configuration page.

Screenshots:
screenshot-2022-02-08-19-33-43
screenshot-2022-02-08-19-34-39
image

Jan

PR #593, opened 2022-04-26
UI output groups

mairas:
UI Output item grouping and ordering functionality extracted from the closed PR #551.


src/sensesp/system/ui_output.h
```diff
@@ -1,90 +1,15 @@
 #ifndef _SENSESP_SYSTEM_UI_OUTPUT_H_
 #define _SENSESP_SYSTEM_UI_OUTPUT_H_
-#include <ArduinoJson.h>
 
-#include <functional>
-#include <map>
+/**
```
JohnySeven, line 4: Maybe adding `#warning` directive would be much better as comment will not be visible to build output.

JohnySeven:
Thanks for finishing it up!

src/sensesp/system/ui_output.h
mairas: Good point. Will do.

PR #594, opened 2022-04-27
Implement custom UI Buttons

mairas:
Implementation extracted from the closed PR #551.

`UIButton` objects create button objects on the device web UI "Control" tab. When the button is clicked, the UIButton object observers are notified.


src/sensesp/ui/ui_output.h
```diff
@@ -0,0 +1,90 @@
...
+ private:
+  std::function<T()> value_function_;
+
+ public:
+  UILambdaOutput(String name, std::function<T()> function,
+                 String group = kUIOutputDefaultGroup,
+                 int order = kUIOutputDefaultOrder)
+      : UIOutputBase(name, group, order), value_function_(function) {}
+
+  T get() { return value_function_(); }
```
ba58smith, line 51: A very minor comment - all other getters are named get_something(). Since this returns the function, would it be more clear to be named get_function()?

mairas: The get method actually returns the return value of `value_function_`. In that regard, something like `get_value()` would've been a more suitable name.

I already implemented the change while keeping the original `get()` as a deprecated method to avoid breaking the already published interface, but then it occurred to me that this is not a getter at all because it returns the value of an arbitrary function instead of member variable contents (that's a pretty narrow definition of a getter, but still...). And the `get()` method mimics `ObservableValue::get()` used in `UIOutput`.

PR #597, opened 2022-05-04
Single page configuration

mairas:
All configuration editors are now shown on a single page in the web UI. Additionally, the cards can be sorted by setting the `Configurable` object sort_order property.


ba58smith:
@mairas , when you finish this, can you put a screenshot of the end result here, for reference? (Or somewhere - maybe in the Docs.) Thanks!

mairas:
> @mairas , when you finish this, can you put a screenshot of the end result here, for reference? (Or somewhere - maybe in the Docs.) Thanks!

Good point - updated the documentation regarding the web UI.

PR #600, opened 2022-05-16
Fix for ESP32 variants  ADC width compile error

CaptIgmu:
After upgrading to **_v2.4.1_** of SensESP and **_v4.2.0_** of platformio/esspressif32 I could finally compile for my **_Adafruit QT Py  ESP32-S2_** with only 1 error: setting the ADC width. The ADCs have changed on the newer ESP32 variants, the **_S2_** can have 13 bits of resolution. This fix should now cover the ESP32 and its variants.  Hope you have a successful compile!


mairas:
Thanks for the PR! Sorry for the small delay; I was travelling for the last week and didn't have my laptop with me.

The PR looks good. I don't have any ESP32-S2 boards at hand at the moment, so I'll have to take your word on it. I'm glad that support for that platform is improved, though!

Merging.

PR #601, opened 2022-05-16
WIP: SensESP logo in WebUI

JohnySeven:
Co-Authored-By: @ba58smith 

This PR changes WebUI logo from house icon to SensESP logo.


mairas:
Thanks! A nice addition. Could you make the branding configurable/optional, though? I am already using SensESP in some commercial projects which can't/shouldn't have SensESP branding in the web UI, so hard-coding the logo doesn't work for all use cases.

JohnySeven:
I see, than I need to come up with way to do it, as it's embedded into HTML page.

My approach would be add new HTTP GET request /logo and it would return SVG in HTTP GET and by default it would return default one. But if you call something like  `httpServer->set_logo_image(my_logo_h);` or maybe use builder for that?

Let me change it to work in progress and I'll see how that goes.

Thanks

mairas:
Yeah, a separate HTTP endpoint probably would be needed for the logo as the page itself is gzipped...

If the logo is replaced, I think it should be ensured that the original logo doesn't remain eating memory real estate in the built image. Not sure what the best approach for this would be; `#ifdef` conditional includes are my first thought. There might be better ones, though.

JohnySeven:
Another question is do we want just SVG or support bitmap formats like JPG or PNG?

JohnySeven:
Nope :-) So the MIME type would get into Content-Type of HTTP response and browser will be able to show the image properly. 

Ok, will try to finish this ASAP, little busy right now getting my boat ready!

mairas:
I'm closing this PR due to no activity. Please reopen or create a new one once you are ready to continue working on the topic.

PR #613, opened 2022-06-23
Store auto-discovered SignalK server address & port

iuriaranda:
When the SignalK server is auto-discovered through mDNS, the address and port are now saved into the `WSClient` instance so they can be later retrieved via the `get_configuration` method. Before this they were never stored, so calling the `get_configuration` function always returned an empty Json object


mairas:
Perfect, thank you!

PR #614, opened 2022-06-25
Add support to deserialize position type from Json

iuriaranda:
This teaches the ArduinoJson library how to deserialize Position type data like.

```json
{
    "latitude": 39.482845,
    "longitude": 2.352644,
    "altitude": 0.1
}
```

This allows the use of the `SKValueListener` for this type of data like `navigation.position`.

### DIFF
--- a/src/sensesp/types/position.h
+++ b/src/sensesp/types/position.h
@@ -34,6 +34,18 @@ struct ENUVector {
   float up = kPositionInvalidAltitude;
 };
 
+/**
+ * @brief Adds support in ArduinoJson to deserialize Position type data.
+ *
+ * This is an ArduinoJson custom converter to convert position type data from
+ * JSON into a Postion struct. The JSON data must at least contain "latitude"
+ * and "longitude", "altitude" is optional. ArduinoJson will automatically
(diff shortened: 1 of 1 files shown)


src/sensesp/types/position.h
```diff
@@ -34,6 +34,19 @@ struct ENUVector {
   float up = kPositionInvalidAltitude;
 };
 
+void convertFromJson(JsonVariantConst src, Position &dst) {
```
mairas, line 37: This is a neat feature of ArduinoJson; I didn't know about it.

Functionality looks fine.

For a random reader, the usage of these functions might be a bit unobvious. I'm referring to the fact that they're supposed to be called automagically by ArduinoJson. Also, they don't follow the standard SensESP naming convention (by necessity, obviously). Could you add a short doc comment to describe what these functions are used for, and  by whom?

iuriaranda: Right. I've added documentation on both functions

mairas:
Hi Iuri, thanks for the doc addition - looks brilliant! And sorry about the delay, I've been a bit under the weather for the past couple of days. Merging now.

PR #627, opened 2022-09-29
TaskQueueProducer changes

mairas:
`set()` method now returns a boolean to indicate whether adding the value to the queue was successful or not.

Also, the default constructor now requires a pointer to `ReactESP` to indicate which task context is the consumer. An overloaded constructor without the pointer argument is provided for backwards compatibility.


mairas:
I'm being a naughty boy and will merge this right away.

PR #636, opened 2023-01-19
Use mDNS always unless a server has been configured

### LINKED ISSUE #634: SensESP stuck on old IP when SignalK changes IP
I restarted my PI after connecting the SensESP to my SignalK server and it got a new IP from DHCP.

now the SensESP is not able to connect, and I see the old IP in the SensESP interface:
Screen Shot 2023-01-17 at 6 33 09 PM

First I was blaming it on SignalK to broadcast an outdated IP over mDNS, but using `avahi-resolve` in the network it correctly returns the new one `192.168.8.110`. 

So I assume the SensESP saves the value and doesn't update it. 

Obviously I can configure a fixed IP to the SignalK server, but wanted to leave the issue here as an improvement proposal.

My expectation would be that the SensESP rediscovers the correct IP when being restarted.

mairas:
The WSClient persistence logic now follows the rest of the codebase (I think...):

- If nothing is configured, mDNS is always used.
- If a server is hardcoded, it is used by default.
- If the server address is set through the config page, that address is used.
- If the server address is cleared through the config page, mDNS is used.

Fixes #634.

### DIFF
--- a/src/sensesp/net/ws_client.cpp
+++ b/src/sensesp/net/ws_client.cpp
@@ -80,10 +80,8 @@ WSClient::WSClient(String config_path, SKDeltaQueue* sk_delta_queue,
       Startable(60) {
   this->sk_delta_queue_ = sk_delta_queue;
 
-  preset_server_address_ = server_address;
-  preset_server_port_ = server_port;
-  this->server_address_ = server_address;
-  this->server_port_ = server_port;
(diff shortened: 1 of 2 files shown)


johannbarbie (approved):
state logic of `connect()` easier to follow without the additional local variables.

mairas:
Seems like I forgot to merge this... Oops!

Issue #638, opened 2023-01-27
temperature_sender.cpp in the examples calculates incorrectly


Techstyleuk:
I was looking more at your Temp sender code and couldn't work it out, but I think I found an error, that you may have corrected with the R1 value:
Code:
  const float Vin = 3.3;
  const float R1 = 51.0;
  auto* analog_input = new AnalogInput(36);
  analog_input->connect_to(new AnalogVoltage())
      ->connect_to(new VoltageDividerR2(R1, Vin, "/12V_alternator/temp/sender"))
      ->connect_to(new TemperatureInterpreter("/12V_alternator/temp/curve"))
      ->connect_to(new Linear(1.0, 0.0, "/12V_alternator/temp/calibrate"))
      ->connect_to(
          new SKOutputFloat(sk_path, "/12V_alternator/temp/sk", metadata));

the "analog input" is measured in Bits from 0-1023, this is then translated to Voltage by the "AnalogVoltage" transform, but this calculation gets you the percentage of the max voltage, not the actual voltage, therefore the rest of the calcs don't work out correctly - this had me stumped for a bit but if you change that line to:
Code:
analog_input->connect_to(new AnalogVoltage(Vin,Vin))
then it worked out, the first "Vin" sets the max voltage to the input voltage you have stated above and then the second "Vin" is the multiplier to get it back to voltage from percentage.

In discussions on the slack channel it was pointed out by @ba58smith that it should read:

Code:
  const float Vin = 3.3;
  const float R1 = 51.0;
  auto* analog_input = new AnalogInput(36, 1000, "/temp_analog_in/", Vin);
      ->connect_to(new VoltageDividerR2(R1, Vin, "/12V_alternator/temp/sender"))
      ->connect_to(new TemperatureInterpreter("/12V_alternator/temp/curve"))
      ->connect_to(new Linear(1.0, 0.0, "/12V_alternator/temp/calibrate"))
      ->connect_to(
          new SKOutputFloat(sk_path, "/12V_alternator/temp/sk", metadata));

I will try to make the change

ba58smith:
@Techstyleuk, I just made up that config path in my answer on Slack: `"/temp_analog_in/",`. To match with this example, it should probably be `"/12V_alternator/temp/analog_in/"`.

[...]

PR #639, opened 2023-01-27
Correction to the calculation per issue #638

Techstyleuk:
The code was updated per issue #638.

The original code did not transform the decimal bit into voltage, so the calc did not correctly calculate.  associated descriptions/explanations were also updated.


ba58smith (approved):
Looks good! Approving.

ba58smith (approved):

ba58smith (approved):

PR #649, opened 2023-04-03
Fixed error when saving configuration for FloatThreshold

rogerlittin:
Removed unused `out_range_` parameter from ThresholdTransform.

This PR fixes a bug that prevents the configuration for a FloatThreshold transform from being saved.
It also removes the unused `out_range_` variable from ThresholdTransform, which is always the opposite of `in_range_`.


mairas:
Sorry, I had missed the notification email! Looks good to me.

PR #655, opened 2023-06-15
Button Handler

mairas:
This PR implements a button handler for SensESP. The BOOT button on GPIO 0 detects the following events:

- Short press (less than 1 second): Restart the device
- Long press (more than 1 second): Reset the network settings
- Very long press (more than 5 seconds): Reset the device to factory settings

### DIFF
--- a/docs/pages/features/index.md
+++ b/docs/pages/features/index.md
@@ -6,6 +6,29 @@ nav_order: 40
 
 # Features
 
+## Button Handling
+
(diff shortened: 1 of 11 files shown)


JohnySeven:
Nice, that is something I was thinking about today, I've replaced my RPI 3A+ with CM4 board and changed WiFi SSID. So, I had to erase flash & upload code to get my boat engine SH-ESP working with new SSID!

Also, it would be nice if LED flash interval would change to indicate activation of network reset or factory settings restore. This is how Mikrotik indicates the board will restore factory settings. It's very convenient for the user.

mairas:
Thanks for the comment. I created a new issue for that.

PR #661, opened 2023-06-21
added constant_sensor class

mestafin:
**constant_sensor**
Library for sensESP to add virtual sensors that emit a constant value

** Overview**
This library add a virtual sensors that periodically emit a constant value. 

The virtual sensors are typically used to send a constant value to SignalK. 
Typical example is the capacity of a tank.

The sensors can be connected to all the standar SKOutput consumers.

The value of the sensor can be se via the web configuration page of sensESP;

**Sensor Types:**

    ConstantFloat
    ConstantInt
    ConstantString

**How to Use**

Example: send the water tank capacity every 30 seconds to SignalK:

```
    int send_delay = 30; 
    const char* config_path = "/tanks.water_capacity";
    auto *capacity = new ConstantFloat(send_delay,config_path);
    capacity->connect_to(new SKOutputFloat(sk_path, sk_config_path, sk_metadata));
```

To set or get the the value of the virtual sensor for use in your code:

```
    capacity->set(value);

    value = capacity->get();
```


src/sensesp/sensors/constant_sensor.cpp
```diff
@@ -0,0 +1,63 @@
+#include "constant_sensor.h"
+
+namespace sensesp {
+
+// ..........................................
+// base class for constant sensors
+// ..........................................
+
+ConstantSensor::ConstantSensor(int send_delay, String config_path = "")
```
mairas, line 9: Nitpick/suggestion: maybe use `send_interval` instead of `send_delay` that would convey the repetitive nature better.

src/sensesp/sensors/constant_sensor.h
```diff
@@ -0,0 +1,94 @@
...
+        "type": "object",
+        "properties": {
+            "value": { "title": "Constant Value", "type": "number", "description": "Constant value" }
+        }
+    })###";
+
+// ..........................................
+//  base class for constant value sensors
+// ..........................................
+class ConstantSensor : public SensorT<float> {
```
mairas, line 39: Let's make ConstantSensor a template, that way you don't have define specialized child classes at all.

You could have:
```
template <class T, auto default_value>
class ConstantSensor: public SensorT<T> {
...
  T value_ = default_value;
...
}
```
and you could have ConstantValues for any type whatsoever, without explicit inheritance.

Then, optionally, you could have convenience typedefs:

```
typedef ConstantSensor<float, 0.0> FloatConstantSensor;
typedef ConstantSensor<int , 0> IntConstantSensor;
typedef ConstantSensor<bool, false> BoolConstantSensor;
typedef ConstantSensor<String, ""> StringConstantSensor;
```
for easier use for those who get weirded out by the templating syntax.

src/sensesp/sensors/constant_sensor.h
```diff
@@ -0,0 +1,94 @@
...
+ */
+
+static const char SCHEMA_CONSTANT_SENSOR[] PROGMEM = R"###({
+        "type": "object",
+        "properties": {
+            "value": { "title": "Constant Value", "type": "number", "description": "Constant value" }
+        }
+    })###";
+
+// ..........................................
```
mairas, line 36: Here, you might as well have the comment in Doxygen format
```
/**
 * brief ...
 */
```
I think it works as well as a visual spacer but also provides a bit of documentation for  the auto-generated class docs: https://signalk.org/SensESP/generated/docs/index.html

src/sensesp/sensors/constant_sensor.h
mairas: (The syntax I wrote might be a bit off; if you can't figure it out, just leave a comment and I'll test it myself.)

mairas:
I couldn't initially understand the use case for a constant value sensor but the tank volume example hit the spot!

The PR description with the usage examples is very good; ~it'd be shame if anything happened to it~ it'd be great to have it included in the project documentation. Could you copy-paste it to the doc comment of `ConstantSensor`? Doxygen supports Markdown, so no additional formatting should be required.

https://www.doxygen.nl/manual/docblocks.html

src/sensesp/sensors/constant_sensor.h
mestafin: I have no idea how to do this. My C++ knowledge is very limited

mestafin:
It is already in the @brief comment of the constant_sensor.h source file.

Or am I missing something?

Johan

mairas:
> It is already in the @brief comment of the constant_sensor.h source file.
> 
> Or am I missing something?

I was just thinking about the usage examples; they're really good to have.

src/sensesp/sensors/constant_sensor.h
mairas: That works as well. I see that you've allowed changes in this PR; I'll add the changes here.

mairas:
Your template code was quite close. C++ has a lot of quirks and baggage accumulated over time. One such thing is that template method definitions need to be in the header file and not in the .cpp file. I changed that and did a bit of other cleanups. I'll need to test the code on an actual device before merging.

mairas:
Oh, and there seems to be some build issue in CI, too.

mairas:
There were some minor issues in the class implementation (changes I made) and the example had a few notable omissions. Fixed now, and the value is properly transmitted. Merging.

Thanks, @mestafin, for the PR!

mestafin:
Thanks for your help with the C++ quirks

PR #677, opened 2024-02-08
Implement captive portal

mairas:
Getting a captive portal working was a bit more work than expected. The default ESP-IDF HTTP server uri handlers were not ideal for implementing middleware-type stuff that pre-empt the normal handlers. Thus, a custom handler mechanism was implemented. Captive portal is enabled via the web UI and then conditionally evaluated for each HTTP request. Finally, a DNS server is started if the captive portal is used.


mairas:
Will force-push due to commit history changes.

PR #687, opened 2024-03-23
Add thermocouple_max6675 example

zschramm:
Wanted to provide sample code for reading 2 thermocouples for exhaust (gas) temperature measurement.


mairas:
Thank you! Much appreciated!

Since the example depends on an external library, could you please move the file to a subdirectory of its own and also provide a sample platformio.ini file that has the dependency definition?

zschramm:
Yes I will do Matti.

BTW I figured out the issue with connecting to my victron cerbo signalk
instance.  I had to disable SSL, once I did that the request showed up.

PR #689, opened 2024-04-15
Improved curve interpolator

rszemeti:
The curve interpolator reacted strangely to values beyond the curve limits.

Values below the lower bound were linearly interpolated between the point 0,0 and the first curve point, values above the upper bound translated to 9999.99

It seems more graceful to simply continue the curve linearly above and below the lower and upper pair of points.  

If "hard limits" are desired, such as a tank cannot be less 0.0 empty or 1.0 full simply adding bounds points with the same value at the top and bottom of the list produces a flat line:

eg: 

(50,0)
(100,0)
(1000,1)
(1050,1) 

Willl produce a "flat" curve lowwr than 100 and above 1000,  effectively limiting the output.   Without the bound points, the linear curve extends above 1000 and below 100 without unpredictable results.

### DIFF
--- a/src/sensesp/transforms/curveinterpolator.cpp
+++ b/src/sensesp/transforms/curveinterpolator.cpp
@@ -33,51 +33,66 @@ CurveInterpolator::CurveInterpolator(std::set<Sample>* defaults,
 }
 
 void CurveInterpolator::set_input(float input, uint8_t inputChannel) {
-  float x0 = 0.0;
-  float y0 = 0.0;
-  bool inputTooLow = false;
-  bool inputTooHigh = false;
-
-  std::set<Sample>::iterator it = samples_.begin();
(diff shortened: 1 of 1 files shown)


mairas:
Thanks. I had noticed the weird behavior you described but instead just added those guard elements you mentioned. :-)

As for the fix, though - I'm wondering would it be better to cap the output to the top value? A common use case might be to set a tank sender mapping as follows:

```
(0,0)
(180,1)
```

Here, if the measured tank sender resistance (due to either the sender itself or ADC inaccuracies - HALMET for example uses 1% resistors there) is above 180 ohms, you'll get a tank fill rate above 100%. If we didn't extrapolate but instead returned the top value, the behavior might be "less unexpected" to the user.

What do you think?

PR #692, opened 2024-04-17
Patch 2 - NMEA2000 Exhaust Gas Temperature Output on PGN130316

zschramm:
Added output over NMEA2000


zschramm:
Sorry I saw you just recently committed my first version of this example, but I thought others would find this useful since it took me a bit to figure out which PGN to use.  Tested and working on my boat.

PR #693, opened 2024-04-17
Patch 2 - Add NMEA2000 Output of Exhaust Gas Temperature

zschramm:
Figure this would be welcome by some as it took me a bit to find the correct PGN to output.   Sorry you just recently merged my last update, but this is the speed I work at ;)


mairas:
Thanks!

PR #699, opened 2024-05-25
Use native ESP-IDF logging

mairas:
Instead of the `debugX` logging functions, native ESP-IDF logging facilities are now used (as allowed by the Arduino Core). This allows redirection of logs in the future.


JohnySeven:
Nice! I do support this at least with a comment! 👍

PR #703, opened 2024-06-06
Implement new merging and repeating transforms

mairas:
Merging transforms take multiple producer inputs and merge them into a tuple that can be passed for processing. `Join` emits an output whenever any input is updated, as long as all inputs are below max_age. `Zip`, on the other hand, emits an output once all inputs have been updated, also observing the age of individual inputs.

Repeating transforms keep emitting the input value at given intervals. `Repeat` emits whenever an input is received and keeps repeating the value every `interval` milliseconds. `RepeatStopping` works like `Repeat` but includes a `max_age`. If no input is received within `max_age` milliseconds, it stops repeating. `RepeatExpring` works like `RepeatStopping`, but instead of stopping altogether, it continues repeating `expired_value` once `max_age` is reached. And finally, `RepeatConstantRate` works like `RepeatExpiring`, but always emits at regular intervals, regardless of when new input as been received.


mairas:
I also added a Throttle transform that limits the input rate. Kinda similar to `RepeatConstantRate` but emits immediately and without the repetition.

PR #706, opened 2024-06-16
Remove input_channel argument from ValueConsumer

mairas:
ValueConsumer input_channels were implemented to allow multiple inputs to a single transform. However, the semantics were never fully thought out, and the argument was used very rarely. Since you can achieve the same result by adding multiple LambdaConsumers in a single class, I've decided to remove the input_channel argument altogether. This might break some transforms temporarily.


mairas:
Wrong destination branch.

PR #713, opened 2024-07-25
Stream producers

mairas:
Implement ValueProducers that read individual characters or lines from an existing stream.


mairas:
Wrong target branch.

PR #717, opened 2024-08-02
to avoid using namespace at global scope

offspring:
adding using directive in header files is not a good idea, with this PR

this code
```cpp
            if(!config.containsKey(str)) {
                debugE("N2kFluidLevelSender: Missing configuration key %s",
                       str.c_str());
                return false;
            }

```
will not require
```cpp
using namespace sensesp;`
```
in the include files

https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#sf7-dont-write-using-namespace-at-global-scope-in-a-header-file

### DIFF
--- a/src/sensesp/system/local_debug.h
+++ b/src/sensesp/system/local_debug.h
@@ -4,30 +4,28 @@
 #include "Arduino.h"
 #include "Print.h"
 
-namespace sensesp {
-
 #ifndef DEBUG_DISABLED
 
 #define rdebugA(fmt, ...)        \
-  if (Debug.isActive(Debug.ANY)) \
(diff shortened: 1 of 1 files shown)


mairas:
Good catch!

In the dev-3 codebase, I've taken a bit different approach with the logging/debugging macros: ESP-IDF native ESP_LOGx macros are used as is. The debugX macros are still provided for backwards compatibility - and actually, most of the codebase still uses them. Probably should do a cleanup before the release.

It would be great if you'd have time to do a similar PR against the dev-3 branch as well. But if not, that's OK, the change is pretty simple.

PR #720, opened 2024-08-13
Use ESP-IDF native logging functions

mairas:
Replaced all instances of the `debugX` functions with `ESP_LOGX`.

Also reformatted the files.


src/sensesp/net/networking.cpp
```diff
@@ -156,22 +157,24 @@ void Networking::start_client_autoconnect() {
...
-      debugW(
+      ESP_LOGW(
+          "SensESP",
           "No valid client settings found. Leaving WiFi client disconnected.");
       return;
     }
 
-    debugI("Connecting to wifi SSID %s (connection attempt #%d).",
-           config.ssid_.c_str(), attempt_num);
+    ESP_LOGI("SensESP", "Connecting to wifi SSID %s (connection attempt #%d).",
```
JohnySeven, line 172: @mairas shouldn't "SensESP" be constant somewhere or `#define SENSESP_SOURCE "SensESP"`? It's used many times in the code. I'm sorry for nitpicking, but that bumped in my eyes looking at this. Thanks and have a great day! Hope to see more ESP-IDF compatible SensESP stuff soon!

mairas: You're 100% correct, thanks for the input! Actually, I was just looking at preprocessor macros - showing the actual source file name might actually be even better. `__FILE__` just returns the full input path, though, and that could make the strings a bit wordy. I'll investigate a bit more...

PR #743, opened 2024-09-25
Rename signalk_position to signalk_types

mairas:
ENUVector and AttitudeVector output is also defined


mairas:
Depends on some other stuff, will redo.

PR #761, opened 2024-10-05
Refactor web UI config item C++ implementation

mairas:
This is a biggie. Previously, there was a `Configurable` class that made a config card appear on the web UI whenever something inherited that. However, the implementation made many assumptions on how the web UI items should behave and it tied the web UI definitions and the internal logic tightly together.

This PR decouples object serialization (for saving and web UI purposes), object saving and config item customization. First two are still implemented using class inheritance but config item customization is now done using templates and function overloading.

If you want something to appear in the web UI, you first define overloaded ConfigSchema and ConfigRequiresRestart functions for that class. Then, you call ConfigItem to actually instantiate the ConfigItemT object that is used to provide the REST API data for drawing the UI. This is a somewhat significant change that impacts the end user (developer) interface, but I hope and believe it's a change for the better.


src/sensesp/ui/ui_controls.h
```diff
@@ -25,63 +24,87 @@ class StringConfig : public Configurable {
...
-    load_configuration();
+      : FileSystemSaveable(config_path), value_(value), title_(title) {
+    load();
   }
 
-  virtual void get_configuration(JsonObject& doc) override;
-  virtual bool set_configuration(const JsonObject& config) override;
-  virtual String get_config_schema() override;
+  virtual bool to_json(JsonObject& doc) override;
+  virtual bool from_json(const JsonObject& config) override;
```
mairas, line 66: what is this?

PR #780, opened 2024-10-27
Restore Threshold transform

mairas:
I had removed the Threshold transform during v3 development, but apparently it was used by many people. Let's bring it back!


KEGustafsson:
Threshold working now similarly as earlier on v2.x version.

PR #807, opened 2025-03-09
Link fix for clients that use RepeatExpiring<int>

drott:
Such as engine_load_ in n2k_senders.h of
HALMET example firmware.

### DIFF
--- a/src/sensesp/types/nullable.cpp
+++ b/src/sensesp/types/nullable.cpp
@@ -28,6 +28,8 @@ uint32_t Nullable<uint32_t>::invalid_value_ = 0xffffffff;
 template <>
 int32_t Nullable<int32_t>::invalid_value_ = 0x7fffffff;
 template <>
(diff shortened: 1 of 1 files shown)


drott:
Without this, and with engine parameter transmission in HALMET example firmware on, I get errors such as:

```
.platformio/packages/toolchain-xtensa-esp-elf/bin/../lib/gcc/xtensa-esp-elf/13.2.0/../../../../xtensa-esp-elf/bin/ld: .pio/build/halmet/src/main.cpp.o:(.literal._ZN7sensesp14RepeatExpiringIiE15repeat_functionEv[_ZN7sensesp14RepeatExpiringIiE15repeat_functionEv]+0x0): undefined reference to `_ZN7sensesp8NullableIiE14invalid_value_E'
collect2: error: ld returned 1 exit status
*** [.pio/build/halmet/firmware.elf] Error 1
```

mairas:
Thanks! Somewhat annoying that int32_t isn't picked up automatically in that case...

PR #808, opened 2025-03-09
Add a DigitalInputPcntCounter to count pulses using PCNT units

### LINKED ISSUE #805: Consider adding a `FrequencyCounter` using ESP32's `PCNT` module
Thank you & kiitos to the maintainers for SensEPS and the HALMET example firmware!

For engine RPM counting such as in the HALMET example firmware, a `DigitalInputCounter` is used which gets relatively resource intensive and has a tradeoff in responsiveness/CPU usage. 

In a local hacky approach in developing a HALMET firmware for engine RPM, I am configuring the ESP32's PCNT module, as below directly in `main.cpp`, then add an interval task closure that performs the PCNT reads using `pcnt_get_counter_value(pcnt_unit, &count);` and converts to a frequency. 

The functionality is available in 
`#include "driver/pcnt.h"`

As much as my search goes, it looks like SensEPS does not have a similar functionality. IMHO this would save interrupts and enable measuring higher frequencies without bogging down the CPU. 

```
  pinMode(PCNT_INPUT_SIG_IO, INPUT); // Or INPUT, depending on your setup

  pcnt_config_t pcnt_config = {
    .pulse_gpio_num = PCNT_INPUT_SIG_IO,
    .ctrl_gpio_num = PCNT_PIN_NOT_USED, // No control pin for simple counting
    .lctrl_mode = PCNT_MODE_KEEP,
    .hctrl_mode = PCNT_MODE_KEEP, // These control modes do not matter in this simple counting configuration.
    .pos_mode = PCNT_COUNT_INC, // Count rising edges
    .neg_mode = PCNT_COUNT_DIS, // Disable counting falling edges
    .counter_h_lim = INT16_MAX,
    .counter_l_lim = INT16_MIN,
    .unit = pcnt_unit,
    .channel = pcnt_channel,
  };

  pcnt_unit_config(&pcnt_config);
  pcnt_counter_clear(pcnt_unit);
  pcnt_counter_resume(pcnt_unit);
```

drott:
Using the new PCNT API in ESP IDF 5.x [1], allocate a unit and channel,
configure edge actions and report back to client on a specified interval.

API surface similar to the previous interrupt-based DigitalInputCounter.
    
[1] https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/peripherals/pcnt.html
    
Fixes #805.

### DIFF
--- a/examples/rpm_counter.cpp
+++ b/examples/rpm_counter.cpp
@@ -2,7 +2,7 @@
 
 // #define SERIAL_DEBUG_DISABLED
 
-#include "sensesp/sensors/digital_input.h"
+#include "sensesp/sensors/digital_pcnt_input.h"
 #include "sensesp/signalk/signalk_output.h"
 #include "sensesp/transforms/frequency.h"
 #include "sensesp_app_builder.h"
@@ -51,9 +51,10 @@ void setup() {
   const char* config_path_skpath = "/sensors/engine_rpm/sk";
 
   //////////
-  // connect a RPM meter. A DigitalInputCounter implements an interrupt
-  // to count pulses and reports the readings every read_delay ms
-  // (500 in the example). A Frequency
+  // connect a RPM meter. A DigitalInputPcntCounter implements 
+  // access to a pulse counter peripheral. Every read_delay ms, it
+  // reads the number of pulses that have occurred since the last
+  // read, and reports that number (500ms in the example). A Frequency
   // transform takes a number of pulses and converts that into
   // a frequency. The sample multiplier converts the 97 tooth
   // tach output into Hz, SK native units.
@@ -64,7 +65,7 @@ void setup() {
   // ESP32 pins are specified as just the X in GPIOX
   uint8_t pin = 4;
 
-  auto* sensor = new DigitalInputCounter(pin, INPUT_PULLUP, RISING, read_delay);
+  auto* sensor = new DigitalInputPcntCounter(pin, INPUT_PULLUP, RISING, read_delay);
 
   auto frequency = new Frequency(multiplier, config_path_calibrate);
 
--- a/src/sensesp/sensors/digital_pcnt_input.cpp
+++ b/src/sensesp/sensors/digital_pcnt_input.cpp
@@ -1,0 +1,96 @@
+#include "digital_pcnt_input.h"
+
+namespace sensesp {
+
+DigitalInputPcntCounter::DigitalInputPcntCounter(uint8_t pin, int pin_mode,
+                                                 int interrupt_type,
+                                                 unsigned int read_delay,
+                                                 String config_path)
+    : DigitalInput(pin, pin_mode),
+      Sensor(config_path),
+      read_delay_(read_delay),
+      interrupt_type_(interrupt_type) {
+  load();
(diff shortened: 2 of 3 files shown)


drott:
Locally tested with a FNIRSI DSO-TC2 as a signal generator, a modified HALMET example firmware and a B&G Vulcan 9.

drott:
> Excellent, thank you! I've wanted to do this for a long time but never had time to dive into it.

Thanks for the feedback!

> In your opinion, is there any reason why this shouldn't replace `DigitalInputCounter` outright?

From my point of view, certainly ok if replaces the interrupt-based one. The project is called Sense*ESP* so I think portability to a different MCU is not a concern here, and we can use the facilities provided by the chip and manufacturer SDK. 

As a new contributor, I didn't feel bold enough to start deleting code :) 

Let me know how you would like to proceed. I can update this PR to remove the previous one, or I am happy to send a follow-up change that removes the interrupt-based on + optionally rename this?

mairas:
Let's replace it! My initial reaction would be to update this PR and do it right away but if you'd prefer doing an intermediate PR, let me know and I'll merge this right away.

As for the Sens_ESP_ issue, I have been thinking about splitting the core parts (producer-consumer etc) into a more platform-agnostic library, possibly merging ReactESP to it, and then keeping the sensors, transforms and ESP32-specific networking code in a platform-specific library. But yes, at the moment it's very _ESP_. :-)

drott:
> Let's replace it! My initial reaction would be to update this PR and do it right away but if you'd prefer doing an intermediate PR, let me know and I'll merge this right away.

Looking closer, `DigitalInputDebounceCounter` depends on the interrupt-driven version, and would be more complicated to derive from the PCNT counter (and might not make so much sense). My idea for debouncing based on PCNT would be to register an event callback with the PCNT, then disable the counting temporarily with a timer, then re-enable it. 

Do we have clients for the `DigitalInputDebounceCounter`? In the HALMET example firmware there are none. Do we want to keep it?

mairas:
Ah. I definitely wouldn't want to remove `DigitalInputDebounceCounter` because every time I've thought that *there's no way anyone could be using this* and removed a piece of code, I've received complaints. 😬

I'm thinking that `DigitalInputDebounceCounter` should then be refactored to include the interrupt based counter code. But including that change in this PR would be too much, I think. Let's just merge this PR as-is and then that refactoring can be done separately.

mairas:
Hm, it seems `driver/pulse_cnt.h` is not available for the legacy Arduino core with the really old ESP-IDF version. I think I could just disable the rpm_counter.h test and add a comment to the example about it requiring pioarduino.

drott:
> Hm, it seems `driver/pulse_cnt.h` is not available for the legacy Arduino core with the really old ESP-IDF version. I think I could just disable the rpm_counter.h test and add a comment to the example about it requiring pioarduino.

Yes, the new version is in the 3.0 SDK only, and the previous `pcnt.h` has a deprecation warning. I didn't want to submit something that immediately has a legacy warning, so locally, I modified my HALMET example firmware to be upgraded to the newer SDK and built with that, which works and which I verified by testing on the chip.

PR #837, opened 2026-01-09
chore: replace FastLED dependency with minimal CRGB implementation

mairas:
## Summary

- Replace FastLED library (~48KB flash savings) with minimal CRGB implementation
- Add gamma correction for perceptually linear LED brightness

### FastLED Removal
- Add `src/sensesp/system/crgb.h` with minimal CRGB struct and blend() function
- Remove FastLED from `library.json` dependencies
- Update includes to use the new local crgb.h

### Gamma Correction
- Add `gamma_correct()` function to boost low brightness values
- Apply gamma correction in `SystemStatusLed::show()` and `RGBSystemStatusLed::show()`
- Uses inverse gamma (γ=2.8) to compensate for non-linear human brightness perception

## Motivation

FastLED was overkill for SensESP's simple status LED needs:
- Only used for CRGB type and blend() function
- ~48KB flash footprint
- Causes ESP32-C3 IDF5 compatibility issues (I2S audio code conflicts)
- Potential WiFi/RMT resource conflicts on some platforms

## Test plan

- [x] Build succeeds without FastLED
- [x] LED patterns work correctly (fades, blends, color transitions)
- [x] Gamma correction makes dim colors (especially blue) more visible
- [x] Flash size reduced by ~48KB
- [x] Tested on HALMET device

### DIFF
--- a/src/sensesp/system/crgb.h
+++ b/src/sensesp/system/crgb.h
@@ -6,11 +6,11 @@
 namespace sensesp {
 
 /**
- * @brief Gamma correction lookup table (γ=2.8, inverted).
+ * @brief Gamma correction lookup table (γ=2.2, inverted).
  *
  * Pre-computed values for inverse gamma correction to avoid expensive
  * floating-point math in the hot path. Generated with:
- *   table[i] = round(pow(i/255.0, 1/2.8) * 255)
+ *   table[i] = round(pow(i/255.0, 1/2.2) * 255)
  */
 // clang-format off
 constexpr uint8_t kGammaTable[256] = {
--- a/src/sensesp/system/led_blinker.cpp
+++ b/src/sensesp/system/led_blinker.cpp
@@ -27,7 +27,7 @@ LEDPatternFragment frag_linear_fade(uint32_t duration_ms,
           from_color = crgb;
         }
         last_elapsed_ms = elapsed_ms;
-        if (elapsed_ms >= fade_duration_ms) {
+        if (fade_duration_ms == 0 || elapsed_ms >= fade_duration_ms) {
           crgb = target_color;
           return;
         }
@@ -42,6 +42,10 @@ LEDPatternFragment frag_linear_invert(uint32_t duration_ms, bool reverse) {
   return LEDPatternFragment(
       duration_ms, [duration_ms, reverse](uint32_t elapsed_ms, CRGB& crgb) {
         // Blend the color from current color to inverted
+        if (duration_ms == 0) {
+          crgb = CRGB(255, 255, 255) - crgb;
+          return;
+        }
         if (reverse) {
           crgb = blend(CRGB(255, 255, 255) - crgb, crgb,
                        elapsed_ms * 255 / duration_ms);
@@ -56,6 +60,10 @@ LEDPatternFragment frag_blend(uint32_t duration_ms, const CRGB& target_color,
                               bool reverse) {
   return LEDPatternFragment(duration_ms, [duration_ms, reverse, target_color](
                                              uint32_t elapsed_ms, CRGB& crgb) {
+    if (duration_ms == 0) {
+      crgb = reverse ? crgb : target_color;
+      return;
+    }
     if (reverse) {
       crgb = blend(target_color, crgb, elapsed_ms * 255 / duration_ms);
     } else {


mairas:
## PR Review: Replace FastLED dependency with minimal CRGB implementation

### Summary

This PR successfully removes the FastLED dependency (~48KB savings) and replaces it with a minimal implementation. The overall approach is sound and the code is well-documented. However, I found several issues that should be addressed, ranging from critical (incorrect gamma table values) to minor improvements.

### Issues Found

#### Critical: Incorrect Gamma Correction Lookup Table

**Location:** `src/sensesp/system/crgb.h:14-30`

**Issue:** The gamma table comment states it was generated with `round(pow(i/255.0, 1/2.8) * 255)` (gamma=2.8), but the actual values do not match this formula. Reverse-engineering the table values suggests they were generated with gamma approximately 2.2, not 2.8.

For example:
- `kGammaTable[1]` is 21, but for gamma=2.8 it should be 35
- `kGammaTable[10]` is 59, but for gamma=2.8 it should be 80
- `kGammaTable[100]` is 166, but for gamma=2.8 it should be 183

**Impact:** The gamma correction will produce different brightness curves than documented/expected. The current table provides less boost to low values than intended.

**Suggestion:** Either:
1. Regenerate the table with the correct formula for gamma=2.8
2. Or update the comment to reflect the actual gamma value used (~2.2)

If gamma=2.2 was intentionally chosen (it's the sRGB standard), that's fine but the documentation should match.

#### Minor: Potential Division by Zero

**Location:** `src/sensesp/system/led_blinker.cpp:32` (and similar in lines 44-45, 54)

**Issue:** The expressions `elapsed_ms * 255 / fade_duration_ms` and `elapsed_ms * 255 / duration_ms` will cause division by zero if the duration parameter is 0.

**Suggestion:** Add a guard at the start of each fragment function, e.g.:
```cpp
if (fade_duration_ms == 0) fade_duration_ms = 1;
```
Or document that duration must be > 0.

#### Minor: lerp8 Rounding Behavior at frac=254

**Location:** `src/sensesp/system/crgb.h:109-119`

**Issue:** The lerp8 lambda uses `>> 8` (divide by 256) which means that even at `frac=254`, the result is slightly biased toward the first color. For example, `lerp8(0, 255, 254)` returns 253, not 254.

This is mitigated by the early returns for `amountOfP2 == 0` and `amountOfP2 == 255`, but values very close to the endpoints (like 254) will be slightly off.

**Impact:** Minimal - the visual difference is negligible for LED brightness. This is a common trade-off in 8-bit fixed-point math.

**Suggestion:** No action needed - this is acceptable for the use case.

#### Minor: Integer Overflow Possible with Very Long Durations

**Location:** `src/sensesp/system/led_blinker.cpp:32`

**Issue:** The expression `elapsed_ms * 255` could overflow `uint32_t` if `elapsed_ms` exceeds ~16.8 million (about 4.7 hours). Current patterns use durations under 1 second, so this is not a practical concern.

**Impact:** None for current usage patterns.

**Suggestion:** No action needed unless someone creates extremely long LED patterns.

### Questions

1. Was gamma=2.2 (sRGB standard) intentionally used instead of gamma=2.8? If so, please update the comment in `crgb.h`.

### Positive Notes

- Good decision to use `shared_ptr` for per-instance state in `frag_linear_fade()` instead of static variables - this correctly fixes a bug in the original code where multiple fade fragments would share state.
- The change from `* 256` to `* 255` in blend calculations is correct for the new implementation where `amountOfP2=255` has an early return.
- Well-documented code with clear explanations of the gamma correction rationale.
- The CRGB struct is nicely minimal while covering all necessary functionality.
- Flash savings of ~48KB is significant for embedded systems.

### Pre-existing Issue (Not Introduced by This PR)

**Location:** `src/sensesp/system/system_status_led.cpp:76-79`

The `get_delta_tx_count_consumer()` function uses a static local `LambdaConsumer` that captures `this`. If multiple `BaseSystemStatusLed` instances call this function, they would all share the same consumer pointing to the first instance's `this`. This is a pre-existing bug not introduced by this PR, but worth tracking as a separate issue.

### Verdict

**Conditional approve** - The implementation is solid and well-thought-out. The gamma table discrepancy should be investigated and either fixed or documented correctly before merging. Once that's resolved and CI passes, this is ready to merge.

**CI Status:** 14 checks passing, 3 checks pending (esp32c3 pioarduino builds)

---

*Review performed by Claude Code with fresh eyes, focusing on correctness, thread safety, and edge cases.*

Issue #848, opened 2026-02-27
Feature Request: Add missing metadata fields.


mxtommy:
Hi! Been a while :-) 

Trying out this code: https://github.com/JohnySeven/WaveshareESP32-S3-Relay-SK and I see the data but it's missing `"supportsPut": true` in the metadata. Current SKMetadata class doesn't support it.

KIP at least uses this to filter paths that you can select for usage with a switch. 

Looking at doc seems there's a couple others missing as well: https://demo.signalk.org/documentation/_signalk/server-api/MetaValue.html

```
displayScale
example
supportsPuts
zones
```

If you can't get around to it I can maybe try and do it myself in a few weeks, just been a few years since I've done any C coding... I'm more than a bit rusty 😅 

Thanks!

PR #850, opened 2026-03-01
Add missing metadata fields

mxtommy:
First time using Claude :-)

Implement #848 

## Backward Compatibility
* Existing constructor signature is unchanged
* add_entry emits identical JSON to before when no new fields are set (all new blocks are conditional)
* New members are value types that copy correctly with the default copy constructor

## signalk_metadata.h
* SKAlarmState enum class with 6 values: kNominal, kNormal, kAlert, kWarn, kAlarm, kEmergency
* SKZone struct with optional lower_/upper_ bounds (NAN = not set), message_, and state_
* 5 new public members on SKMetadata: example_, supports_put_, display_scale_lower_, display_scale_upper_, zones_
Private alarm_state_to_string() helper

## signalk_metadata.cpp
* Constructor initializes sentinels: supports_put_ = -1, both scale bounds = NAN
* add_entry() conditionally serializes all 4 new fields
* alarm_state_to_string() maps enum values to the spec's lowercase strings

Usage example:
```
SKMetadata* meta = new SKMetadata("rpm", "Engine RPM");
meta->display_scale_lower_ = 0.0f;
meta->display_scale_upper_ = 8000.0f;
meta->zones_.push_back(SKZone(SKAlarmState::kNominal, "Normal range", 600.0f, 6500.0f));
meta->zones_.push_back(SKZone(SKAlarmState::kAlarm, "Overspeed", 7000.0f));
meta->supports_put_ = 0;  // false
```

### DIFF
--- a/src/sensesp/signalk/signalk_metadata.cpp
+++ b/src/sensesp/signalk/signalk_metadata.cpp
@@ -1,5 +1,7 @@
 #include "signalk_metadata.h"
 
+#include <cmath>
+
 namespace sensesp {
 
 SKMetadata::SKMetadata(const String& units, const String& display_name,
@@ -9,7 +11,10 @@ SKMetadata::SKMetadata(const String& units, const String& display_name,
       units_{units},
       description_{description},
       short_name_{short_name},
-      timeout_{timeout} {}
+      timeout_{timeout},
+      supports_put_{-1},
+      display_scale_lower_{NAN},
+      display_scale_upper_{NAN} {}
 
 void SKMetadata::add_entry(const String& sk_path, JsonArray& meta) {
   JsonObject json = meta.add<JsonObject>();
@@ -34,6 +39,48 @@ void SKMetadata::add_entry(const String& sk_path, JsonArray& meta) {
 
   if (this->timeout_ >= 0.0) {
     val["timeout"] = this->timeout_;
+  }
+
+  if (!this->example_.isEmpty()) {
+    val["example"] = this->example_;
+  }
+
+  if (this->supports_put_ != -1) {
+    val["supportsPut"] = (this->supports_put_ == 1);
+  }
+
+  if (!std::isnan(this->display_scale_lower_) ||
+      !std::isnan(this->display_scale_upper_)) {
+    JsonObject scale = val["displayScale"].to<JsonObject>();
+    if (!std::isnan(this->display_scale_lower_)) {
+      scale["lower"] = this->display_scale_lower_;
+    }
+    if (!std::isnan(this->display_scale_upper_)) {
+      scale["upper"] = this->display_scale_upper_;
+    }
+  }
+
+  if (!this->zones_.empty()) {
+    JsonArray zones_arr = val["zones"].to<JsonArray>();
+    for (const SKZone& zone : this->zones_) {
+      JsonObject zone_obj = zones_arr.add<JsonObject>();
+      zone_obj["state"] = alarm_state_to_string(zone.state_);
+      zone_obj["message"] = zone.message_;
+      if (!std::isnan(zone.lower_)) zone_obj["lower"] = zone.lower_;
+      if (!std::isnan(zone.upper_)) zone_obj["upper"] = zone.upper_;
+    }
+  }
+}
+
+const char* SKMetadata::alarm_state_to_string(SKAlarmState state) {
+  switch (state) {
+    case SKAlarmState::kNominal:   return "nominal";
+    case SKAlarmState::kNormal:    return "normal";
+    case SKAlarmState::kAlert:     return "alert";
+    case SKAlarmState::kWarn:      return "warn";
+    case SKAlarmState::kAlarm:     return "alarm";
+    case SKAlarmState::kEmergency: return "emergency";
+    default:                       return "normal";
   }
 }
 
--- a/src/sensesp/signalk/signalk_metadata.h
+++ b/src/sensesp/signalk/signalk_metadata.h
@@ -2,8 +2,38 @@
 #define SENSESP_SIGNALK_SIGNALK_METADATA_H_
 
 #include <ArduinoJson.h>
+#include <cmath>
+#include <vector>
 
 namespace sensesp {
+
+/**
+ * @brief Alarm state values for Signal K Zone metadata.
+ * @see https://demo.signalk.org/documentation/_signalk/server-api/ALARM_STATE.html
+ */
+enum class SKAlarmState {
+  kNominal,
+  kNormal,
+  kAlert,
+  kWarn,
+  kAlarm,
+  kEmergency
+};
+
+/**
+ * @brief Represents a single Signal K alarm zone within metadata.
+ * @see https://demo.signalk.org/documentation/_signalk/server-api/Zone.html
+ */
+struct SKZone {
+  float lower_;    ///< Lower bound of the zone; NAN if not set
+  float upper_;    ///< Upper bound of the zone; NAN if not set
+  String message_;
+  SKAlarmState state_;
+
+  SKZone(SKAlarmState state, const String& message,
+         float lower = NAN, float upper = NAN)
+      : lower_{lower}, upper_{upper}, message_{message}, state_{state} {}
+};
 
 /**
  * @brief Holds Signal K meta data that is associated with
@@ -28,6 +58,11 @@ class SKMetadata {
   String description_;
   String short_name_;
   float timeout_;
+  String example_;
+  int supports_put_;            ///< -1 = not set, 0 = false, 1 = true
+  float display_scale_lower_;   ///< NAN = not set
+  float display_scale_upper_;   ///< NAN = not set
+  std::vector<SKZone> zones_;
 
   /**
    * @param units The unit of measurement the value represents. See
@@ -49,7 +84,11 @@ class SKMetadata {
              float timeout = -1.0);
 
   /// Default constructor creates a blank Metadata structure
-  SKMetadata() : timeout_{-1} {}
+  SKMetadata()
+      : timeout_{-1},
+        supports_put_{-1},
+        display_scale_lower_{NAN},
+        display_scale_upper_{NAN} {}
 
   /**
    * Adds an entry to the specified meta array that represents this metadata
@@ -59,6 +98,9 @@ class SKMetadata {
    * @param[out] meta The array the metadata entry is supposed to be added to
    */
   virtual void add_entry(const String& sk_path, JsonArray& meta);
+
+ private:
+  static const char* alarm_state_to_string(SKAlarmState state);
 };
 
 }  // namespace sensesp


mairas:
## Code Review Results

This is a multi-persona automated review (six independent reviewer agents, findings merged and deduplicated). The claims that drive the verdict — the stale merge base, the `SKZone` collision, and the `supportsPut` regression — were verified by hand against `origin/main` before posting.

**Scope:** merge-base 642609e04 → PR head 2e24be118 (2 files, +91/−2)
**Intent:** Implement [#848](https://github.com/SignalK/SensESP/issues/848) — add `example`, `supportsPut`, `displayScale`, and `zones` to `SKMetadata`, with unchanged JSON output when the new fields are unset.
**Mode:** interactive

**Reviewers:** correctness, testing, maintainability, project-standards, api-contract, adversarial
- api-contract — the diff adds public members and changes the JSON emitted to Signal K servers
- adversarial — 91 changed non-test lines

Route legend: `safe_auto` = mechanical fix, `gated_auto` = concrete fix that changes behavior and needs a decision, `manual` = needs design work, `advisory` = informational. Since this is a fork branch, every fix falls to the PR author.

### The headline problem: the branch is based on a March snapshot of main

The PR's merge base is 642609e04 (3.2.3-alpha). Current main is 4ce19e310 (3.5.1-alpha), and both changed files now merge-conflict with it (verified with `git merge-tree`). Two of the changes re-implement things main has since shipped, in incompatible ways — see findings 1 and 2. A rebase onto current main is the first step; the two P1 findings dissolve into it.

### P1 -- High

| # | File | Issue | Reviewer | Confidence | Route |
|---|------|-------|----------|------------|-------|
| 1 | `signalk_metadata.h:27` | New `sensesp::SKZone` collides with the `SKZone` that already exists on main in `signalk_metadata_listener.h` (different shape: `has_lower`/`lower`/`has_upper`/`upper`/`state`/`message`, no trailing underscores). Two definitions of one name in one namespace: any firmware including both headers fails to compile, and the rest is an ODR violation. | api-contract | 0.92 | `manual` |
| 2 | `signalk_metadata.h:62` | `supportsPut` re-implemented as `int` tri-state, but main has shipped `bool supports_put_` plus a sixth constructor parameter since v3.3.0. Merging this would break firmware using the six-argument constructor, and flip `if (meta->supports_put_)` on a default-constructed object from false to true (the unset sentinel is −1, which is truthy). The int design also serializes any value other than 1 as `false`. | api-contract, maintainability, adversarial | 1.00 | `manual` |

Suggested resolution for both: rebase, drop the `supportsPut` part entirely (already done on main), and either reuse the existing `SKZone` for outbound zones or give the outbound type a distinct name matching the existing struct conventions.

### P2 -- Moderate

| # | File | Issue | Reviewer | Confidence | Route |
|---|------|-------|----------|------------|-------|
| 3 | `signalk_metadata.cpp:52` | The `displayScale` guard is `\|\|`, so setting one bound emits e.g. `{"displayScale":{"upper":8000}}`. Every branch of the spec's `oneOf` for displayScale requires both `lower` and `upper` (with `additionalProperties: false`), and the server-api type makes both non-optional — a consumer computing `(v - lower) / (upper - lower)` gets NaN. Change the guard to `&&` and document that the bounds are set together. | correctness, api-contract, adversarial, maintainability | 0.98 | `gated_auto` |
| 4 | `signalk_metadata.h:33` | Both `SKZone` bound parameters default to NAN, so `SKZone(state, message)` compiles and serializes a bound-less zone. signalk-server expands a missing lower to −Infinity and a missing upper to +Infinity and picks the first matching zone, so a bound-less zone placed first swallows every value and no later alarm zone ever fires — silently. Drop the defaults on at least one bound, or warn when a zone has neither, and document that zone order matters. | adversarial, maintainability | 0.90 | `manual` |
| 5 | `signalk_metadata.cpp:38` | `add_entry` gains eight conditional branches and none is tested, and the PR's central claim — identical JSON when no new fields are set — is asserted only in the PR body. Current main has a host-run `test/native` suite where serialization tests fit without hardware: pin the pre-change JSON for a default-constructed and a units-only `SKMetadata`, then cover the new shapes (one-bound displayScale, zone with one/zero bounds, each alarm state string). | testing (echoed by all six) | 0.90 | `gated_auto` |
| 6 | `signalk_metadata.cpp:83` | The `default:` arm in `alarm_state_to_string` is unreachable (all six enumerators are handled) and suppresses `-Wswitch`, so a seventh alarm state added later compiles clean and silently serializes as `"normal"`. Drop the arm and return `"normal"` after the switch — same behavior, and the compiler flags missed enumerators. `SKWSClient::get_connection_status` already uses that pattern. | maintainability, testing, project-standards | 0.95 | `safe_auto` |
| 7 | `signalk_metadata.cpp:15` | The three sentinels (`supports_put_`, both scale bounds) are initialized in two separate constructor initializer lists. A future constructor that misses one leaves a float uninitialized, and the isnan guard then emits garbage. Use default member initializers in the class body — the prevailing idiom in this codebase, and what the repo's clang-tidy `modernize` checks target. | maintainability | 0.80 | `safe_auto` |
| 8 | `signalk_metadata.h:103` | `alarm_state_to_string` is `private`, but the class doc's extension mechanism is subclassing `SKMetadata` and overriding `add_entry()`. A subclass emitting zones must duplicate the enum-to-string table, and copies drift from the spec strings. Make it `protected`. | maintainability, correctness, project-standards, api-contract | 0.90 | `safe_auto` |

### P3 -- Low

| # | File | Issue | Reviewer | Confidence | Route |
|---|------|-------|----------|------------|-------|
| 9 | `signalk_metadata.cpp:68` | `zone_obj["message"]` is written unconditionally — the only unguarded field in `add_entry`. A zone built without a message emits `"message":""`, overriding the spec's documented default of "Warning" with blank alarm text. Guard with `isEmpty()` like the sibling fields. | maintainability, project-standards | 0.88 | `safe_auto` |
| 10 | `signalk_metadata.h:61` | `example_` and `zones_` carry no doxygen while the members beside them do, and the metadata section of `docs/pages/concepts/index.md` is unchanged — the worked example for zones and displayScale exists only in this PR's description. Move it into the class doc and the docs page. | maintainability, project-standards | 0.75 | `safe_auto` |
| 11 | `signalk_metadata.cpp:76` | The new code is not clang-format clean: running the repo's formatter rewrites five hunks (switch layout, trailing-comment alignment, constructor wrapping). Run `clang-format -i` on both files before the next push. | project-standards | 0.88 | `safe_auto` |
| 12 | `signalk_metadata.h:27` | CI only compiles examples, and no example sets any of the new fields, so nothing — including the PR body's usage snippet — ever compiles the new API in CI. Setting the new fields in one CI-built example would cover it across the build matrix. | testing | 0.85 | `gated_auto` |
| 13 | server interaction | On every websocket reconnect SensESP re-sends the full meta array, and signalk-server's zones handler tears down and rebuilds the watcher per path on each meta message, resetting its duplicate suppression — so each reconnect can re-fire notifications for every zoned path. On marginal boat WiFi that is alarm re-firing on exactly the paths users care about. Worth knowing about; the fix, if any, is server-side. | adversarial | 0.68 | `advisory` |

### Pre-existing Issues (action required)

These predate the PR and were verified against current main. Follow-up issues are filed, so nothing here blocks this PR.

| # | File | Issue | Reviewer | Action |
|---|------|-------|----------|--------|
| 1 | `signalk_output.h:72` | `set_metadata` copies via `make_shared<SKMetadata>(*meta)`: slices subclasses (dropping their `add_entry` override) and leaks the argument in the `set_metadata(new ...)` idiom, while the constructor path takes ownership. | correctness, maintainability, adversarial | file-issue → [#1064](https://github.com/SignalK/SensESP/issues/1064) |
| 2 | `signalk_metadata.h:65` (main) | `SKMetadata` has a virtual method but no virtual destructor; `shared_ptr<SKMetadata>(meta)` deletes through the base pointer. | correctness | file-issue → [#1064](https://github.com/SignalK/SensESP/issues/1064) |
| 3 | `signalk_delta_queue.cpp:94` (main) | `get_deltas` reads and writes `meta_sent_` outside the semaphore while `reset_meta_send()` writes it under the semaphore from the websocket task; a reconnect during delta build can lose the metadata delta for the whole connection. Matters more once zones ride on metadata. | adversarial | file-issue → [#1065](https://github.com/SignalK/SensESP/issues/1065) |
| 4 | `signalk_metadata.h:48` (main) | `supportsPut` doc doesn't say an `SKPutRequestListener` must also be registered; the flag is never consulted by the PUT handling path, so it advertises a capability the device may not have. | api-contract | file-issue → [#1066](https://github.com/SignalK/SensESP/issues/1066) |

### Coverage

- 35 raw findings from 6 reviewers, deduplicated to 13 in-scope plus 4 pre-existing. Suppressed below the 0.60 confidence gate: 0.
- Residual risks: `SKDeltaQueue::get_deltas` never checks `JsonDocument::overflowed()` or the `serializeJson` result, and zones make the single metadata delta materially larger — heap pressure yields silently truncated JSON. `example_` is typed `String`, so a numeric example serializes as a string. Nothing validates `lower < upper` for zones or displayScale. Sharing one `SKMetadata*` across two `SKOutput`s creates two independent `shared_ptr` control blocks (double delete) — and zones make sharing metadata across paths attractive.
- All 6 reviewers returned results.

---

> **Verdict:** Not ready — rebase required.
>
> **Reasoning:** The branch predates three minor releases; both files merge-conflict with main, `supportsPut` already shipped there as a `bool` (finding 2 would regress it), and the `SKZone` name is now taken (finding 1). After a rebase, the remaining useful scope of [#848](https://github.com/SignalK/SensESP/issues/848) is `example`, `displayScale`, and `zones` — the serialization approach for those is sound, and the fixes below are all small.
>
> **Fix order:** rebase onto main and drop the `supportsPut` portion → resolve the `SKZone` collision → `&&` guard for displayScale (finding 3) → zone guards (findings 4, 9) → serialization tests in `test/native` pinning the backward-compat JSON (finding 5) → mechanical cleanups (findings 6–8, 10–12).

mairas:
Rebasing this in place was not possible: the head branch is your fork's `main`, and the branch now conflicts with upstream main in both files. I have rebased your commit onto current main — authorship preserved — and opened https://github.com/SignalK/SensESP/pull/1067 to carry it forward.

Conflict resolution, in short: the `supportsPut` portion was dropped because main has shipped `bool supports_put_` (with a constructor parameter) since v3.3.0, and the zone struct is renamed to `SKMetadataZone` because `sensesp::SKZone` is now taken by `signalk_metadata_listener.h`. Everything else is your code as written. The remaining findings from the review comment above carry over to the new PR.

Thanks for the contribution — the follow-up will land under your authorship. If you'd rather keep working on it yourself, the rebased branch is `feat/metadata-fields` on this repo; for future PRs, a topic branch instead of the fork's `main` avoids this situation entirely.

PR #864, opened 2026-03-20
fix(signalk): fix detect_ssl() crash and harden websocket send calls

### LINKED ISSUE #862: fix(signalk): websocket send calls use portMAX_DELAY, can block forever
## Problem

Three `esp_websocket_client_send_text` call sites use `portMAX_DELAY` as the timeout:

- `subscribe_listeners()` (signalk_ws_client.cpp, around line 279)
- `on_receive_put()` (around line 439)
- `sendTXT()` (around line 453)

If the WebSocket connection silently dies, these calls block the `SKWSClient` task forever — the task never recovers.

The `send_delta()` path was recently fixed in #860 to use `kWsSendTimeoutTicks` (5 seconds) with error handling that calls `restart()`. The same pattern should be applied to the other three call sites.

## Suggested fix

Replace `portMAX_DELAY` with `kWsSendTimeoutTicks` at all three sites and add the same error-handling pattern used in `send_delta()` (log a warning, call `restart()`).

### LINKED ISSUE #863: fix(signalk): potential use-after-free on client_ after restart() from send_delta()
## Problem

`send_delta()` runs in `ExecuteWebSocketTask` every 5ms. When a send fails, it calls `restart()`, which calls `esp_websocket_client_stop()` and `esp_websocket_client_destroy()` on `client_`, then sets it to `nullptr`.

However, the websocket event handler (`websocket_event_handler`) runs in the `esp_websocket_client`'s internal task context. After `restart()` destroys the client, the internal task might still be delivering a final event (e.g., `WEBSOCKET_EVENT_DISCONNECTED`), accessing the already-destroyed client handle.

`subscribe_listeners` and `on_receive_put` are called from the event handler, so they could execute while `client_` is being destroyed.

## Impact

Potential use-after-free crash, likely intermittent and hard to reproduce. More likely to trigger now that `send_delta()` calls `restart()` on timeout (since #860).

## Suggested approach

- Unregister the event handler before destroying the client in `restart()`
- Or use a state flag checked in the event handler to skip processing after restart is initiated
- Or ensure `esp_websocket_client_stop()` guarantees no more events are delivered before returning (check ESP-IDF docs)

mairas:
## Summary

- Fix double-free heap corruption crash in `detect_ssl()` when server returns HTTP→HTTPS redirect
- Fix broken SSL redirect detection (`esp_http_client_get_header` reads request headers, not response headers)
- Replace `portMAX_DELAY` with finite timeout on all remaining `esp_websocket_client_send_text` calls
- Fix use-after-free race in `restart()` by setting disconnected state before destroying the client

## Context

When a Signal K server uses TLS termination with HTTP→HTTPS redirect on the same port (e.g., via Traefik), `detect_ssl()` crashes with `assert failed: multi_heap_free multi_heap_poisoning.c:279 (head != NULL)`. Tested on ESP32-C3 (HALSER board).

The crash had two root causes:
1. `esp_http_client_get_header("Location")` always returned NULL because it reads *request* headers, not response headers. This caused the redirect branch to clean up the client, then fall through to a second cleanup — **double free**.
2. Even if the header was found, the original code read it after `esp_http_client_cleanup()` — **use-after-free**.

The fix uses an `HTTP_EVENT_ON_HEADER` event handler to correctly capture the `Location` response header.

While investigating, two additional pre-existing issues were identified and fixed:
- Three websocket send calls used `portMAX_DELAY`, which can block the task forever if the connection dies silently (closes #862)
- `restart()` destroyed `client_` before setting the disconnected state, creating a race with event handler callbacks (closes #863)

## Test plan

- [x] Build for ESP32-C3 (HALSER board) — succeeds
- [x] Flash and verify `detect_ssl()` correctly detects HTTP 301→HTTPS redirect
- [x] Verify TOFU TLS handshake succeeds with Signal K server behind Traefik proxy
- [x] Verify no crash on repeated connection cycles
- [ ] Verify websocket data flow after Signal K server approves access request

### DIFF
--- a/src/sensesp/signalk/signalk_ws_client.cpp
+++ b/src/sensesp/signalk/signalk_ws_client.cpp
@@ -23,7 +23,8 @@
 
 namespace sensesp {
 
-constexpr int kWsClientTaskStackSize = 8192;  // Reduced from 16KB to save heap
+constexpr int kWsClientTaskStackSize = 8192;  // Stack for ExecuteWebSocketTask
+constexpr int kWsTransportTaskStackSize = 6144;  // Stack for esp_websocket_client internal task
 constexpr TickType_t kWsSendTimeoutTicks = pdMS_TO_TICKS(5000);
 
 SKWSClient* ws_client;
@@ -987,7 +988,7 @@ void SKWSClient::connect_ws(const String& host, const uint16_t port) {
   // Configure WebSocket client
   esp_websocket_client_config_t config = {};
   config.uri = url.c_str();
-  config.task_stack = 6144;
+  config.task_stack = kWsTransportTaskStackSize;
   config.buffer_size = 1024;
   if (auth_header.length() > 0) {
     config.headers = auth_header.c_str();


mairas:
## Review Response

### Summary
Good review with thorough analysis. One finding fixed, four acknowledged as valid but out of scope for this PR.

### Accepted

- **Finding 2: Magic number 6144 / stale comment on kWsClientTaskStackSize**
- **Reason**: Legitimate readability and maintainability issue. Two different task stacks controlled by a named constant and a magic number respectively, with a misleading comment, is confusing.
- **Action**: Added `constexpr int kWsTransportTaskStackSize = 6144` and replaced the magic number at the config site. Fixed the stale comment on `kWsClientTaskStackSize` to describe what it is ("Stack for ExecuteWebSocketTask") rather than its change history. Pushed as a separate commit.

### Acknowledged (valid, out of scope)

- **Finding 1: restart() race / TOCTOU on task_connection_state_ and client_**
- **Reason**: Correct analysis. The `task_connection_state_` variable is accessed from multiple FreeRTOS tasks without synchronization, and the check-then-send pattern has an inherent TOCTOU gap. This PR narrows the window (setting state before stopping the client) but does not close it. However, this is a pre-existing architectural issue -- the lack of a mutex protecting state+client as a unit predates this PR entirely. Fixing it properly requires introducing a mutex around all check-and-send sequences and all state transitions, which is a larger refactor that should not be bundled into this bug fix PR.

- **Finding 3: No evidence 6KB stack is safe**
- **Reason**: Fair point. The 6KB value was chosen based on reasoning about the task's responsibilities (event dispatch, not TLS) rather than measured high-water marks. Adding `uxTaskGetStackHighWaterMark` logging is a good idea for validation but is development instrumentation, not a fix for this PR. The previous default (likely 4KB from ESP-IDF) was causing issues, and 6KB provides headroom. If stack overflow is observed, the constant is now clearly named and easy to adjust.

- **Finding 4: subscribe_listeners() and on_receive_put() silently swallow send failures**
- **Reason**: Valid asymmetry. `send_delta()` calls `restart()` on failure while these paths only log. The reasoning is that `subscribe_listeners()` runs during `on_connected()` (in the WS event task), and calling `restart()` from within the event handler callback could create reentrancy issues with the WS client. Similarly, `on_receive_put()` runs in the event handler context. The `send_delta()` path is different because it runs in `ExecuteWebSocketTask`. Making the error handling consistent requires careful thought about which task context each path executes in. Pre-existing issue, not introduced by this PR.

- **Finding 5: portMAX_DELAY in take_received_updates_semaphore()**
- **Reason**: Acknowledged as out of scope by the reviewer themselves. Agreed -- the semaphore timeout and the confusing `timeout_ms=0` meaning "infinite" are worth addressing but are pre-existing and unrelated to the SSL detection fix.

### Not Acted On (minor/no-fix-needed)

- **Finding 6** (unnecessary strcasecmp calls): Reviewer correctly noted this is negligible.
- **Finding 7** (String allocation in event handler): Reviewer correctly noted this fails safe.
- **Finding 8** (stale comment): Fixed as part of finding 2.

PR #939, opened 2026-04-05
Add "supportsPut" to metadata

### LINKED ISSUE #936: smart switch example not working with KIP
I can't get the smart switch example (smart_switch_example.cpp) working with SignalK/KIP.
The path is not shown as PUT Support and thus not selectable in the path settings of the widget switch panel.

If I try the same path from Node-Red PUT is shown and everything (on SignalK) is working fine.

Bug or feature :-)

Is the example relevant with SensESP v3?

joelkoz:
The "supportsPut" property is needed by devices that use an SKPutRequestListener to allow other devices to know the path can be updated via a PUT request. Fixes #936

### DIFF
--- a/examples/smart_switch/smart_switch_example.cpp
+++ b/examples/smart_switch/smart_switch_example.cpp
@@ -110,8 +110,9 @@ void setup() {
   // or not it has changed.  That keeps the value on the server fresh and
   // lets the server know the switch is still alive.
   // To inform other devices that this is a switch that supports "PUT" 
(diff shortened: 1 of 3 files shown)


joelkoz:
Hold off on this.  There is another issue - boolean values are not supposed to have a "units" field specified at all.  Right now, there is no way to define a meta data structure in SensESP that has NO units field.  I need to make another fix for that. I'll do it in this branch

mairas:
LGTM, and Claude only complained about existing smells.

PR #940, opened 2026-04-07
refactor(net): transport-agnostic networking + ESP32-P4 Ethernet (closes #849)

### LINKED ISSUE #849: RFC: Transport-Agnostic Networking and Source Identification for SensESP
# RFC: Transport-Agnostic Networking and Source Identification for SensESP

**Version:** v1
**Status:** Draft — seeking feedback before implementation
**Related issues:** [#334](https://github.com/SignalK/SensESP/issues/334), [#738](https://github.com/SignalK/SensESP/issues/738), [#680](https://github.com/SignalK/SensESP/discussions/680), [#813](https://github.com/SignalK/SensESP/issues/813)
**Related Signal K server:** [signalk-server#2162](https://github.com/SignalK/signalk-server/issues/2162), [signalk-server#2379](https://github.com/SignalK/signalk-server/pull/2379)

---

## Summary

This RFC proposes two related improvements to SensESP:

1. **Source identification** — SensESP devices currently identify
   themselves to the Signal K server with only a hostname. This makes
   it hard to distinguish multiple devices or build stable source
   priority configurations. This RFC adds a persistent hardware
   identifier (MAC address) and device type to delta messages and
   access requests, directly addressing the gap @tkurki identified in
   [signalk-server#2162](https://github.com/SignalK/signalk-server/issues/2162).

2. **Transport-agnostic networking** — lift the networking layer from
   its WiFi-specific implementation to the unified Network API in
   Arduino ESP32 3.x. This makes SensESP transport-agnostic, enabling
   wired Ethernet, WiFi, and future transports (PPP/cellular) through
   a single, cleaner abstraction.

Source identification is a small, self-contained change (~10 lines)
that can land first. The network abstraction builds on it and is
additive — existing WiFi sketches compile and behave identically.

---

### The community has been asking for this

In [issue #334](https://github.com/SignalK/SensESP/issues/334) (December
2020), @joelkoz proposed modularizing the communications layer. @mairas
responded with a clear vision:

> *"WifiConnectionState would be NetworkConnectionState and would be
> applied to wired ethernet, CAN bus, NMEA 0183 (RS-422) and RS-485
> as well."*

In [discussion #680](https://github.com/SignalK/SensESP/discussions/680),
users asked for Bluetooth and serial connections as alternatives to WiFi.
The workaround today is to use `SensESPMinimalApp` and manually wire the
delta queue — it works, but it is not ergonomic.

In [issue #813](https://github.com/SignalK/SensESP/issues/813), users
requested deep sleep support — which is much simpler when the networking
layer is modular and WiFi is not assumed to be always-on.

### Signal K server needs better source identification from SensESP

In [signalk-server#2162](https://github.com/SignalK/signalk-server/issues/2162),
@tkurki identified a concrete problem with SensESP devices:

> *"One particular problem is (SensESP) sensors sending data over WebSocket,
> where we need better UI to identify sensors. This also ties to
> authentication: it is not the ws connection nor ip address that we can
> rely on."*

[PR #2379](https://github.com/SignalK/signalk-server/pull/2379) (source
priority UX rework) explicitly lists "SensESP / WebSocket sensor
identification" as a follow-up item.

Today, `signalk_delta_queue.cpp` sends exactly one field to identify
the device:

```cpp
source["label"] = SensESPBaseApp::get_hostname();
```

That is the only identification. When a user has multiple SensESP
devices reporting to the same server, the server has no reliable way to
tell them apart or to build a stable source priority configuration
around them.

For comparison, N2K devices include a CAN Name — a 64-bit identifier
with manufacturer, device class, and unique serial. It would be great
if SensESP could provide a similar level of identification over
WebSocket.

The good news is that this fits naturally into the network abstraction
work proposed here. The `NetworkInterface` API already exposes a stable
MAC address that can serve as a persistent device identifier regardless
of transport.

### The platform shift already happened

[Issue #738](https://github.com/SignalK/SensESP/issues/738) tracks
migration to ESP-IDF 5.x / Arduino 3.x via pioarduino. [PR #801](https://github.com/SignalK/SensESP/pull/801) merged pioarduino support, and
`pioarduino_esp32` is now the default build environment. This means
SensESP already runs on Arduino ESP32 3.x, which provides:

- **`NetworkInterface`** — a common base class for WiFi STA, WiFi AP,
  Ethernet, and PPP, with methods like `connected()`, `hasIP()`,
  `localIP()`, `macAddress()`, `setHostname()`
- **`Network.onEvent()`** — a unified event system covering all interface
  types (`ARDUINO_EVENT_ETH_GOT_IP`, `ARDUINO_EVENT_WIFI_STA_GOT_IP`, etc.)
- **`Network.isOnline()`** — a single call that checks whether *any*
  interface has a routable IP address
- **`NetworkClient` / `NetworkServer` / `NetworkUDP`** — transport-agnostic
  socket classes that work identically over WiFi, Ethernet, or PPP

SensESP's higher-level components (`esp_websocket_client`,
`esp_http_client`, mDNS, OTA) already operate at the TCP/IP layer and
are transport-agnostic. The only coupling is in the networking setup and
a handful of `WiFi.*` status calls.

### Real hardware exists and is waiting

There is a growing family of ESP32 boards with Ethernet and PoE
designed for exactly this kind of embedded IoT use:

| Board | Ethernet PHY | PoE | Notes |
|-------|-------------|-----|-------|
| [Aptinex IsolPoE ESP32](https://aptinex.com/product/aptinex-isolpoe-esp32-prodev-kit-with-isolated-poe/) | LAN8720A | Isolated 12W | USB-C, 15 GPIOs |
| [Olimex ESP32-POE-ISO](https://www.olimex.com/Products/IoT/ESP32/ESP32-POE-ISO/open-source-hardware) | LAN8710A | IEEE 802.3af | Open hardware, industrial |
| [wESP32](https://wesp32.com/) | LAN8720A | IEEE 802.3at 13W | Purpose-built for wired IoT |
| [Olimex ESP32-GATEWAY](https://www.olimex.com/Products/IoT/ESP32/ESP32-GATEWAY/open-source-hardware) | LAN8710A | No | Low cost, Ethernet + WiFi |

These boards are well-suited for marine environments where wired
connections are more reliable than WiFi, and PoE eliminates the need for
separate 12V power wiring.

---

### The core insight

SensESP currently touches `WiFi.*` in only **3 source files** and uses
the result in **7 files total**. The WebSocket client, HTTP server, mDNS,
and OTA all operate on the TCP/IP stack and do not care about the
underlying transport. The change surface is small.

### Proposed architecture

**Source identification (addresses signalk-server#2162)**

Enrich the delta `source` object in `SKDeltaQueue::get_delta()` and
the access request in `SKWSClient::send_access_request()`:

```cpp
// Delta source object — before
source["label"] = SensESPBaseApp::get_hostname();

// Delta source object — after
source["label"] = SensESPBaseApp::get_hostname();
source["type"] = "SensESP";
source["src"] = WiFi.macAddress();  // stable hardware ID

// Access request — before
doc["description"] = "SensESP device: " + hostname;

// Access request — after
doc["description"] = "SensESP device: " + hostname;
doc["deviceId"] = WiFi.macAddress();
doc["firmwareVersion"] = SENSESP_VERSION;
```

The MAC address provides a stable identifier that does not change
across reboots, firmware updates, or hostname changes — comparable to
the CAN Name that N2K devices use. The server can use `source.src` as
a persistent key for source priority configuration.

Initially this uses `WiFi.macAddress()` directly. Once the network
abstraction is in place, it switches to `active_interface->macAddress()`
so it works across WiFi, Ethernet, and future transports.

The list of SK paths a device emits is already available via the
existing metadata mechanism (`SKEmitter::get_sources()` /
`add_metadata()`). No changes needed there.

This is ~10 lines of new code, but it directly addresses what @tkurki
identified as a gap and makes SensESP devices proper participants in
the source priority system alongside N2K and NMEA 0183 sources.

**Network state (replaces `WiFiState` / `WiFiStateProducer`)**

A new `NetworkStateProducer` uses `Network.onEvent()` to listen for
`*_GOT_IP` and `*_DISCONNECTED` events from any interface. It emits a
generalized `NetworkState`:

```cpp
enum class NetworkState {
  kDisconnected,       // no interface has an IP
  kConnected,          // at least one interface has a routable IP
  kWifiAPModeActivated // WiFi AP active (provisioning mode)
};

// Backward compatibility
using WiFiState = NetworkState;
```

This replaces `WiFiStateProducer` which currently listens only to
`ARDUINO_EVENT_WIFI_*` events. The new producer responds to Ethernet,
WiFi, and PPP events uniformly.

**Interface provisioning (split from `Networking`)**

The current `Networking` class handles five concerns simultaneously:
WiFi STA connection, WiFi AP provisioning, WiFi scanning, captive portal
DNS, and configuration persistence. This RFC proposes splitting it:

- **`WiFiProvisioner`** — all existing WiFi functionality: STA connection,
  AP mode, captive portal, scanning, credential storage. No behavior
  change for existing users.
- **`EthernetProvisioner`** — calls `ETH.begin()` with user-supplied pin
  configuration. Supports DHCP (default) and static IP. Approximately
  50 lines of code.

Both provisioners emit their state into the shared `NetworkStateProducer`.
They are independent modules — you can use one, both, or neither.

**Signal K connection (clean up `SKWSClient`)**

The single WiFi gate in `signalk_ws_client.cpp` line 543:

```cpp
// Before
if (!WiFi.isConnected() && WiFi.getMode() != WIFI_MODE_AP) {

// After
if (!Network.isOnline()) {
```

No other changes needed. `esp_websocket_client` and `esp_http_client`
are already transport-agnostic.

**Status and diagnostics**

Status page items and system info sensors switch from `WiFi.localIP()` /
`WiFi.RSSI()` / `WiFi.macAddress()` to the interface-agnostic equivalents
provided by `NetworkInterface`:

```cpp
interface->localIP()     // works for WiFi and Ethernet
interface->macAddress()  // works for WiFi and Ethernet
interface->connected()   // works for WiFi and Ethernet
```

RSSI is only shown when WiFi is the active interface. Ethernet mode shows
link status instead.

**Builder API**

The `SensESPAppBuilder` gains new methods alongside the existing ones:

```cpp
// Existing (unchanged)
builder.set_wifi_client("SSID", "password");
builder.set_wifi_access_point("SensESP", "configure");

// New
builder.set_ethernet(ETH_PHY_LAN8720, 0, 23, 18, 17, ETH_CLOCK_GPIO17_OUT);

// Dual mode: WiFi AP for provisioning + Ethernet for data
builder.set_wifi_access_point("SensESP", "configure");
builder.set_ethernet(ETH_PHY_LAN8720, 0, 23, 18, 17, ETH_CLOCK_GPIO17_OUT);
```

### Files affected

| File | Change | Nature |
|------|--------|--------|
| `net/wifi_state.h` | Rename to `network_state.h`, generalize enum, add backward compat alias | Rename |
| `net/networking.h/.cpp` | Rename to `wifi_provisioner.h/.cpp`, extract `NetworkStateProducer` | Split |
| `net/network_state_producer.h` | **New** — ~40 lines, listens to `Network.onEvent()` | New |
| `net/ethernet_provisioner.h/.cpp` | **New** — ~50 lines, `ETH.begin()` + config | New |
| `controllers/system_status_controller.h` | Accept `NetworkState` instead of `WiFiState` | Trivial |
| `signalk/signalk_ws_client.cpp` | Line 543: `Network.isOnline()`; enrich access request with `deviceId` | Minor |
| `signalk/signalk_delta_queue.cpp` | Add `type` and `src` (MAC) to delta source object | Minor |
| `net/http_server.h/.cpp` | Remove `#include "WiFi.h"`, captive portal only when WiFi AP active | Minor |
| `sensors/system_info.h/.cpp` | Use `NetworkInterface*` for IP/MAC | Minor |
| `sensesp_app.h` | Hold `NetworkStateProducer`, optional provisioner(s), adapt status items | Moderate |
| `sensesp_app_builder.h` | Add `set_ethernet()`, keep existing WiFi API unchanged | Additive |
| `net/web/app_command_handler.*` | WiFi scan handlers only registered when `WiFiProvisioner` present | Minor |
| `platformio.ini` | Add env for Ethernet-capable boards, consider dropping legacy `[arduino]` env | Config |

The net code delta is approximately +100 lines of new logic
(`NetworkStateProducer` + `EthernetProvisioner`) with simplification
elsewhere. Existing WiFi behavior is preserved — this is additive.

---

### Immediate benefits

- **Reliable multi-device identification** — users running multiple
  SensESP devices can distinguish them on the server and configure
  stable source priorities, just like N2K devices.
- **Wired Ethernet with PoE** — single-cable installations in engine rooms,
  mast bases, lazarettes. More reliable than WiFi in electrically noisy
  marine environments.
- **Ethernet + BLE coexistence** — with WiFi disabled, BLE gets exclusive
  access to the ESP32's shared RF module. No coexistence issues, better
  range. Ideal for BLE sensor gateways (Victron SmartShunt, Ruuvi tags,
  Xiaomi sensors) feeding data into Signal K.
- **Simpler wired installations** — no SSID configuration, no WiFi
  password management, no connectivity drops. Plug in Ethernet, get DHCP,
  connect to Signal K.
- **WiFi AP for config + Ethernet for data** — provision via phone, run
  via wire. Best of both worlds.

### Architectural benefits

- **Cleaner separation of concerns** — provisioning (how you get on the
  network) is separated from connectivity (whether you are on the network)
  is separated from communication (talking to Signal K). Each can evolve
  independently.
- **Easier to add future transports** — PPP for cellular modems
  (offshore/bluewater boats with satellite or LTE), serial-over-USB, or
  even CAN-based connections. Each is just a new provisioner.
- **Simpler deep sleep support** ([#813](https://github.com/SignalK/SensESP/issues/813))
  — when the networking layer is modular, wake-connect-send-sleep cycles
  are much easier to implement without WiFi assumptions baked into the
  core.
- **Fulfills the vision of [#334](https://github.com/SignalK/SensESP/issues/334)**
  — the communications layer modularization that has been on the roadmap
  since 2020.

### Community benefits

- **Lowers the barrier for PoE board users** — a growing segment of ESP32
  hardware that currently cannot use SensESP without custom workarounds.
- **Marine-grade reliability** — wired Ethernet is standard practice in
  professional marine electronics (NMEA networks are wired for good
  reason). It would be nice if SensESP could offer this too.
- **New use cases** — BLE-to-SignalK gateways, PoE-powered remote sensor
  nodes, multi-interface installations.

---

## What does NOT change

- **Existing WiFi user code** — `set_wifi_client()`, `set_wifi_access_point()`,
  `enable_ota()`, all system info sensors. Existing sketches compile and
  run without modification.
- **Signal K protocol handling** — `SKWSClient`, `SKDeltaQueue`,
  `SKOutput`, `SKListener`. All unchanged except for richer `source`
  metadata in delta messages (additive, backward compatible).
- **Sensor framework** — transforms, consumers, producers. All unchanged.
- **Web UI** — configuration interface continues to work. WiFi scan page
  only appears when WiFi provisioner is active.
- **Build system** — pioarduino remains the default. Board-specific envs
  are additive.

---

## Open questions

1. **Drop the legacy `[arduino]` env (`espressif32 @ ^6.9.0`)?**
   Arduino ESP32 2.x does not have `NetworkInterface` or the unified
   Network API. Supporting it would require a compatibility shim. Given
   that pioarduino is already the default and the official PlatformIO
   ESP32 platform is stalled, is it time to make Arduino 3.x the minimum
   requirement?

2. **Versioning** — this is a breaking change for the `Networking` class
   name and `WiFiState` enum (though backward-compat aliases are
   proposed). Should this be SensESP 4.0, or can it land in 3.3.0 with
   deprecation notices?

3. **Board-specific pin presets** — should the library include predefined
   pin configurations for popular PoE boards (Olimex, wESP32, Aptinex)?
   Or leave that to examples?

4. **Frontend/web UI** — should the status page adapt dynamically based
   on active interface (hide SSID/RSSI when on Ethernet, show link
   speed instead)? Or is a static layout acceptable for now?

5. **Source identification fields** — the delta `source` object is
   defined loosely in the Signal K spec. What fields should SensESP
   include beyond `label`? Proposed: `type` ("SensESP"), `src`
   (MAC address as stable ID). Should `firmwareVersion` go in
   `source` or only in the access request? This should align with
   whatever @tkurki settles on for the unified source model in
   [signalk-server#2162](https://github.com/SignalK/signalk-server/issues/2162).

---

## Implementation plan

The work can be done incrementally in reviewable PRs. I would suggest
starting with source identification — it is small, immediately useful,
and connects to the server-side work already in progress. The network
abstraction then follows as a series of focused refactoring steps.

**Phase 1 — Source identification**

1. **PR 1: Enrich source identification**
   Add `type` and `src` (MAC) to the delta source object. Add `deviceId`
   and `firmwareVersion` to the access request. ~10 lines of new code.
   Addresses [signalk-server#2162](https://github.com/SignalK/signalk-server/issues/2162).
   Can use `WiFi.macAddress()` initially and switch to
   `NetworkInterface` once available.

**Phase 2 — Network abstraction**

2. **PR 2: Introduce `NetworkState` and `NetworkStateProducer`**
   Rename the enum, create the unified event listener, add backward
   compat aliases. Wire into `SystemStatusController`. No behavior change.

3. **PR 3: Extract `WiFiProvisioner` from `Networking`**
   Move WiFi-specific code (AP, STA, scan, captive portal) into
   `WiFiProvisioner`. `SensESPApp` uses it identically. Existing user
   code unchanged.

4. **PR 4: Fix the `SKWSClient` WiFi gate**
   Replace `WiFi.isConnected()` with `Network.isOnline()`. One line.

5. **PR 5: Add `EthernetProvisioner` and builder support**
   New class, new builder method, example sketch for a PoE board.

6. **PR 6: Generalize system info and status page**
   Use `NetworkInterface*` for IP/MAC/status reporting. Update PR 1's
   source identification to use the transport-agnostic interface.

Each PR is independently reviewable and testable. PRs 2–4 are pure
refactoring with no behavior change, reducing risk.

---

## Willingness to contribute

I am the author of [signalk-server PR #2379](https://github.com/SignalK/signalk-server/pull/2379)
(source priority UX rework). While working on that, it became clear that
ESP device identification is an open topic on the server side — @tkurki
mentioned it explicitly. The source identification improvement (PR 1)
would be a natural starting point if you are interested, since it is
small and self-contained.

With source identification landed, the network abstraction (PRs 2–6)
follows naturally. I have an [Aptinex IsolPoE ESP32](https://aptinex.com/product/aptinex-isolpoe-esp32-prodev-kit-with-isolated-poe/)
board (ESP32 + LAN8720A + isolated PoE) on hand for testing Ethernet
support on real hardware.

I would appreciate early feedback on the approach — particularly around
the open questions above — before writing code, so that the
implementation aligns with the project's direction.

This builds on the excellent groundwork already in place: the pioarduino
migration, the `SensESPMinimalApp` decoupling, the `WiFiStateProducer`
design that was explicitly made to be replaceable, and the reactive
producer/consumer architecture that makes swapping components clean.

---

## References

- [Arduino ESP32 Network API documentation](https://docs.espressif.com/projects/arduino-esp32/en/latest/api/network.html)
- [Arduino ESP32 Ethernet API documentation](https://docs.espressif.com/projects/arduino-esp32/en/latest/api/ethernet.html)
- [Arduino ESP32 `NetworkInterface` header](https://github.com/espressif/arduino-esp32/blob/master/libraries/Network/src/NetworkInterface.h)
- [ESP32 RF coexistence guide](https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-guides/coexist.html) (WiFi/BLE shared RF)
- [SensESP issue #334 — Modularize protocols and communications layer](https://github.com/SignalK/SensESP/issues/334)
- [SensESP issue #738 — Migrate to ESP-IDF 5.x and Arduino 3.x](https://github.com/SignalK/SensESP/issues/738)
- [SensESP discussion #680 — Connect by bluetooth or USB serial](https://github.com/SignalK/SensESP/discussions/680)
- [SensESP issue #813 — Deep sleep support](https://github.com/SignalK/SensESP/issues/813)
- [SensESP PR #801 — Pioarduino support](https://github.com/SignalK/SensESP/pull/801)
- [signalk-server issue #2162 — Source priority and identification](https://github.com/SignalK/signalk-server/issues/2162)
- [signalk-server PR #2379 — Source priority UX rework](https://github.com/SignalK/signalk-server/pull/2379) (lists SensESP identification as follow-up)

dirkwa:
Implements the transport-agnostic networking layer proposed in #849 and ships the first concrete non-WiFi provisioner (ESP32-P4 Ethernet) on top of it. No behavior change for existing WiFi users — every public class name, enum name, and builder method that existed before still works via compatibility typedefs, so existing sketches compile and run unchanged.

The PR is split into **two commits** for reviewability:

1. **`refactor(net): transport-agnostic networking layer`** — pure refactor, no new features visible to users. Introduces the abstraction, renames `Networking`→`WiFiProvisioner`, unifies the event producer, wires the status page and sensors through the new interface. Drive-by fixes for a long-standing `WiFi.config()` argument-order bug and BSSID hex formatting.
2. **`feat(net): ESP32-P4 Ethernet provisioner + p4_ethernet example + docs`** — concrete ethernet implementation, `set_ethernet()` builder method, downstream `examples/p4_ethernet/` project, and documentation updates (features page, migration guide, example README).

Bundling these into one PR because they tell one story (the refactor exists *because of* the ethernet use case) and the ethernet commit is what makes the refactor demonstrable and reviewable on real hardware.

### New abstraction

- **`NetworkState`** (enum, `network_state.h`) — transport-agnostic link state. Implemented as a strong alias of `WiFiState` so existing `kWifi*` member names keep working. New neutral aliases (`kNetwork*`) are provided for new code.
- **`NetworkStateProducer`** (`network_state.{h,cpp}`) — a single unified producer owned by `SensESPApp` that subscribes to the Arduino-ESP32 `Network.onEvent()` bus and emits state for WiFi STA, WiFi AP, **and** Ethernet events (`ARDUINO_EVENT_WIFI_STA_GOT_IP`, `WIFI_AP_START`, `ETH_GOT_IP`, `ETH_LOST_IP`, `ETH_DISCONNECTED`, …). Replaces the legacy `WiFiStateProducer` that only listened to WiFi events.
- **`NetworkProvisioner`** (`network_provisioner.h`) — abstract base that exposes the runtime transport-agnostic API (`local_ip()`, `gateway_ip()`, `mac_address()`, `is_connected()`). Concrete provisioners inherit from this. No RTTI / no capability-cast methods — `SensESPApp` holds both a base `NetworkProvisioner` pointer and an optional concrete `WiFiProvisioner` pointer so code that needs WiFi-only features (scanning, soft-AP IP) has a direct handle without `dynamic_cast`.

### Concrete WiFi implementation

- **`WiFiProvisioner`** (`wifi_provisioner.{h,cpp}`) — the renamed former `Networking` class, now inheriting from `NetworkProvisioner`. Adds WiFi-specific public methods (`start_wifi_scan`, `get_wifi_scan_results`, `ssid`, `rssi`, `soft_ap_ip`, `is_captive_portal_enabled`) directly on the concrete type. State-producer ownership is removed — the unified `NetworkStateProducer` in `SensESPApp` handles it.
- `networking.h` becomes a compatibility shim that defines `using Networking = WiFiProvisioner`, so every existing user sketch that includes `sensesp/net/networking.h` and references `Networking` or `WiFiState` continues to compile unchanged.

### Wiring changes

- `SensESPApp` now holds three shared pointers: `network_state_producer_` (unified, always non-null), `network_provisioner_` (always non-null — WiFi by default), `wifi_provisioner_` (non-null only when WiFi is active). A new public `set_network_provisioner_factory()` hook lets external code install a custom provisioner.
- `SensESPAppBuilder` gains `disable_wifi()` which clears the default WiFi STA/AP settings. `set_ethernet()` (see commit 2) calls it implicitly.
- `signalk_ws_client.cpp` no longer tests `WiFi.isConnected() || WiFi.getMode() != WIFI_MODE_AP` — it queries `provisioner->is_connected()` instead. Behaviour identical for WiFi users.
- `system_info.cpp` `IPAddrDev::update()` reads from the provisioner (`local_ip()`) instead of `WiFi.localIP()` so the sensor works automatically on any future transport.
- `base_command_handler.cpp` cross-origin check reads from the provisioner instead of `WiFi.localIP()`.
- `http_server.{h,cpp}` no longer pulls in `WiFi.h`. The captive-portal soft-AP IP is now passed explicitly via a second argument to `set_captive_portal()`, so the HTTP server doesn't depend on any specific transport.
- `app_command_handler.{h,cpp}` `add_app_http_command_handlers()` now takes a `shared_ptr<WiFiProvisioner>` (by value) and is only registered when the active provisioner is WiFi. The transport-agnostic subset (`add_tofu_reset_handler`) is exposed as a separate function so non-WiFi apps can still register it.
- `sensesp_app.h` status-page block populates WiFi-only items (`wifi_ssid_ui_output_`, `wifi_rssi_ui_output_`) from `wifi_provisioner_` if present, and leaves them empty otherwise. `mac_address_ui_output_` reads from the base provisioner.

### Drive-by bug fixes

Two latent bugs were uncovered by CodeRabbit while reviewing the refactor. Both were present in `main` and are orthogonal to the refactor, but are fixed here because the refactor is already touching the affected code heavily:

1. **`WiFi.config()` argument order** in `wifi_provisioner.cpp::start_client_autoconnect()`. Arduino-ESP32's signature is `(local_ip, gateway, subnet, dns1, dns2)`, but the pre-refactor code passed `(ip, dns, gateway, netmask)` — silently setting `gateway=dns`, `subnet=gateway`, `dns1=netmask`. **Every static-IP WiFi STA config saved in SensESP was broken as a result.** Documented in the migration guide.
2. **BSSID hex formatting** in `WiFiNetworkInfo::as_json()`. The code built MAC-address strings with `String(byte, HEX)` which produces `"A"` for `0x0A` instead of `"0A"`. Scan results with low hex bytes printed malformed MACs. Replaced with a `snprintf("%02X:…")` buffer.

Happy to split either into its own PR if reviewers prefer.

### `EthernetProvisioner` (ESP32-P4 only)

`src/sensesp/net/ethernet_provisioner.{h,cpp}` — concrete `NetworkProvisioner` subclass that calls Arduino-ESP32's `ETH.begin()` with an external RMII PHY, runs DHCP, and exposes the resulting IPv4 / MAC / link state through the transport-agnostic base class. It does NOT own a state producer — the unified `NetworkStateProducer` in `SensESPApp` picks up `ARDUINO_EVENT_ETH_*` events from the same `Network.onEvent` bus it already listens to for WiFi.

The entire class is guarded behind `#ifdef CONFIG_IDF_TARGET_ESP32P4`. On any other target the header expands to nothing, so a user sketch that tries to instantiate `EthernetConfig` on a classic ESP32 gets a clean compile error at the use site (`'EthernetConfig' was not declared in this scope`) rather than a silent runtime failure. The guard can be widened when other ESP32 variants grow a usable Arduino-ESP32 `ETH.h` implementation.

Ships a single hardware factory `EthernetConfig::waveshare_esp32p4_poe()` for the Waveshare [ESP32-P4-WIFI6-POE-ETH](https://www.waveshare.com/wiki/ESP32-P4-WIFI6-POE-ETH) dev board (IP101GRI PHY at addr 1, MDC=31, MDIO=52, reset=51, external 25 MHz crystal frequency-doubled into GPIO50).

### `SensESPAppBuilder::set_ethernet()`

Templated convenience wrapper that installs the ethernet factory into `SensESPApp` via the existing `set_network_provisioner_factory()` hook and implicitly calls `disable_wifi()`. The template is parameterised over the config struct so `sensesp_app_builder.h` does not need to include any ethernet headers — users who actually call `set_ethernet()` include `ethernet_provisioner.h` in their own translation unit. This keeps the builder header target-neutral.

### `examples/p4_ethernet/`

Downstream PlatformIO project that builds a full SensESP app on ethernet. `platformio.ini` pins pioarduino release `55.03.37` (Arduino-ESP32 3.3.7, ESP-IDF v5.5.2) and targets the `esp32-p4` board variant with a 16 MB flash override for the Waveshare board's larger SPI flash. Consumes SensESP from the parent directory via `symlink://../..`. Ships a `README.md` documenting the hardware, build steps, expected serial output, and what the example proves about the refactor.

### Documentation

- **`docs/pages/features/index.md`** — new "Network transports" section explaining the transport-agnostic model, showing WiFi and Ethernet examples side by side, and describing how to write a custom `NetworkProvisioner` subclass.
- **`docs/pages/migration/index.md`** — new "Transport-agnostic networking" section with the compatibility table (`Networking`→`WiFiProvisioner`, `WiFiState`→`NetworkState`, etc.), a "switch to the new names" snippet for new code, and a prominent warning about the static-IP bug fix so existing users know their saved configs may start working differently after the upgrade.

## CodeRabbit pass

`cr review --plain --type uncommitted` was run iteratively on both commits. **Accepted** findings:

- null-check on `SensESPApp::get()` in `IPAddrDev::update()` (can be called before `setup()` has finished wiring the provisioner on a very fresh boot)
- null-check on `bssid` pointer in `WiFiNetworkInfo` constructor
- BSSID hex zero-padding fix (drive-by, above)
- `WiFi.config()` argument order fix (drive-by, above)
- null-check on factory return value in `SensESPApp::setup()` (falls back to a default `WiFiProvisioner` if a user-supplied factory returns nullptr)
- removed unused `bool ap_mode;` declaration inherited from `networking.cpp`
- clarified CH340 device path in `examples/p4_ethernet/README.md` (mentions both `/dev/ttyACM0` and `/dev/ttyUSB0`)

**Declined** (out of scope — pre-existing issues that should get their own PRs if they prove to be real problems in practice):

- a recommendation to `delay(300)` after `WiFi.disconnect()` in `start_wifi_scan()`
- a note about static locals in the `reconnect_cb` lambda in `start_client_autoconnect()`

Second pass on the final diff: zero findings.

## Tested

- **`pio run -e pioarduino_esp32`** on the library top level — compiles cleanly; only the expected `setup()` / `loop()` linker undefs remain (library-only build)
- **`pio run -e pioarduino_esp32c3`** — compiles cleanly (same status)
- **`examples/p4_ethernet`** builds cleanly with pioarduino 55.03.37 and produces a 1.58 MB firmware image (80 % of the `min_spiffs.csv` 1.9 MB app slot)
- Flashed to a **Waveshare ESP32-P4-WIFI6-POE-ETH** board (MAC `80:F1:B2:D2:CC:6E`, chip rev v1.3, 16 MB flash, 32 MB PSRAM):
  - Obtained DHCP lease `192.168.99.13` from a Linux host running `nmcli connection add type ethernet ... ipv4.method shared ipv4.addresses 192.168.99.1/24`
  - SensESP HTTP UI reachable at `http://192.168.99.13/`, `/api/info` returns the full status JSON
  - Status page shows empty `SSID` / `WiFi signal strength=0` (correct — WiFi inactive) and `MAC Address` populated from `ETH.macAddress()` (correct)
  - Configured Signal K server via REST (`PUT /api/config/System/Signal%20K%20Settings` with address `192.168.99.1` port `3000`), restarted, verified `SK connection status = Connected` and `SK Delta TX count` ticking up
  - signalk-server 2.24 on the Linux host receives the `uptime` / `freeMemory` / `ipAddress` deltas under `sensorDevice.sensesp-p4-eth` — `ipAddress` correctly reports `192.168.99.13`, proving `IPAddrDev::update()` reads through the refactored `NetworkProvisioner::local_ip()`

Classic WiFi runtime testing on a real ESP32/S3/C3 board was not performed in this cycle because the only hardware on hand is the P4, which has no native WiFi. The refactor preserves the WiFi code path bit-for-bit, so this should be a no-op for existing WiFi users — but a sanity-check from someone with classic hardware would be welcome.

Closes #849

### DIFF
--- a/docs/pages/features/index.md
+++ b/docs/pages/features/index.md
@@ -59,5 +59,74 @@ STEP 2:
 
 ----
 
+## Network transports
+
+As of the transport-agnostic networking refactor, SensESP can reach
+the Signal K server over any interface that Arduino-ESP32 3.x can
+bring up, not only WiFi. A single `NetworkStateProducer` owned by
+`SensESPApp` subscribes to the unified `Network.onEvent()` bus and
+emits a transport-agnostic `NetworkState` whenever any interface's
+IP-layer state changes. The HTTP server, mDNS, Signal K websocket
+client, and system-info sensors all query a `NetworkProvisioner`
+abstract base instead of touching `WiFi` directly, so they work the
+same on WiFi and Ethernet.
+
+### WiFi (default)
+
+Existing user sketches do not need to change. `SensESPAppBuilder`
+defaults to constructing a `WiFiProvisioner` (the former `Networking`
+class — the old name is still available as a deprecated typedef).
+
+```c++
+SensESPAppBuilder builder;
+auto app = builder.set_hostname("my-device")
+                  ->set_wifi_client("SSID", "password")
+                  ->get_app();
+```
+
+### Ethernet (ESP32-P4)
+
+Native RMII Ethernet is supported on ESP32-P4 boards with an external
+PHY. The library currently ships a configuration preset for the
+Waveshare ESP32-P4-WIFI6-POE-ETH dev board (IP101GRI PHY, external
+25 MHz crystal). To use it:
+
+```c++
+#include "sensesp/net/ethernet_provisioner.h"
+#include "sensesp_app_builder.h"
+
+SensESPAppBuilder builder;
+auto app = builder.set_hostname("my-device")
+                  ->set_ethernet(
+                      sensesp::EthernetConfig::waveshare_esp32p4_poe())
+                  ->get_app();
+```
+
+`set_ethernet()` implicitly calls `disable_wifi()` so no WiFi soft-AP
+fall-back or captive portal is registered. The SensESP web UI remains
+reachable at the DHCP-assigned address on the Ethernet side.
+
+`EthernetProvisioner` is only declared on targets where the
+Arduino-ESP32 ETH driver is usable — currently ESP32-P4. Calling
+`set_ethernet()` on an unsupported target produces a compile error at
+the use site (`EthernetConfig was not declared in this scope`), not a
+silent runtime failure. See `examples/p4_ethernet/` for a full
+buildable example.
+
+### Writing your own provisioner
+
+Subclass `sensesp::NetworkProvisioner` (see
+`sensesp/net/network_provisioner.h`) and bring your interface up
+yourself — e.g. PPP over UART for a cellular modem. The base class
+only requires `local_ip()`, `gateway_ip()`, `mac_address()`, and
+`is_connected()`. Install your factory by calling
+`app->set_network_provisioner_factory(...)` before
+`SensESPApp::setup()` runs (typically from a builder extension). The
+unified `NetworkStateProducer` will pick up state changes automatically
+as long as your interface emits the standard `ARDUINO_EVENT_*`
+events through `Network.onEvent`.
+
+----
+
 ## WiFi manager
 ## System info sensors
--- a/docs/pages/migration/index.md
+++ b/docs/pages/migration/index.md
@@ -345,3 +345,71 @@ sensesp_app = builder.enable_free_mem_sensor()
                      ->enable_ota("my_password")
                      ->get_app();
 ```
+
+## Transport-agnostic networking (v3 minor update)
+
+SensESP used to assume WiFi was the only way to reach the Signal K
+server. The networking layer has been refactored to be transport-
+agnostic: WiFi is still the default and existing sketches compile
+and behave identically, but the library can now also run over
+native Ethernet (on ESP32-P4) and is ready for future transports.
+The refactor RFC is [#849](https://github.com/SignalK/SensESP/issues/849).
+
+### No action required for WiFi users
+
+Every public class name, enum name, and builder method that existed
+before still works via compatibility typedefs:
+
+| Old name                          | New name                               |
+|-----------------------------------|----------------------------------------|
+| `Networking`                      | `WiFiProvisioner`                      |
+| `WiFiState` (enum)                | `NetworkState`                         |
+| `WiFiStateProducer`               | `NetworkStateProducer`                 |
+| `SensESPApp::get_networking()`    | `SensESPApp::get_wifi_provisioner()`   |
+
+The kept aliases mean a user sketch like
+
+```c++
+auto state = sensesp_app->get_networking()->get_wifi_state_producer();
+```
+
+compiles unchanged against the new release.
+
+### Optional: switch to the new names
+
+For new code or if you want to use the new transport-agnostic API,
+prefer:
+
+```c++
+// Include the concrete header directly (not "networking.h")
+#include "sensesp/net/wifi_provisioner.h"
+#include "sensesp/net/network_state.h"
+
+// Access the provisioner through its new accessor
+auto wifi = sensesp_app->get_wifi_provisioner();
+
+// Or, for code that should work on any transport, use the
+// NetworkProvisioner base and the unified state producer:
+auto provisioner = sensesp_app->get_network_provisioner();
+auto ip  = provisioner->local_ip();
+auto mac = provisioner->mac_address();
+
+auto state_producer = sensesp_app->get_network_state_producer();
+state_producer->connect_to(&my_consumer);
+```
+
+### Static-IP WiFi configurations
+
+If you use a saved static-IP configuration on a WiFi client, the
+release in which this refactor lands also fixes a long-standing bug
+in `WiFiProvisioner::start_client_autoconnect()` where the arguments
+to `WiFi.config()` were passed in the wrong order — silently setting
+`gateway = DNS server`, `subnet = gateway`, and `DNS1 = netmask`. If
+your saved WiFi client config has `useDHCP: false`, verify your
+gateway / netmask / DNS values look correct after upgrading; existing
+saved configs should actually start working for the first time.
+
+### Adding Ethernet
+
+See `docs/pages/features/index.md` (section "Network transports")
+for a short usage example.
--- a/examples/p4_ethernet/README.md
+++ b/examples/p4_ethernet/README.md
@@ -1,0 +1,88 @@
+# p4_ethernet
+
+Minimal SensESP application that runs over native RMII Ethernet on a
+Waveshare [ESP32-P4-WIFI6-POE-ETH](https://www.waveshare.com/wiki/ESP32-P4-WIFI6-POE-ETH)
+dev board. WiFi is disabled entirely — the device brings up the
+onboard IP101GRI PHY, runs DHCP, and reaches the Signal K server
+over the wired link (optionally powered via PoE through the RJ45
+magjack).
+
+This example exists primarily as an end-to-end smoke test for the
+transport-agnostic networking refactor
+([RFC #849](https://github.com/SignalK/SensESP/issues/849)): it
+exercises the new `NetworkProvisioner` / `NetworkStateProducer` path
+and the `EthernetProvisioner` concrete implementation from a real
+`SensESPAppBuilder` invocation.
+
+## Hardware
+
+- Board: Waveshare ESP32-P4-WIFI6-POE-ETH (ESP32-P4 ECO2 / rev v1.3
+  silicon, 16 MB flash, 32 MB PSRAM). Other ESP32-P4 boards with an
+  IP101GRI / TLK110-class PHY on the same RMII pinout should also
+  work, possibly by extending `EthernetConfig` with a new factory.
+- USB-C: the board exposes a single USB-C wired through a CH340
+  USB-UART bridge (QinHeng 1a86:55d3) to the P4's UART0 (GPIO 37
+  TX / GPIO 38 RX). The P4's native USB-OTG is not exposed, so
+  `ARDUINO_USB_CDC_ON_BOOT` is deliberately left undefined —
+  Arduino `Serial` maps to UART0 and reaches the host. On recent
+  Linux kernels (with `cdc-acm` handling the newer CH340 variant)
+  the device enumerates as `/dev/ttyACM0`; older systems loading
+  the legacy vendor-specific driver may see `/dev/ttyUSB0` instead.
+  On macOS it is usually `/dev/cu.usbmodem*`; on Windows a COM port.
+- Ethernet: standard RJ45 patch cable to any switch or router that
+  serves DHCP. For isolated bench testing, a Linux host running
+  `nmcli connection add type ethernet ifname ethX con-name test
+  ipv4.method shared ipv4.addresses 192.168.99.1/24` will spin up a
+  NetworkManager-managed `dnsmasq` on that interface.
+
+## Build and flash
+
+This is a standalone PlatformIO project that consumes SensESP from
+the parent directory via a `symlink://../..` library dep.
+
+```sh
+cd examples/p4_ethernet
+pio run -t upload   # PlatformIO auto-detects the serial port
+pio device monitor  # view boot log at 115200 bps
+```
+
+The `[env:waveshare_esp32p4_eth]` environment in `platformio.ini`
+pins pioarduino release 55.03.37 (Arduino-ESP32 3.3.7, ESP-IDF
+v5.5.2) and targets the `esp32-p4` board variant with 16 MB flash
+override for the Waveshare board's larger SPI flash.
+
+## Expected behaviour
+
+On a successful boot the serial monitor should show:
+
+```text
+[p4_eth] boot
+...
+I (XXX) sensesp/net/ethernet_provisioner.cpp: Bringing up Ethernet ...
+I (XXX) sensesp/net/ethernet_provisioner.cpp: Ethernet driver started; ...
+...
+I (XXX) sensesp/net/network_state.cpp: net_state: Ethernet got IP
+[p4_eth] SensESPApp constructed
+```
+
+Once DHCP completes, the SensESP web UI should be reachable at
+`http://<leased-ip>/`. The status page will show an empty SSID and
+RSSI (both WiFi-specific), but `MAC Address` will be populated from
+the Ethernet MAC, and the Signal K settings page can be used to
+point the device at a running `signalk-server`.
+
+## What this example proves
+
+- `NetworkStateProducer` correctly receives `ARDUINO_EVENT_ETH_GOT_IP`
+  through the unified `Network.onEvent` bus and emits
+  `kNetworkConnected`.
+- `SensESPApp::setup()` dispatches to the builder-supplied factory,
+  leaving `wifi_provisioner_` null, and the status page suppresses
+  its WiFi-only items.
+- `IPAddrDev` (the built-in sensor) reads its value through
+  `NetworkProvisioner::local_ip()`, so `sensorDevice.<hostname>.ipAddress`
+  on the Signal K server reflects the Ethernet IP without any
+  Ethernet-specific code in the sensor.
+- `SKWSClient::connect()` gates on `provisioner->is_connected()`
+  instead of `WiFi.isConnected()` and successfully dials the Signal
+  K server over the wired link.
--- a/examples/p4_ethernet/platformio.ini
+++ b/examples/p4_ethernet/platformio.ini
@@ -1,0 +1,59 @@
+; PlatformIO project for the M2 Ethernet smoke test on Waveshare
+; ESP32-P4-WIFI6-POE-ETH. Builds a real SensESP app using the new
+; transport-agnostic NetworkProvisioner / EthernetNetworking path with
+; WiFi disabled.
+;
+; Usage:
+;   cd examples/p4_ethernet
+;   pio run -t upload && pio device monitor
+;
+; Expected runtime behaviour:
+;   - PHY link up message
+;   - DHCP lease obtained, IP printed
+;   - SensESP web UI reachable at http://<dhcp-ip>/
+;   - mDNS service announced
+;   - SK WS client attempts connection (configure SK server in web UI)
+
+[platformio]
+default_envs = waveshare_esp32p4_eth
+
+[env]
+upload_speed = 2000000
+monitor_speed = 115200
+monitor_filters = esp32_exception_decoder
+build_unflags = -Werror=reorder
+board_build.partitions = min_spiffs.csv
+
+build_flags =
+    -D CORE_DEBUG_LEVEL=ARDUHAL_LOG_LEVEL_VERBOSE
+    -D USE_ESP_IDF_LOG
+
+; deep LDF: scan SensESP headers for transitive deps (esp_websocket_client)
+; without descending into framework libs (which broke chain+ earlier).
+lib_ldf_mode = deep
+
+lib_deps =
+    ; Use the local SensESP working tree.
+    symlink://../..
+    ; SensESP's WS client transitively needs this from the IDF component
+    ; registry. Listed here so PIO actually installs it for the symlink
+    ; build (the SensESP library.json doesn't declare it as a hard dep).
+    esp_websocket_client=https://components.espressif.com/api/downloads/?object_type=component&object_id=dbc87006-9a4b-45e6-a6ab-b286174cb413
+
+[env:waveshare_esp32p4_eth]
+; Pinned pioarduino 55.03.37 — same as the p4_blink example.
+platform = https://github.com/pioarduino/platform-espressif32/releases/download/55.03.37/platform-espressif32.zip
+framework = arduino
+
+; Waveshare board = ESP32-P4 ECO2 silicon (rev v1.3) with 16 MB flash + 32 MB PSRAM.
+board = esp32-p4
+board_upload.flash_size = 16MB
+board_upload.maximum_size = 16777216
+
+build_flags =
+    ${env.build_flags}
+    -D SENSESP_SSL_SUPPORT=1
+    -DBOARD_HAS_PSRAM
+    ; Serial maps to UART0 → onboard CH340 → /dev/ttyACM0.
+    ; Native USB-OTG is not exposed on this board, so do NOT enable
+    ; ARDUINO_USB_CDC_ON_BOOT.
--- a/examples/p4_ethernet/src/main.cpp
+++ b/examples/p4_ethernet/src/main.cpp
@@ -1,0 +1,43 @@
+/**
+ * @file main.cpp
+ * @brief Minimal SensESP app running on native Ethernet on the
+ *        Waveshare ESP32-P4-WIFI6-POE-ETH board.
+ *
+ * Uses the transport-agnostic NetworkProvisioner abstraction and the
+ * Waveshare-tuned EthernetProvisioner config. WiFi is disabled
+ * entirely — the device runs over native ESP32-P4 RMII Ethernet to
+ * the onboard IP101GRI PHY and can be PoE-powered via the RJ45
+ * magjack.
+ *
+ * Verification checklist on the host (with a DHCP server on the
+ * same LAN):
+ *   1. Device boots, serial prints "Bringing up Ethernet"
+ *   2. PHY link comes up when a cable is plugged in
+ *   3. DHCP lease obtained; check the ARP table on the host
+ *   4. http://<leased-ip>/ serves the SensESP web UI
+ *   5. Configure the Signal K server address in the web UI
+ *   6. SK connection status on the status page flips to "Connected"
+ */
+
+#include "sensesp/net/ethernet_provisioner.h"
+#include "sensesp_app_builder.h"
+
+using namespace sensesp;
+
+void setup() {
+  Serial.begin(115200);
+  delay(500);
+  Serial.println("[p4_eth] boot");
+
+  SensESPAppBuilder builder;
+  auto sensesp_app = builder.set_hostname("sensesp-p4-eth")
+                         ->set_ethernet(EthernetConfig::waveshare_esp32p4_poe())
+                         ->enable_uptime_sensor()
+                         ->enable_free_mem_sensor()
+                         ->enable_ip_address_sensor()
+                         ->get_app();
+
+  Serial.println("[p4_eth] SensESPApp constructed");
+}
+
+void loop() { event_loop()->tick(); }
--- a/library.json
+++ b/library.json
@@ -44,6 +44,10 @@
       "owner": "bxparks",
       "name": "AceButton",
       "version": "^1.10.1"
+    },
+    {
+      "name": "esp_websocket_client",
+      "version": "https://components.espressif.com/api/downloads/?object_type=component&object_id=dbc87006-9a4b-45e6-a6ab-b286174cb413"
     }
   ],
   "version": "3.2.3-alpha",
--- a/src/sensesp/net/ethernet_provisioner.cpp
+++ b/src/sensesp/net/ethernet_provisioner.cpp
@@ -1,0 +1,67 @@
+#include "sensesp.h"
+
+#include "ethernet_provisioner.h"
+
+#if defined(CONFIG_IDF_TARGET_ESP32P4)
+
+namespace sensesp {
+
+EthernetProvisioner::EthernetProvisioner(const EthernetConfig& config)
+    : config_(config) {
+  ESP_LOGI(__FILENAME__,
+           "Bringing up Ethernet (PHY type=%d addr=%d MDC=%d MDIO=%d "
+           "PWR=%d clk_mode=%d)",
+           (int)config_.phy_type, (int)config_.phy_addr, config_.mdc_pin,
+           config_.mdio_pin, config_.power_pin, (int)config_.clock_mode);
+
+  // Set the hostname before bringing up the link so DHCP REQUESTs carry
+  // it. SensESP stores the hostname on the SensESPBaseApp singleton.
+  String hostname = SensESPBaseApp::get_hostname();
+  ETH.setHostname(hostname.c_str());
+
+  // Use the explicit RMII begin() so SensESP doesn't depend on the
+  // variant supplying compile-time ETH_PHY_* macros — although on the
+  // current Waveshare ESP32-P4 board the variant happens to match.
+  bool started = ETH.begin(config_.phy_type, config_.phy_addr, config_.mdc_pin,
+                           config_.mdio_pin, config_.power_pin,
+                           config_.clock_mode);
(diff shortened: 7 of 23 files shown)


mairas:
While I totally appreciate the renaming and refactoring of `Networking`, I was going to whine that it's a change that should only be done in a major version release - until I saw the compatibility shim. I like everything here so far. I don't have a whole lot to say about `EthernetProvisioner` and I don't have the hardware to test it, either.

I'll have a closer look tomorrow (hopefully) and also usher Claude Code at it, but LGTM so far.

mairas:
## Code Review: transport-agnostic networking refactor

Excellent work overall — the architecture is sound, the separation between provisioning and state production is clean, and the drive-by bug fixes (`WiFi.config()` argument order, BSSID hex formatting) are welcome. The ESP32-P4 Ethernet implementation follows Arduino-ESP32 patterns correctly.

However, the backward compatibility claim ("every public class name, enum name, and builder method that existed before still works") is overstated. There are a few real API breaks that should be documented.

### P1 — Backward compatibility gaps

**1. `WiFiProvisioner` no longer inherits `ValueProducer<WiFiState>`**

- Old: `class Networking : public FileSystemSaveable, public Resettable, public ValueProducer<WiFiState>`
- New: `class WiFiProvisioner : public FileSystemSaveable, public Resettable, public NetworkProvisioner`

The typedef `using Networking = WiFiProvisioner` preserves the *name* but not the *interface*. Code that called `emit()`, `connect_to()`, or `get()` on a `Networking` object (treating it as a `ValueProducer`) will fail to compile. The migration path is `app->get_network_state_producer()`, but this isn't mentioned in the migration table.

**2. `WiFiStateProducer` class removed from public API**

Any code that instantiated `WiFiStateProducer` directly will break. Replaced by `NetworkStateProducer`, owned by `SensESPApp`.

**3. `get_wifi_state_producer()` method removed**

Code calling `app->get_networking()->get_wifi_state_producer()` won't compile. The migration guide's compatibility table lists `WiFiStateProducer → NetworkStateProducer` but doesn't mention that the accessor method moved from the provisioner to the app.

**4. `get_networking()` return type changed from reference to value**

- Old: `std::shared_ptr<Networking>& get_networking()` (reference to member)
- New: `std::shared_ptr<WiFiProvisioner> get_networking()` (temporary)

Code like `auto& net = app->get_networking()` compiles but creates a dangling reference.

**Impact:** Basic sketches that just call `set_wifi_client()` / `get_app()` are unaffected. Advanced users who subscribe to WiFi state events via the provisioner will hit compile errors. I'd recommend either documenting these as intentional breaking changes in the migration guide, or adding a compat shim (e.g., a forwarding `get_wifi_state_producer()` on `WiFiProvisioner`).

### P2 — Moderate issues

**5. Captive portal soft-AP IP timing change**

`setup()` now reads `wifi_provisioner_->soft_ap_ip()` once during construction and caches it in `captive_portal_ap_ip_`. The old code read `WiFi.softAPIP()` at HTTP request dispatch time. If the soft-AP hasn't finished initializing when `setup()` reads the IP, it gets `0.0.0.0` and captive portal breaks. Likely fine in practice (the WiFiProvisioner constructor calls `WiFi.softAP()` synchronously), but worth verifying.

**6. `is_connected()` semantic inconsistency**

`WiFiProvisioner::is_connected()` returns `true` in AP/APSTA mode (device is a hotspot, no upstream). `EthernetProvisioner::is_connected()` only returns `true` with link+IP. The base class doc says "True iff the link is currently up and we have an IP address" — WiFi's AP-mode behavior doesn't match this contract. Pre-existing behavior, but now codified in the abstract interface.

**7. New transitive dependency**

`esp_websocket_client` added to `library.json`. Was already an implicit dependency, but making it explicit changes the build graph for all consumers. Worth noting in release notes.

### P3 — Minor

**8. `SensESPApp::get()` not null-checked** in `signalk_ws_client.cpp:595` and `base_command_handler.cpp:19` before calling `->get_network_provisioner()`. The PR already added this check in `IPAddrDev::update()` — same pattern should be applied for consistency. Theoretical risk only (ESP32 never calls `destroy()`).

### What looks good

- `NetworkProvisioner` is a minimal, well-defined interface (4 pure virtual methods)
- Template-based `set_ethernet()` avoids pulling ETH headers into the builder
- `#ifdef CONFIG_IDF_TARGET_ESP32P4` guard gives compile-time errors on unsupported targets
- `defer_emit` thread-safety pattern is consistent with the existing (proven) `WiFiStateProducer` approach
- Hostname timing in `EthernetProvisioner` (register `ETH_START` handler before `ETH.begin()`) follows the Arduino-ESP32 reference pattern
- `WiFi.config()` fix is correct — Arduino-ESP32 signature is `(local_ip, gateway, subnet, dns1, dns2)`
- Documentation (features page, migration guide, example README) is thorough

mairas:
I think we should maintain backward compatibility for now. Making `WiFiProvisioner` a `Provider<WiFiState>` may not be the cleanest design but it is used in existing code, so I wouldn't want to break it.

Also, moderate issues 5 and 6 sound worth fixing. Numbers 7 and 8 can be ignored.

dirkwa:
Thanks for the thorough review! Pushed `dcab14d88` addressing the feedback:

### P1 — Backward compatibility (all fixed)

**1. `WiFiProvisioner` now inherits `ValueProducer<WiFiState>`** — code that did `networking->connect_to(consumer)` compiles and works again. `SensESPApp::setup()` wires a forwarding lambda from `NetworkStateProducer` so the WiFiProvisioner emits the same state transitions the old `Networking` did.

**2+3. `WiFiStateProducer` / `get_wifi_state_producer()`** — added a deprecated `get_wifi_state_producer()` method on `WiFiProvisioner` that returns `this` (since WiFiProvisioner is now itself a `ValueProducer<WiFiState>`). Code that called `app->get_networking()->get_wifi_state_producer()->connect_to(...)` compiles unchanged.

**4. `get_networking()` return type** — changed from `shared_ptr<WiFiProvisioner>` (value) to `shared_ptr<WiFiProvisioner>&` (reference), matching the old signature and avoiding dangling references.

### P2 — Moderate issues

**5. Captive portal soft-AP IP timing** — verified safe. `WiFi.softAP()` is called synchronously in the `WiFiProvisioner` constructor (line 76 of `wifi_provisioner.cpp`), so `soft_ap_ip()` returns a valid IP by the time `setup()` reads it. No change needed.

**6. `is_connected()` semantic inconsistency** — updated the `NetworkProvisioner::is_connected()` doc to explicitly state that WiFi AP mode counts as connected (device is network-reachable via the soft-AP). This is intentional behavior — the captive portal and HTTP server depend on it. Changing it would break AP mode functionality.

mairas:
The CI checks weren't being run for non-maintainer PRs due to a minor bug, but when run, would fail. I noticed that when trying to add a missing WiFiStateManager compatibility typedef CC found. I enabled the checks for everyone in PR #943, and now the test build is failing for Arduino 2.0 targets. Would you mind having a look?

mairas:
I also pushed the WiFiStateProvider typedef since it's not related to the CI job failure.

dirkwa:
All green. 

The Node.js 20 deprecation warnings on the GitHub Actions runners have to be fixed in the CI scripts.

mairas:
Thanks!

PR #942, opened 2026-04-09
feat: add ESP32-P4 support to the [espidf] build mode

dirkwa:
## Summary

Extends SensESP's existing `framework = espidf, arduino` build mode
(the `[espidf]` template at `platformio.ini:104`) to **ESP32-P4**,
alongside the existing `espidf_esp32` and `espidf_esp32c3` envs.
This is the IDF-component build mode — Arduino-ESP32 is compiled as
a component inside an IDF project, and the project owns its own
`sdkconfig.defaults`.

Depends on #940 (transport-agnostic networking + Ethernet
provisioner) being merged first — this PR uses the
`EthernetProvisioner` from that one in its example.

## What this unlocks

Concretely: you can now build SensESP for ESP32-P4 via `pio run -e
espidf_esp32p4` and get a clean library build to the same status as
`espidf_esp32` / `espidf_esp32c3`. This is the foundation for any
advanced P4 work that needs full IDF sdkconfig control — custom BT
host stack selection, esp_hosted coprocessor wiring, memory pool
tuning, anything that the Arduino-ESP32 prebuilt libs lock down.

The included example (`examples/p4_bluedroid_poc/`) is a
hardware-verified stage-1 demonstration: SensESP + Ethernet on a
Waveshare ESP32-P4-WIFI6-POE-ETH board running a firmware with
**Bluedroid** selected as the BT host stack (instead of the NimBLE
default) + esp_hosted's Bluedroid VHCI driver enabled. Stage-1
doesn't yet call into BLE from the app code — that's left for a
follow-up because of a separate C6 slave firmware issue (see Known
issue below). But the build and boot path works end-to-end.

### Commit 1 — `feat(build): add [env:espidf_esp32p4] for the espidf build mode`

Root `platformio.ini` additions:

- **New `[espidf_cloud_certs]` intermediate section.** Moves the
  `board_build.embed_txtfiles` list (esp_insights +
  esp_rainmaker cert files) out of the top-level `[espidf]` section
  into its own optional mixin. The existing `espidf_esp32` and
  `espidf_esp32c3` envs now extend `(espidf, espidf_cloud_certs,
  <board>)` instead of `(espidf, <board>)`, preserving their
  behaviour bit-for-bit.
- **Why the split was needed.** Arduino-ESP32's own
  `idf_component.yml` explicitly excludes `esp_insights` and
  `esp_rainmaker` on `esp32c2` and `esp32p4`:
  ```yaml
  espressif/esp_insights:
    rules:
    - if: target not in [esp32c2, esp32p4]
  ```
  On those targets the embed paths fail during configure with
  `Source 'managed_components/espressif__esp_insights/server_certs/
  https_server.crt' not found`. Moving the embed list out lets
  esp32p4 (and the currently-disabled esp32c2) run the
  `[espidf]` template cleanly.
- **New `[esp32p4]` board section + `[env:espidf_esp32p4]` env,**
  mirroring the `[esp32]` / `[esp32c3]` pattern.
- **`.gitignore` additions** for the pioarduino-generated
  artifacts that `framework = espidf, arduino` writes at the repo
  root on first configure (`CMakeLists.txt`, `src/CMakeLists.txt`,
  `dependencies.lock`, `sdkconfig.<env>`), with an explicit
  `!sdkconfig.defaults` unignore so the hand-authored baseline
  stays tracked. Keeps `git status` clean across all espidf envs.

### Commit 2 — `feat(examples): add p4_bluedroid_poc`

Standalone PlatformIO project under `examples/p4_bluedroid_poc/`
that consumes SensESP via `symlink://../..`:

- `platformio.ini` — `framework = espidf, arduino` on pioarduino
  55.03.37
- `sdkconfig.defaults` — selects Bluedroid, enables esp_hosted's
  Bluedroid HCI VHCI driver, and sets
  `CONFIG_ESP32P4_SELECTS_REV_LESS_V3=y` for ES silicon (required
  on current Waveshare boards — without it the firmware crashes
  with an Illegal Instruction panic at the entry point because
  of an incompatible memory-layout / startup code path)
- `idf_component.yml` — pins `espressif/esp_hosted` to 2.12.3
- `p4_16mb.csv` — 16 MB partition table, two 6.5 MB OTA slots
- `src/main.cpp` — minimal `SensESPAppBuilder` invocation
  (`set_hostname`, `set_ethernet(waveshare_esp32p4_poe())`,
  `disable_wifi`, `enable_ota`), plus a 5-second heartbeat log
- `README.md` — hardware setup, build and flash commands, expected
  boot output, what the example proves, and an honest Known
  Issue section about the C6 slave-fw state

## Tested

- `pio run -e espidf_esp32` — compiles cleanly; only the expected
  `setup()` / `loop()` linker undefs remain (library-only build).
  **No regression** from before this PR.
- `pio run -e espidf_esp32c3` — same result, no regression.
- `pio run -e espidf_esp32p4` — same result. This is new — there
  was no P4 env in the `[espidf]` template before.
- **Hardware-verified end-to-end boot** of the
  `examples/p4_bluedroid_poc` firmware on a Waveshare
  ESP32-P4-WIFI6-POE-ETH (rev v1.3 ES silicon, MAC
  `80:f1:b2:d2:ca:6a`) over USB-C. Successful boot sequence:
  - ESP-IDF 5.5.2.260206 init, chip rev v1.3 detected
  - All memory pools (RETENT_RAM, RAM, RTCRAM, TCM, 32 MB PSRAM)
    initialised
  - SensESP filesystem, EthernetProvisioner bringing up the
    IP101GRI PHY, hostname set via the ETH_START handler (from
    the hostname-fix commit in #940), DHCP lease obtained at
    192.168.99.131
  - mDNS responder, HTTP server on :80, OTA server on
    `esp32-<mac>.local:3232`, SK websocket client attempting to
    connect
  - `POC: alive — uptime=…` heartbeat every 5 s, ~455 KB free
    heap
  - Total: ~23.5 % flash / ~10.6 % RAM on the 6.5 MB / 320 KB
    partition

## Known issue — C6 slave firmware BLE adv forwarding

The example's stage-1 `main.cpp` deliberately does not call any
BLE API. During the scaffolding work I confirmed that the
Bluedroid host stack initialises cleanly on this build path
(`esp_bluedroid_init` / `_enable` return OK,
`esp_ble_gap_set_ext_scan_params` and `esp_ble_gap_start_ext_scan`
both succeed), but LE advertising reports never arrive from the
onboard ESP32-C6 companion. The same silence was reproduced
earlier with NimBLE on two separate boards, which strongly suggests
the blocker is in the Arduino-ESP32-distributed esp_hosted C6
slave firmware (`esp32c6-v2.11.6.bin`), not in the P4 host stack
choice.

Since that's a downstream issue best tracked against
[esp-hosted-mcu](https://github.com/espressif/esp-hosted-mcu)
rather than SensESP, stage-2 is left as a follow-up. This PR is
deliberately scoped to **build-mode support for ESP32-P4**, not
working BLE on P4.

## On your CI placement issue

I saw your note that the `[espidf]` mode is currently commented
out in `.github/workflows/ci.yml` because of "a sdkconfig file
placement issue in the CI environment". I'm not 100 % sure the
bug fixed here is the same one — the failure I reproduced was
the `embed_txtfiles` paths failing during configure on esp32p4,
which then cascades into various downstream errors including
sdkconfig / ldgen confusion. If that cascade is what CI was
hitting, this PR should unblock it. If your CI is hitting a
distinct placement issue on top of this, I'd be happy to dig
into it as a follow-up — just let me know what the symptom is.

## Follow-ups (not part of this PR)

1. Stage-2 Bluedroid BLE scanner code, once the C6 slave-fw story
   is resolved upstream.
2. An upstream bug report to pioarduino about
   `chip_variant: esp32p4_es` not propagating into IDF's sdkconfig
   when building with `framework = espidf, arduino` — currently
   users have to discover and set `CONFIG_ESP32P4_SELECTS_REV_LESS_V3=y`
   manually. Not blocking for this PR.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,8 +17,8 @@ managed_components
 # `platformio.ini` (which declares the envs and their custom_sdkconfig)
 # and `sdkconfig.defaults` (the hand-authored baseline). Everything in
 # the list below is machine-generated from those two sources.
-CMakeLists.txt
-src/CMakeLists.txt
-dependencies.lock
+/CMakeLists.txt
+/src/CMakeLists.txt
+/dependencies.lock
 sdkconfig.*
 !sdkconfig.defaults
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,15 @@ docs/Gemfile.lock
 docs/.bundle
 docs/_site
 managed_components
+
+# pioarduino's `framework = espidf, arduino` build mode generates these
+# at the repo root on first configure. They are per-environment build
+# artifacts and should not be committed — the project-tracked inputs are
+# `platformio.ini` (which declares the envs and their custom_sdkconfig)
+# and `sdkconfig.defaults` (the hand-authored baseline). Everything in
+# the list below is machine-generated from those two sources.
+CMakeLists.txt
+src/CMakeLists.txt
+dependencies.lock
+sdkconfig.*
+!sdkconfig.defaults
--- a/examples/p4_bluedroid_poc/README.md
+++ b/examples/p4_bluedroid_poc/README.md
@@ -1,0 +1,127 @@
+# p4_bluedroid_poc
+
+Stage-1 SensESP application that runs over native RMII Ethernet on a
+Waveshare [ESP32-P4-WIFI6-POE-ETH](https://www.waveshare.com/wiki/ESP32-P4-WIFI6-POE-ETH)
+dev board, built via the new `[env:espidf_esp32p4]` env in the root
+`platformio.ini` and the companion `framework = espidf, arduino` build
+mode. WiFi is disabled entirely — the device brings up the onboard
+IP101GRI PHY, runs DHCP, and reaches the Signal K server over the
+wired link.
+
+This example is a standalone PlatformIO project (not an env in the
+root `platformio.ini`) because it needs its own `sdkconfig.defaults`,
+`idf_component.yml`, and partition table. It is the first end-to-end
+demonstration of SensESP running under `framework = espidf, arduino`
+on ESP32-P4 hardware with **Bluedroid** selected as the BT host stack
+instead of the NimBLE-only Arduino-ESP32 prebuilt libs.
+
+## What this example is — and isn't
+
+This is the **stage-1 baseline**: it stands up SensESP + Ethernet on
+a Bluedroid-configured firmware and verifies the whole build and boot
+path works. Bluedroid is compiled into the binary (via
+`sdkconfig.defaults`) but no BLE code is called from `main.cpp` — the
+firmware's only runtime behaviour is the usual SensESP stack plus a
+5-second heartbeat log.
+
+The **stage-2 BLE scanner** — which would add a Bluedroid GAP scan
+loop over the esp_hosted VHCI transport to the onboard ESP32-C6
+companion — is deliberately not included in this commit. During the
+PoC work we confirmed that the Bluedroid stack initialises cleanly on
+P4 via this build path (`esp_bluedroid_init` / `_enable` return OK,
+`esp_ble_gap_set_ext_scan_params` and `esp_ble_gap_start_ext_scan`
+both succeed), but advertising reports never arrive from the C6
+slave. The same silence was reproduced earlier with the NimBLE host
+stack on two separate boards, which strongly suggests the blocker is
+in the ESP-Hosted C6 slave firmware (the Arduino-ESP32-distributed
+`esp32c6-v2.11.6.bin`), not in the P4 host stack.
+
+Since that is a downstream issue best tracked against
+[esp-hosted-mcu](https://github.com/espressif/esp-hosted-mcu) rather
+than SensESP, stage-2 is left as a follow-up. The scaffolding in this
+example makes it a one-file edit to add the scanner once the slave-fw
+story resolves.
+
+## Hardware
+
+- Board: Waveshare ESP32-P4-WIFI6-POE-ETH (ESP32-P4 ECO2 / rev v1.3
+  ES silicon, 16 MB flash, 32 MB PSRAM, onboard ESP32-C6 over SDIO).
+  The `sdkconfig.defaults` in this directory explicitly selects
+  `CONFIG_ESP32P4_SELECTS_REV_LESS_V3=y` because ES silicon uses a
+  different startup code path than the rev 3.01+ default — without
+  this knob the firmware crashes with an `Illegal Instruction` panic
+  during very early boot. See the comment at the top of
+  `sdkconfig.defaults` for the details.
+- USB-C: same CH340 pass-through to P4 UART0 (GPIO 37/38) as the
+  `p4_ethernet` example. Enumerates as `/dev/ttyACM0` on Linux with
+  recent `cdc-acm`, `/dev/cu.usbmodem*` on macOS.
+- Ethernet: RJ45 to any DHCP-serving switch or router. For isolated
+  bench testing on Linux:
+
+  ```sh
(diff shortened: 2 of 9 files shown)


dirkwa:
Small correction: I just pushed a third commit (`f92e10b48 fix(build): select ESP32-P4 ES silicon in root sdkconfig.defaults`) because the "Tested" claim in the previous commit was incomplete.

When I started work on a follow-up and ran `pio run -e espidf_esp32p4` from the repo root, I noticed the library build was *not* actually reaching the expected `setup()` / `loop()`-only undef state — it was also failing with `undefined reference to _bss_start`, `_bss_end`, `_data_start`, `_heap_start`. That's the same ES (Early Sample) silicon mismatch the example's `sdkconfig.defaults` already handles locally, but I hadn't propagated the fix to the root library build.

The new commit adds `CONFIG_ESP32P4_SELECTS_REV_LESS_V3=y` to the shared root `sdkconfig.defaults`. The symbol is gated on `IDF_TARGET_ESP32P4` in IDF's own kconfig, so it is silently ignored on `espidf_esp32` / `espidf_esp32c3` builds — no regression.

After this fix, all three espidf envs reach the same library-only `setup()/loop()` undef state that the original commit claimed. I verified each one from a clean rebuild:

- `pio run -e espidf_esp32` ✓
- `pio run -e espidf_esp32c3` ✓
- `pio run -e espidf_esp32p4` ✓

Apologies for the noise — I should have caught this before pushing the original commit. The PR description's "Tested" section now accurately describes the state with the new commit included.

dirkwa:
Rebased on current master.

mairas (commented):
## Review

Well-scoped, well-documented PR. The `[espidf_cloud_certs]` extraction is a clean solution to the esp_insights/esp_rainmaker exclusion on P4. The example is honest about stage-1 limitations and the C6 slave firmware blocker. All CI checks pass.

### Minor issues (we'll fix these on our end)

**1. Redundant `disable_wifi()` in the example** (`examples/p4_bluedroid_poc/src/main.cpp`)

`set_ethernet()` already calls `disable_wifi()` internally (`sensesp_app_builder.h:128`). The existing `p4_ethernet` example doesn't call it explicitly either. Since this is example code that people will copy, worth keeping clean.

**2. `.gitignore` patterns match too broadly**

The new `CMakeLists.txt`, `src/CMakeLists.txt`, and `dependencies.lock` entries lack a leading `/`, so they match anywhere in the tree — not just the repo root where pioarduino generates them. `sdkconfig.*` matching tree-wide is fine (espidf examples also generate these), but the CMake/lock patterns should be scoped to root:

```gitignore
/CMakeLists.txt
/src/CMakeLists.txt
/dependencies.lock
sdkconfig.*
!sdkconfig.defaults
```

### Observations (non-blocking)

- **CI coverage**: The espidf envs still have no CI — pre-existing, not introduced here. Your offer to help debug the CI sdkconfig issue is appreciated, we may take you up on it.
- **Root sdkconfig.defaults comment** is quite verbose (11 lines for 2 config lines). The same info is in the PR description and the example's sdkconfig — a shorter comment would suffice.

### What looks good

- `[espidf_cloud_certs]` extraction is correctly motivated and zero-regression for existing envs
- Partition table math checks out
- `esp_hosted` pin to 2.12.3 with rationale
- ES silicon config in shared sdkconfig.defaults is safe (gated on `IDF_TARGET_ESP32P4`)
- Honest scoping — build-mode only, BLE follow-up with clear blocker explanation

Overall this looks good to merge with the minor fixes above. We'll apply those ourselves.

PR #944, opened 2026-04-10
fix: preserve user-configured hostname across reboots

dirkwa:
## Summary

- `set_hostname()` in the builder unconditionally overwrites the persisted hostname on every boot, reverting any hostname set via the web UI back to the compile-time default
- Only apply the builder's hostname when the persisted value still equals the factory default (`"SensESP"`), preserving user customizations

## Test plan

- [x] Set hostname via `builder.set_hostname("my-device")` — first boot uses "my-device"
- [x] Change hostname via web UI to "custom-name", reboot — hostname stays "custom-name"
- [x] Erase NVS, reboot — hostname reverts to builder's "my-device"

### DIFF
--- a/src/sensesp_base_app.h
+++ b/src/sensesp_base_app.h
@@ -135,10 +135,10 @@ class SensESPBaseApp {
 
   ~SensESPBaseApp() { instance_ = nullptr; }
 
-  void init_hostname() {
+  void init_hostname(const String& default_hostname = kDefaultHostname) {
     hostname_ = std::make_shared<PersistingObservableValue<String>>(
-        kDefaultHostname, "/system/hostname");
-    ConfigItem(hostname_);  // Make hostname configurable
+        default_hostname, "/system/hostname");
+    ConfigItem(hostname_);
   }
 
   /**
@@ -165,10 +165,14 @@ class SensESPBaseApp {
   std::shared_ptr<Filesystem> filesystem_;
 
   const SensESPBaseApp* set_hostname(String hostname) {
+    // Pass the builder's hostname as the default for the
+    // PersistingObservableValue constructor. If a user has
+    // configured a different hostname via the web UI, load()
+    // in the constructor will override this default — no
+    // special-case guard needed.
     if (!hostname_) {
-      init_hostname();
+      init_hostname(hostname);
     }
-    hostname_->set(hostname);
     return this;
   }
 };


mairas:
## Configuration Priority Issue

The PR identifies a real bug — `set_hostname()` unconditionally clobbers user-configured hostnames on every boot. But the fix breaks SensESP's established configuration priority pattern.

### How SensESP Configuration Works

Every `PersistingObservableValue` follows this sequence:

1. **Constructor** sets the hard-coded default
2. **`load()`** runs in the constructor, overwriting the default with any persisted user config
3. **Runtime changes** (web UI) call `set()`, which updates the value and saves to flash

The invariant: **persisted user configuration wins over hard-coded defaults.** This happens automatically — `PersistingObservableValue`'s constructor calls `load()` before anything else touches the value.

### The Real Bug

The builder calls `set_hostname()` *after* `init_hostname()` has already constructed the `PersistingObservableValue` (which loaded the persisted value). `set_hostname()` then calls `hostname_->set(hostname)`, which:
1. Overwrites the just-loaded user config
2. **Destroys the persisted config** (because `PersistingObservableValue::set()` calls `save()`)

So it's not just ignoring user config at runtime — it's erasing it from flash on every boot.

### Why This Fix Has Problems

The PR compares against `kDefaultHostname` (`"SensESP"`), but this has an edge case: if a user deliberately sets their hostname to `"SensESP"` via the web UI, the guard sees `"SensESP" == "SensESP"` and overwrites it with the builder value. The user's intentional choice is lost.

More fundamentally, the fix is in the wrong layer. The correct approach would follow the same pattern every other configurable object uses: **don't call `set()` after `load()` has already restored user config.** The builder's hostname should become the *default* passed to `PersistingObservableValue`'s constructor, before `load()` runs — rather than being applied after the fact with a heuristic string comparison.

A cleaner fix would restructure so that `set_hostname()` provides the default to `init_hostname()`, and `init_hostname()` passes it to the `PersistingObservableValue` constructor. That way `load()` naturally overrides it if persisted config exists, with no special-case guard needed.

dirkwa:
Good point — fixed. `set_hostname()` now passes the builder's hostname as the default to `init_hostname()`, which feeds it to the `PersistingObservableValue` constructor before `load()` runs. No `set()` call, no string comparison heuristic. Persisted config wins naturally.

PR #945, opened 2026-04-11
fix(signalk): clear stale polling href on 404 response

dirkwa:
## Summary

- When polling a persisted access-request href, a 404 means the server doesn't recognize the request (server restarted, device moved to a different server, or security disabled)
- Previously the client retried the stale href indefinitely, preventing connection
- Now 404 is treated the same as 500: clear `polling_href_` and start a fresh access-request flow on the next connect cycle

## Test plan

- [x] Device with stale `polling_href` pointing to a server with security disabled reconnects after one retry cycle instead of looping forever
- [x] Device moved to a different server recovers without manual NVS clear


mairas:
Thanks!

PR #946, opened 2026-04-11
feat(ble): BLE gateway with GATT client for ESP32-P4 and ESP32-C5

dirkwa:
## Summary

Adds a BLE gateway stack that bridges BLE devices to signalk-server's BLE provider API. Tested with 3x ESP32-P4 (Ethernet) + 1x ESP32-C5 (WiFi) running simultaneously.

**Library** (`src/sensesp/net/ble/`):
- `BLEProvisioner` abstract base with scanning + GATT client interface
- `EspHostedBluedroidBLE` for P4+C6 via esp_hosted SDIO (Bluedroid)
- `NativeBLE` for chips with native BT using Bluedroid
- `NimBLEProvisioner` for memory-constrained chips using NimBLE (~15KB less RAM than Bluedroid)
- `BluedroidGATTC` shared GATT client (connect, discover, subscribe, read, write)
- `BLESignalKGateway` service: HTTP POST advertisements + optional control WebSocket (hello, status, GATT commands)
- 4-level scan watchdog (restart, RPC reset, GPIO hard-reset, reboot)
- `enable_control_ws` config flag — disable WS on RAM-constrained chips

**P4 example** (`examples/p4_bluedroid_poc/`):
- Bluedroid + esp_hosted, RMII Ethernet, full WS + GATT support
- 3 GATT slots, ~22 hits/sec at 50% scan duty

**C5 example** (`examples/c5_ble_gateway/`):
- NimBLE + WiFi, POST-only (WS disabled to save RAM)
- ~2 hits/sec at 9% scan duty, heap stable at ~20KB

**Misc**:
- `SKWSClient::get_auth_token()` accessor for credential piggybacking

## Test plan

- [x] 3 P4 gateways on Ethernet running simultaneously, all visible in BLE Manager with full metadata
- [x] 1 C5 gateway on WiFi, POST-only, advertisements flowing to server
- [x] OTA firmware update across all P4 boards
- [x] P4 gateways reconnect automatically after server restart
- [x] GATT slots advertised (3/gateway on P4), server acknowledges
- [x] C5 heap stays above 15KB with WS disabled
- [x] Builds on Arduino-ESP32 (non-IDF) targets (CI)


mairas:
This is a significant addition that I definitely will be using myself on my boat - but I think it should be a library of its own.

I want to avoid having hardware adaptation layers in SensESP whenever they can be avoided. The BLE code has a somewhat weak coupling to the rest of the codebase, pulls in a non-trivial amount of new dependencies, and implements a fairly specific piece of functionality. I think it's better to have this code as a seprarate library that users can depend on if they require the functionality - similar to https://github.com/SensESP/OneWire, for example.

I'm closing this PR. When you have the new repo set up, could you create a new PR to link the library to https://signalk.org/SensESP/pages/additional_resources/?

dirkwa:
Good morning,

glad you like it. Will do - coming soon.

PR #954, opened 2026-04-12
Fix Clear Token button not clearing the auth token

mairas:
## Summary

- Fix the "Clear Token" button in the Signal K Settings panel which did nothing when clicked
- Root cause: Preact's `setConfig()` is asynchronous, so calling the save function immediately after setting state serialized the *old* config (with token still present)
- Pass the modified config directly to the save function, bypassing the async state timing issue
- Also clean up incorrect `setRequestSave: (b: boolean) => void` prop typing to `onSave: (configOverride?: JsonObject) => Promise<void>`

## Test plan

- [ ] Load the Signal K Settings page on a device with a stored auth token
- [ ] Click "Clear Token" and verify the PUT request body contains `token: ""`
- [ ] Verify the Connection Settings and SSL Settings "Save" buttons still work correctly
- [ ] Verify the device requests re-authorization from Signal K server after restart

### DIFF
--- a/frontend/src/pages/SignalK/SignalKSettingsPanel.tsx
+++ b/frontend/src/pages/SignalK/SignalKSettingsPanel.tsx
@@ -85,6 +85,7 @@ function SKConnectionSettings({
   setConfig,
   onSave,
 }: SKConnectionSettingsProps): JSX.Element {
+  const [saving, setSaving] = useState(false);
   const id = useId();
 
   const mdns = config.use_mdns === true;
@@ -156,12 +157,15 @@ function SKConnectionSettings({
             <button
               type="submit"
               className="btn btn-primary"
-              onClick={(e) => {
+              disabled={saving}
+              onClick={async (e) => {
                 e.preventDefault();
-                void onSave();
+                setSaving(true);
+                await onSave();
+                setSaving(false);
               }}
             >
-              Save
+              {saving ? "Saving..." : "Save"}
             </button>
           </form>
         </div>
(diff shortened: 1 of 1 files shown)


mairas (commented):
## Code Review — CE Multi-Persona Review

**Scope:** 1 file, ~40 changed lines
**Reviewers:** correctness, testing, maintainability, project-standards, kieran-typescript, agent-native, learnings-researcher (7 total)

### Root Cause Confirmed

The correctness reviewer verified the diagnosis: `setConfig()` in Preact is asynchronous, so calling `handleSave()` immediately after captured the stale config with the token still present. The fix correctly passes the modified config directly to `onSave(newConfig)`, bypassing the timing issue.

### Findings

No P0 or P1 issues. Two minor P3 advisories:

| # | Location | Issue | Reviewers | Route |
|---|----------|-------|-----------|-------|
| 1 | `SignalKSettingsPanel.tsx:161` | `void onSave()` discards Promise — fire-and-forget is intentional but errors won't surface at click site | kieran-typescript, maintainability | advisory |
| 2 | `SignalKSettingsPanel.tsx:15` | `configOverride?` is an implicit contract — callers must know when to pass it vs rely on state | maintainability, kieran-typescript | advisory |

Both are minor design observations. Finding #1 is not a regression — the old code also discarded the return value with `setRequestSave(true)` and worse typing. Finding #2 is the pragmatic minimal fix for the bug.

### Pre-existing Issues (informational, not blocking)
- Save buttons have no loading/disabled state (unlike the TOFU Reset button which manages `setResetting`)
- Event handlers (`handleMDNSChange`, `handleSSLChange`, `handleVerifyChange`) lack explicit type annotations
- Zero frontend test infrastructure despite Jest in package.json

### Verdict

**Ready to merge.** The fix correctly addresses the async state timing bug with the simplest correct approach. No regressions, type safety improved, prop naming cleaned up.

🤖 Review by Claude Code (CE multi-persona review)

PR #960, opened 2026-05-19
fix(http): raise HTTPServer max_open_sockets to 13

dirkwa:
## Summary

`esp_http_server` defaults `config_.max_open_sockets = 7`, of which 3 are reserved internally by the IDF httpd implementation — leaving only **4 simultaneous client sockets**.

Modern browsers open 6+ parallel TCP connections when loading the embedded SensESP SPA (HTTP/1.1 module-federated JS/CSS chunks). The extras get TCP-RST, the React app's parallel `/api/*` fetches fail, and the UI is left stuck on its initial "device is restarting" state with broken CSS — even though the backend is healthy.

This raises the default to **13** (so 10 client sockets after the 3 reserved), exposed as an overridable macro `HTTP_SERVER_MAX_OPEN_SOCKETS` consistent with the existing `HTTP_SERVER_STACK_SIZE` and `HTTP_DEFAULT_PORT` pattern.

## Dependency note

The HTTPServer setting alone is not sufficient — `CONFIG_LWIP_MAX_SOCKETS` must be ≥ `HTTP_SERVER_MAX_OPEN_SOCKETS + 3` (default 10 is too low for max_open_sockets=13). A header comment documents this. Users keeping the IDF default of 10 will need to either lower `HTTP_SERVER_MAX_OPEN_SOCKETS` via build flag or raise `CONFIG_LWIP_MAX_SOCKETS` in their sdkconfig.

## Test plan

- Tested on an ESP32-P4 (Waveshare Touch-LCD-7B) running esp_hosted SDIO WiFi + a SensESP-based cockpit app with embedded SPA.
- Before the patch: web UI half-loaded with no CSS, stuck on "device is restarting", `/api/*` requests returned empty replies. `curl` parallel burst of 10 → 0 succeed.
- After the patch (with `CONFIG_LWIP_MAX_SOCKETS=16`): web UI loads fully, all tabs work, `/api/info`, `/api/routes`, `/api/config` all return 200. `curl` parallel burst of 10 → 6-10 succeed.

### DIFF
--- a/src/sensesp/net/http_server.h
+++ b/src/sensesp/net/http_server.h
@@ -24,6 +24,19 @@ namespace sensesp {
 #define HTTP_DEFAULT_PORT 80
 #endif
 
+// Maximum simultaneous client sockets the HTTP server will accept.
+// The default esp_http_server value (7) is too low for modern browsers,
+// which open 6+ parallel connections to fetch JS/CSS bundles of the
+// embedded SPA — extras get TCP-RST and the web UI hangs half-loaded.
+//
+// Note: esp_http_server reserves 3 sockets internally, so the effective
+// client cap is HTTP_SERVER_MAX_OPEN_SOCKETS. Additionally,
+// `CONFIG_LWIP_MAX_SOCKETS` must be >= HTTP_SERVER_MAX_OPEN_SOCKETS + 3
+// or lwIP itself will refuse the extra sockets.
+#ifndef HTTP_SERVER_MAX_OPEN_SOCKETS
+#define HTTP_SERVER_MAX_OPEN_SOCKETS 13
+#endif
+
 #include <ctype.h>
 #include <stdlib.h>
 
@@ -66,6 +79,7 @@ class HTTPServer : public FileSystemSaveable {
     config_.server_port = port;
     config_.stack_size = HTTP_SERVER_STACK_SIZE;
     config_.max_uri_handlers = 20;
+    config_.max_open_sockets = HTTP_SERVER_MAX_OPEN_SOCKETS;
     config_.uri_match_fn = httpd_uri_match_wildcard;
     String auth_realm_ = "Login required for " + SensESPBaseApp::get_hostname();
     load();
--- a/src/sensesp/net/http_server.h
+++ b/src/sensesp/net/http_server.h
@@ -25,16 +25,19 @@ namespace sensesp {
 #endif
 
 // Maximum simultaneous client sockets the HTTP server will accept.
-// The default esp_http_server value (7) is too low for modern browsers,
-// which open 6+ parallel connections to fetch JS/CSS bundles of the
-// embedded SPA — extras get TCP-RST and the web UI hangs half-loaded.
 //
-// Note: esp_http_server reserves 3 sockets internally, so the effective
-// client cap is HTTP_SERVER_MAX_OPEN_SOCKETS. Additionally,
-// `CONFIG_LWIP_MAX_SOCKETS` must be >= HTTP_SERVER_MAX_OPEN_SOCKETS + 3
-// or lwIP itself will refuse the extra sockets.
+// Kept at the esp_http_server default (7, of which 3 are reserved internally).
+// The web UI stalls modern browsers suffered with this default were caused by
+// keep-alive socket starvation, not the cap itself, and are addressed by
+// enabling `lru_purge_enable` (see the HTTPServer constructor) rather than by
+// claiming more of the lwIP socket budget — which the Signal K WebSocket
+// client, mDNS, and SNTP also draw from.
+//
+// Raising this requires `CONFIG_LWIP_MAX_SOCKETS >= HTTP_SERVER_MAX_OPEN_SOCKETS
+// + 3`; the espidf framework defaults that to 10, so values above 7 also need
+// `CONFIG_LWIP_MAX_SOCKETS` raised in sdkconfig.defaults or httpd_start() fails.
 #ifndef HTTP_SERVER_MAX_OPEN_SOCKETS
-#define HTTP_SERVER_MAX_OPEN_SOCKETS 13
+#define HTTP_SERVER_MAX_OPEN_SOCKETS 7
 #endif
 
 #include <ctype.h>
@@ -80,6 +83,15 @@ class HTTPServer : public FileSystemSaveable {
     config_.stack_size = HTTP_SERVER_STACK_SIZE;
     config_.max_uri_handlers = 20;
     config_.max_open_sockets = HTTP_SERVER_MAX_OPEN_SOCKETS;
+    // The embedded SPA is served over HTTP/1.1 keep-alive, so a browser holds
+    // several connections open. Without LRU purge, once every socket slot is
+    // camped by an idle keep-alive connection the server defers (starves) any
+    // new connection — a late chunk, a retried fetch, a second tab — until one
+    // times out. On high-latency links (e.g. esp_hosted SDIO WiFi) that window
+    // is long enough to fail the SPA's parallel /api/* requests outright.
+    // Enabling LRU purge evicts the least-recently-used (idle) socket to admit
+    // the newcomer instead, which the browser simply reopens.
+    config_.lru_purge_enable = true;
     config_.uri_match_fn = httpd_uri_match_wildcard;
     String auth_realm_ = "Login required for " + SensESPBaseApp::get_hostname();
     load();


dirkwa:
CI failure here is in `ReactESP/event_loop.h` (`'std::vector' does not name a template type` / `node_type` missing), not in any file this PR touches:

```
.pio/libdeps/arduino_esp32/ReactESP/src/event_loop.h:143:8: error: 'vector' in namespace 'std' does not name a template type
.pio/libdeps/arduino_esp32/ReactESP/src/event_loop.h:138:41: error: 'node_type' in 'using TimedEventSet = ...' does not name a type
```

Main's last CI run was 2026-04-12 (green); my run today (2026-05-19) failed identically across every job. Looks like a transitive dependency update (ReactESP `^3.2.0` resolves to a newer release, or a newer Arduino-ESP32 toolchain with stricter include requirements) bit-rotted the build. The two-hunk diff in this PR is purely additive: a new `#ifndef HTTP_SERVER_MAX_OPEN_SOCKETS` macro and one `config_.max_open_sockets = …` assignment — nothing that could trigger an STL include failure in another library.

Happy to rebase once main's CI is fixed.

mairas:
Thanks for tracking this down — the symptom is real, and the overridable-macro approach matches the existing `HTTP_SERVER_STACK_SIZE` / `HTTP_DEFAULT_PORT` convention. I dug into the root cause and I think there's a lower-risk fix that targets it more directly. Two blocking concerns with `13` as the default, then a suggestion.

### Root cause: socket starvation under keep-alive, not just a low cap

The default config has `max_open_sockets = 7` and leaves `lru_purge_enable` at its default of `false`. The SPA is served over HTTP/1.1 keep-alive, so a browser opens up to 6 parallel connections and holds them open. With no LRU purge, once those slots are camped, `httpd_accept_conn` defers any additional connection (a late chunk, a retried fetch, a second tab) until a keep-alive connection times out or completes. On a fast LAN that's a sub-second blip; on a high-latency link (your ESP32-P4 + esp_hosted SDIO path) connections stay open long enough that the SPA's parallel `/api/*` calls get starved and fail outright. Same cause, amplified by the link.

That points at **`lru_purge_enable = true`** as the most targeted fix: when a new connection arrives and all slots are full, the server evicts the least-recently-used socket (typically an idle keep-alive one the browser simply reopens) instead of starving the newcomer. It addresses the starvation without raising `max_open_sockets` at all — so no extra memory, no socket-budget pressure, no startup-config dependency. Worth considering whether this alone resolves the P4 case.

### 1. `max_open_sockets = 13` breaks `httpd_start` on espidf-framework builds

`esp_http_server` enforces this at startup (`httpd_main.c`):

```c
if (HTTPD_MAX_SOCKETS < config->max_open_sockets + 3) {   // HTTPD_MAX_SOCKETS == CONFIG_LWIP_MAX_SOCKETS
    ESP_LOGE(...);
    return ESP_ERR_INVALID_ARG;
}
```

So `13` requires `CONFIG_LWIP_MAX_SOCKETS >= 16`. The arduino/pioarduino prebuilt libs ship `16`, so they pass — but exactly (13 + 3 = 16). A pure `framework = espidf` build gets the IDF default of `10` (lwIP `Kconfig`: `default 10`), and SensESP's `sdkconfig.defaults` doesn't override it. On those builds `httpd_start` returns `ESP_ERR_INVALID_ARG` and the server never starts.

It also fails silently — `HTTPServer` logs the error and then registers handlers on the failed handle anyway:

```cpp
esp_err_t error = httpd_start(&server_, &config_);
if (error != ESP_OK) {
  ESP_LOGE(__FILENAME__, "Error starting HTTP server: %s", esp_err_to_name(error));
}
// ... registers URI handlers regardless
```

So espidf users get a device that boots healthy with a dead web UI — worse than the bug being fixed. CI only covers arduino + pioarduino, so it won't catch this. If the cap is raised, `CONFIG_LWIP_MAX_SOCKETS` needs to be set in `sdkconfig.defaults` so the default and its dependency travel together.

### 2. `13` consumes the entire socket budget

13 client + 3 internal = all 16 lwIP sockets on the arduino default. The Signal K WebSocket client (the device's primary function), mDNS, and SNTP draw from the same pool. At idle it's fine, but a burst of parallel browser connections can transiently starve the SK connection. A value that leaves headroom — e.g. `10` (needs `LWIP_MAX_SOCKETS >= 13`, leaves 3 sockets of margin) — avoids claiming the whole pool.

### Suggestion

- Try `lru_purge_enable = true` first; it may resolve the starvation on its own.
- If a cap bump is still wanted, use `10` rather than `13`, and set `CONFIG_LWIP_MAX_SOCKETS` explicitly in `sdkconfig.defaults`.

The macro and the header comment are good as-is.

dirkwa:
You nailed the root cause — it's keep-alive socket starvation, not the cap. I've reworked the PR to enable `lru_purge_enable = true` instead of raising `max_open_sockets`.

**Changes (new commit on top, branch also rebased onto current `main`):**

- `config_.lru_purge_enable = true` — evicts the least-recently-used idle keep-alive socket to admit a new connection, instead of deferring it until one times out. This is the targeted fix for the starvation you described, and the high-latency esp_hosted SDIO path is exactly why the window was long enough to fail the SPA's parallel `/api/*` calls.
- Reverted the `HTTP_SERVER_MAX_OPEN_SOCKETS` default `13 → 7`. You're right that 13 breaks pure-espidf builds (needs `CONFIG_LWIP_MAX_SOCKETS >= 16`, IDF defaults to 10) and claims the whole lwIP budget the SK WebSocket client / mDNS / SNTP share. Keeping 7 stays within the espidf default (7 + 3 = 10), so the fix works on all frameworks with no `sdkconfig` dependency. The macro stays overridable for anyone who wants more sockets, with the `CONFIG_LWIP_MAX_SOCKETS` requirement now documented in the header comment.

So no extra memory, no socket-budget pressure, no startup-config dependency — same as your suggestion.

**On the silent `httpd_start` failure** you flagged (handlers + mDNS registered on a failed handle): that's a real bug but independent of this change — `lru_purge`-only no longer triggers it. I'd rather fix it in a focused follow-up PR than widen this one.

Status: compiles clean on `pioarduino_esp32` (and the earlier CI failure was the ReactESP bitrot, fixed by 3.3.1 in `main`). On-hardware re-test of the lru_purge approach on the P4 is pending — the original P4 validation was against the previous `13 + CONFIG_LWIP_MAX_SOCKETS=16` approach, so I want to confirm the new mechanism on-device before calling it verified.

dirkwa:
Heads-up on the red CI: it's not the change. Both runs on this branch failed in PlatformIO's Library Manager step, before any compilation, on a download from Espressif's component registry:

```
Library Manager: Installing https://components.espressif.com/api/downloads/?object_type=component&object_id=dbc87006-...
Package download failed: Got the unrecognized status code '403' ... (attempts 1/5–4/5)
PackageException: Got the unrecognized status code '403' ...
```

It only hits the `pioarduino` jobs (which fetch ESP-IDF managed components from `components.espressif.com`); the `arduino` builds and all `test` jobs pass, since they use prebuilt libs and never touch that registry. The 403 appears to be a transient window on Espressif's side — `main` and another PR built green on the same jobs ~6h earlier today, so the same matrix passes whenever the registry cooperates. The change itself compiles clean locally on `pioarduino_esp32`.

Will retry once the registry settles; no code issue to address here.

dirkwa:
Dug into the persistent CI red — it's an Espressif **CloudFront edge-specific** outage, not this change and not a transient blip. Sharing the diagnosis since it affects every `pioarduino` build repo-wide, not just this PR.

**Symptom:** all `pioarduino` build + test jobs fail in the Library Manager step downloading the pinned `esp_websocket_client` component:

```
Library Manager: Installing https://components.espressif.com/api/downloads/?object_type=component&object_id=dbc87006-9a4b-45e6-a6ab-b286174cb413
Package download failed: Got the unrecognized status code '403' ... (attempts 1/5–4/5)
PackageException: ... status code '403' ...
```

(`dbc87006…` is `espressif/esp_websocket_client`, pinned via the object-id download URL in `platformio.ini`.) The `arduino`-framework jobs pass — they use prebuilt libs and never fetch from the registry.

**Root cause — it's regional CDN, not the reference:** the same URL returns **200 from some CloudFront edges and 403 from others**. From here it resolves via the Auckland edge (`x-amz-cf-pop: AKL53`) → HTTP 200; the GitHub-hosted runners are in `eastus2` and hit the US-East edge → HTTP 403. So it's an Espressif `components.espressif.com` CDN-edge issue on the US-East PoP, persisting ~7h now (note `main` last went green at 13:46 on 06-04, before it started). Because the download routes through the same CDN regardless of how the component is referenced, switching the `lib_deps` syntax (object-id URL → `espressif/esp_websocket_client` name) almost certainly wouldn't dodge it.

**Possible mitigations (your call — flagging, not changing anything):**
- Wait for Espressif's US-East edge to recover (most likely self-heals) and re-run.
- If repeated CI flakiness from this pin is worth removing structurally: pin `esp_websocket_client` to the IDF-bundled copy / a mirror, or cache it, so CI doesn't depend on the live registry CDN for a component that ships with ESP-IDF anyway.

The #960 change itself compiles clean on `pioarduino_esp32` locally and all `test` jobs pass when the registry cooperates. Happy to help with whichever mitigation you prefer.

dirkwa:
@mairas 

So, finally the CDN is working again.

PR #963, opened 2026-05-26
chore: add stale workflow for awaiting-response issues

msallin:
Adds the `actions/stale` workflow that marks issues labeled `awaiting response` as stale after 30 days of inactivity and closes them 30 days after that (60 days total).

- Only acts on issues labeled `awaiting response` (`any-of-issue-labels`); other issues are never touched.
- Exempts `good first task` and `help wanted`.
- Never touches pull requests (`days-before-pr-stale: -1`).
- Daily schedule (`17 1 * * *`) plus `workflow_dispatch` for manual runs.
- 10-minute timeout, single-flight via `concurrency: stale`.

Identical to the workflow already in nmea0183-signalk / charts-plugin / signalk-to-nmea0183 / course-provider-plugin / aisreporter / nmea0183-utilities / signalk-derived-data. Rolling it out org-wide so the `awaiting response` convention works consistently.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,42 @@
+# Marks issues labeled "awaiting response" as stale after 30 days of inactivity
+# and closes them 30 days after that (60 days total). Issues labeled
+# "good first task" or "help wanted" are exempt. Pull requests are never
+# touched. The workflow runs daily and can also be triggered manually.
+name: Stale issues
+
+on:
+  schedule:
+    - cron: '17 1 * * *'
+  workflow_dispatch:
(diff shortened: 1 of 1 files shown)


mairas:
I sympathize with the idea of having consistent repo settings in SignalK, but I would prefer not to have SensESP issues auto-labeled stale or closed. The project is somewhat, ahem, short-staffed, and occasionally it takes quite a bit of time until I get to fix open issues even if they were valid. Also, the amount of inbound issues isn't so great that it would cause huge problems like on the busier siblings.

I'll be a brute and just close this.

Issue #983, opened 2026-06-15
Log viewer — Unit 5: rebuild bundle + footprint check


mairas:
Part of #978.

**Goal:** Regenerate the embedded frontend and confirm the flash cost is acceptable.

**Requirements:** supports R12 (page must ship in the embedded bundle); enforces the flash-footprint constraint.

**Dependencies:** Unit 4 (#978).

**Files:**
- Modify (generated): `src/sensesp/net/web/autogen/frontend_files.h` via the existing Vite build → `convert_html.py` pipeline; do not hand-edit.

**Approach:**
- Run the frontend build so the gzipped bundle (now including LogPage) regenerates the autogen header. Compare firmware `.bin` size and app-partition headroom before/after; record the delta.
- If the gzipped delta is materially larger than the ~5–15 KB estimate, investigate (stray import, duplicated dep) before proceeding.

**Patterns to follow:** existing `extra_script.py` / `scripts/convert_html.py` flow.

**Test scenarios:** none — build/asset regeneration. Behavioral tests live in Unit 4.

**Verification:** a release build links and flashes; the Log pill loads the new page from the embedded bundle; firmware size delta within expectation and the app partition (e.g. `min_spiffs.csv` boards) still has headroom.

Issue #982, opened 2026-06-15
Log viewer — Unit 4: LogPage frontend (Preact)


mairas:
Part of #978.

**Goal:** A built-in Log page that tails `/api/log` with coloring, autoscroll, and a DOM cap.

**Requirements:** R12–R17.

**Dependencies:** Unit 2 (endpoint), Unit 3 (route advertised) — #978.

**Files:**
- Create: `frontend/src/pages/Log/index.tsx`, `frontend/src/pages/Log/LogLines.tsx`
- Modify: `frontend/src/App.tsx` (add `LogPage` to `KNOWN_COMPONENTS`), `frontend/src/config.ts` (add `log_path: "/api/log"`)
- Test: MVP default is **on-device/manual** verification (the chain is proven on-device anyway). A frontend unit-test harness is **not** wired (jest is in devDeps but there's no script/config; Preact+Vite+pnpm would need vitest or jest+jsdom) — establishing one is generic infrastructure, better scoped as its own follow-up. If added later, `frontend/src/pages/Log/LogLines.test.tsx` covers the scenarios below.

**Approach:**
- `index.tsx` mirrors `pages/Status/index.tsx` (AppPage + PageHeading + PageContents).
- `LogLines.tsx` borrows `InfoGroups.tsx`'s poll structure (`setInterval` 500 ms, fetch, loading state) but **must diverge from its error handling** — `InfoGroups` halts polling permanently on first error; the Log page must keep polling through failures and auto-resume (a single WiFi blip must not kill the tail).
- Cursor + reboot: send `since=<cursor>`, advance from `next`. Track `session`; on change (or `next < since`) reset the cursor and surface a "device restarted" notice rather than stalling.
- Capped client array (1000 lines, drop oldest). Color each line by its leading level letter via Bootstrap text classes (E→`text-danger`, W→`text-warning`, I→`text-body`, D/V→`text-secondary`); keep the level letter visible as a non-color cue; legible warning contrast on a light background. Monospace scroll container. No new deps, no parsing beyond the first letter.
- Multi-line entries (backtraces): a line whose first char isn't a level letter inherits the previous line's class.
- Autoscroll: pin to bottom; if scrolled up (within ~40 px = pinned), stop following; resume at bottom.
[...]

Issue #981, opened 2026-06-15
Log viewer — Unit 3: wire LogBuffer + route into app startup


mairas:
Part of #978.

**Goal:** Initialize capture early and advertise the Log page to the UI.

**Requirements:** R1 (early install), R12 (route entry).

**Dependencies:** Units 1–2 (#978).

**Files:**
- Modify: `src/sensesp_app.h` (startup: create LogBuffer + install hook before/with HTTP handler registration; register the `/api/log` handler defined in Unit 2).
- Modify: `src/sensesp/net/web/base_command_handler.cpp` (`add_routes_handlers()`: add `RouteDefinition("Log", "/log", "LogPage")`).

**Approach:**
- Install the capture hook as early as practical in app construction (after heap is available) to catch the most startup logging; UART output preserved via the chained handler regardless.
- Append the Log route after "System" in the static route vector — must **not** be first (the first route is the default redirect target).

**Patterns to follow:** existing `add_base_app_http_command_handlers()` call site and the static route list in `add_routes_handlers()`.

**Test scenarios:** none — wiring/registration only; behavior exercised by Units 1, 2, 4. Verified on-device.

**Verification:** `/api/routes` includes the Log entry; the nav pill appears after System; logs emitted during startup are retrievable once WiFi/HTTP is up.

mairas:
Done in #984 (merged as b4dd1f5b). LogBuffer install + Log route wired into SensESPApp startup.

Issue #980, opened 2026-06-15
Log viewer — Unit 2: /api/log HTTP endpoint


mairas:
Part of #978.

**Goal:** Serve buffered lines by cursor as minimal JSON, gated by the global dispatcher auth (no `check_origin`).

**Requirements:** R8–R11.

**Dependencies:** Unit 1 (#978).

**Files:**
- Modify: `src/sensesp/net/web/base_command_handler.cpp`, `base_command_handler.h` (or create `src/sensesp/net/web/log_command_handler.{h,cpp}` if `base_command_handler.cpp` grows unwieldy).
- Test: covered by the Unit 1 on-target Unity tests for the snapshot contract; handler wiring verified on-device.

**Approach:**
- Define the `/api/log` `HTTPRequestHandler` for `1 << HTTP_GET` (startup registration owned by Unit 3), building the JSON response like the `/api/info` handler. **No per-handler access control: do not add `check_origin`** (it's anti-CSRF for POSTs and gives no read protection). `/api/log` is gated by the global dispatcher auth at parity with `/api/info` and config; the soft-AP path is auth-exempt for all alike.
- Parse `since` from the query string (default 0 / absent → full buffer).
- Call the Unit 1 snapshot-since API; build `{session, next, gap, lines[]}` with ArduinoJson `JsonDocument` + `serializeJson`; `httpd_resp_set_type(application/json)`.
- Keep the response minimal — array of raw strings, no per-line objects.

**Patterns to follow:** `/api/info` handler in `base_command_handler.cpp` (JSON build + `httpd_resp_sendstr`; note `/api/info` has no `check_origin`).

**Test scenarios:**
- Happy: GET with no `since` returns the full retained buffer + next cursor.
- Happy: GET with `since`=next returns empty `lines`, unchanged cursor.
- Edge: malformed/non-numeric `since` treated as 0 (full buffer), no crash.
- Error: with `auth_required_` enabled, an unauthenticated request gets 401 from the global dispatcher (not handler logic); no `check_origin` involved.
- Integration: after `ESP_LOGW`, a subsequent GET includes that line with its `W` level letter intact.

**Verification:** `curl`/browser GET against a running device returns growing JSON deltas; gap flag flips true after the client stalls past the retention window.

PR #984, opened 2026-06-15
feat(log): web UI log viewer

### LINKED ISSUE #978: feat: SensESP web UI log viewer (MVP)
## Overview

Surface the device's `ESP_LOGx` output in the existing SensESP web UI so logs can be read over WiFi without a USB serial cable. A firmware-side capture hook feeds log lines into a small bounded RAM buffer; a polling HTTP endpoint serves new lines by cursor; a new built-in "Log" page in the Preact UI tails them with level coloring and autoscroll. RAM-only, no flash persistence.

This is a **developer-first live tail** (serial-monitor-equivalent over WiFi). A non-programmer boat owner is a secondary, opportunistic viewer — the MVP does not try to make raw logs actionable for them. It is a *live* tail, not a post-hoc fault-capture tool: the ~2 s window shows what's happening now.

Planned via brainstorm → plan → multi-persona document-review (2 passes). Implement in a real SensESP clone.

## Requirements

- **R1–R3** Chained `esp_log_set_vprintf` hook feeding every `ESP_LOGx` line into a RAM buffer; preserves UART output; never recurses; safe from arbitrary task contexts.
- **R4–R7** RAM-only bounded buffer; ~2 s retention (primary bound); hard 200-line cap (spam guard); monotonic per-line sequence as polling cursor.
- **R8–R11** HTTP GET `/api/log?since=<seq>` → lines newer than cursor + next cursor + session id; no-cursor returns the whole buffer; signal a gap when the cursor predates the oldest retained line; detect reboot (session change / `next < since`). Access control: none of its own — gated by SensESP's **global** dispatcher auth like every endpoint; no `check_origin` (anti-CSRF for POSTs, not read protection).
- **R12–R17** Built-in "Log" page (in `KNOWN_COMPONENTS` + `/api/routes`, after "System"); 500 ms polling; raw text colored by level; tail-follow that pauses on scroll-up; 1000-line browser DOM cap; liveness + gap indication.
- **R18** Network buffer captures all log levels (UART parity). Wider exposure than `/api/info` (request URIs, SK auth-flow detail, SSIDs, library/user content), readable unauthenticated over the soft-AP. Accepted for the developer-first MVP; mitigated only by the global device auth (**defaults off**); documented.

## Scope boundaries

- No flash/NVS persistence; no across-reboot/post-mortem log.
- Not a post-hoc fault-capture tool (~2 s live tail only).
- Not aimed at making raw logs actionable for non-programmers (no copy/share, no interpretation aids).
- No search, filtering, copy, or download in the MVP.
- No SSE/WebSocket — cursor polling only.
- No parsed/columnar rendering — raw colored lines only.
- No capture of pre-WiFi early boot, `ESP_EARLY_LOGx`, or `ets_printf`/`esp_rom_printf` (bypass the hook by design).

## Key technical decisions

- **Flash footprint governs.** The frontend bundle is gzipped + embedded in the app partition; JS is the dominant cost. No new frontend deps; reuse existing components/Bootstrap classes; C++ reuses ArduinoJson + existing handler machinery.
- **Cursor polling over SSE/WebSocket** — avoids tying up an `esp_http_server` socket per viewer; 500 ms latency is invisible here; extensible to SSE later.
- **Bounded dynamic line list (heap), not a pre-reserved arena** — normal volume is a few lines/s, so only a handful are ever live; short-lived uniform allocations are fragmentation-benign under TLSF, matching SensESP/ReactESP's existing churn. 200-line cap + per-line ~128 B cap are a runaway guard (bounds ~25 KB), not a reserved size.
- **Concurrency: FreeRTOS mutex, copy under the lock** (SensESP precedent). Push formats outside the lock; snapshot copies the matching lines into a local container *inside* the lock (~25 KB bound) then serializes after release — never hands back raw record pointers (would risk use-after-free on concurrent eviction). Per-task re-entrancy guard; chained vprintf uses `va_copy`.
- **Minimal payload** `{session, next, gap, lines[]}` — raw text carries level/time/tag; `gap` when `since` predates oldest; `session` (per-boot id) lets the client detect a reboot/sequence reset.
- **Load-bearing build flag** — capture works only because SensESP compiles with `-D USE_ESP_IDF_LOG` (else Arduino-ESP32 reroutes `ESP_LOGx` to `log_printf`, bypassing the hook). Emit a compile `#warning` if absent.

## High-level flow

```
ESP_LOGx (any task) → vprintf hook → [chain to prev → UART] + [mutex: append to bounded list]
bounded list (heap, evict age>2s | 200-line cap)  ← GET /api/log?since=N (global auth)
  → mutex: copy matching lines → {session,next,gap,lines} (ArduinoJson) → browser
browser: poll 500ms, append, color by level, autoscroll unless scrolled-up, cap 1000 lines,
         detect session change → reset cursor + "device restarted"
```

## Risks

| Risk | Mitigation |
|------|------------|
| Downstream build drops `-D USE_ESP_IDF_LOG`, silently disabling capture | Compile `#warning`; document the dependency |
| Lock held too long / snapshot use-after-free | FreeRTOS mutex; push formats outside lock; snapshot copies bounded set inside lock, serializes after release; never returns raw record pointers |
| High log volume floods poll / grows list | Age + 200-line caps; gap flag signals missed lines |
| Log path allocates — alloc failure | Drop the line on failure; never throw from the log path |
| Frontend bundle growth eats flash | No new deps; measure delta and gate (Unit 5) |
| Unauthenticated read of logs | Gated by global dispatcher auth at parity with config; residual: soft-AP path auth-exempt for all endpoints; **auth defaults off** — recommend enabling it |
| Logs may contain secrets (request URIs, SK auth-flow, SSIDs) — wider surface than `/api/info` | Accepted for MVP (all levels, no level-gate); documented; recommend enabling auth + not logging secrets |
| Recursive logging from the hook | Per-task re-entrancy guard; hook path performs no `ESP_LOGx` |

## Implementation units

- [ ] #979 — Unit 1: LogBuffer capture hook + bounded buffer (firmware)
- [ ] #980 — Unit 2: `/api/log` HTTP endpoint (firmware)
- [ ] #981 — Unit 3: wire LogBuffer + route into app startup (firmware)
- [ ] #982 — Unit 4: LogPage frontend (Preact)
- [ ] #983 — Unit 5: rebuild bundle + footprint check

---
Planning artifacts (local): `_local/brainstorms/2026-06-15-sensesp-web-log-viewer-requirements.md`, `_local/plans/2026-06-15-001-feat-web-log-viewer-plan.md`.

mairas:
## What

Adds a Log page to the SensESP web UI so device logs are visible in a browser without a USB serial connection.

## Why

Until now the only way to see `ESP_LOGx` output was over USB serial. Once a device is deployed on a boat that's rarely practical. This surfaces the same log stream in the existing web UI.

## How

- **Capture** (`log_buffer.{h,cpp}`): a `vprintf` hook mirrors ESP-IDF log output into a bounded, mutex-protected ring buffer. Reads copy under lock and hand back a cursor, so the writer can keep advancing without risking use-after-free.
- **Endpoint** (`base_command_handler.cpp`): a cursor-based `GET /api/log` returns buffered lines as JSON. Access control mirrors `/api/info` — no separate auth.
- **Startup wiring** (`sensesp_app.h`): log capture is installed during `SensESPApp` initialization.
- **Frontend** (`pages/Log/`): polls `/api/log` with the cursor, renders lines with level coloring and autoscroll, detects device reboots (cursor reset) and shows a single Bootstrap toast, plus a jump-to-latest control when the user has scrolled up.
- **Bundle**: embedded `frontend_files.h` regenerated from the production build.

## Notes

- Log content is as sensitive as serial output; access parity with `/api/info` is intentional and documented.
- The buffer lives in RAM and is bounded; oldest lines are dropped under pressure.

Closes #978, #979, #980, #981, #982, #983.


mairas:
## Code review — web UI log viewer

**Intent:** Capture `ESP_LOGx` via a vprintf hook into a bounded, mutex-protected ring buffer; expose it through cursor-based `GET /api/log` (access parity with `/api/info`); render it in a Preact Log page with level coloring, autoscroll, reboot detection, and jump-to-latest. Memory-safe under concurrent access; no regression to the existing web UI.

**Reviewers:** correctness, testing, maintainability, project-standards (always-on) + security, api-contract, reliability, adversarial (conditional — public endpoint, concurrency, 769-line diff). The generated `frontend_files.h` was excluded from review.

**Verdict: Ready with fixes.** One real security exposure (P1) should be resolved before merge; the rest is hardening and cleanup that can land here or as follow-ups. api-contract found no server/client mismatch.

> Note: a couple of the deeper findings cite line numbers from their working read that don't all line up exactly; locations below are given by symbol/behavior.

### P1 — High

| # | Location | Issue | Reviewer(s) | Conf | Route |
|---|----------|-------|-------------|------|-------|
| 1 | `signalk/signalk_ws_client.cpp` (`poll` response log, ~L1023; also ~L808/L914/L309) | **SK auth token can leak via `/api/log` when DEBUG logging is on.** The poll-response payload is logged verbatim at DEBUG, and that same payload carries `accessRequest.token` (read at ~L1057). The new vprintf hook captures every `ESP_LOGx` line, and `/api/log` now serves the buffer over HTTP. The web server defaults to `auth_required_ = false`, so on a typical deployment this is unauthenticated. The DEBUG line predates this PR, but making the log buffer network-readable is what turns a serial-only sink into a network exposure — so it belongs in this PR. | security | 0.72 | **fix-here** (pre-existing). Redact the token at the log site, matching the existing `%.8s...[redacted]` pattern at L750; or log only non-secret fields. |
| 2 | `system/log_buffer.cpp` (`push_line` / vprintf trampoline) | **Capture path takes a blocking mutex (`portMAX_DELAY`) — illegal if ever invoked from an ISR.** Mitigating reality: standard `ESP_LOGx` is not ISR-safe, and ISR/early logging uses the DRAM/early path that bypasses the vprintf hook, so practical risk is low. A cheap `xPortInIsrContext()` guard (skip capture, still forward to UART) removes the foot-gun and matches the header's "ISR-path bypasses the hook" claim, which is currently unenforced. | reliability | 0.70 | gated — add guard. |

### P2 — Moderate

| # | Location | Issue | Reviewer(s) | Conf | Route |
|---|----------|-------|-------------|------|-------|
| 3 | `frontend/.../Log/LogLines.tsx` (`setInterval` poll) | **Overlapping polls + no backoff.** `poll()` runs every 500ms regardless of whether the prior request resolved; `fetch` has no timeout. On a slow device, two in-flight polls read the same cursor and the same delta gets appended twice (distinct React keys → duplicate rows). When the device is unreachable, polling continues at full rate with no backoff. | adversarial, reliability | 0.82 | gated — self-scheduling `setTimeout` gated on completion + backoff after N failures. |
| 4 | `system/log_buffer.cpp` (constructor `session_id_ = esp_random()`) | **Session id drawn before RF is up may repeat across reboots.** The constructor runs at `SensESPApp` construction, before `install()` and WiFi/BT bring-up; `esp_random()` is only truly random once the RF subsystem is active. If two boots draw the same id and the client cursor was 0/null, the reboot goes undetected (no toast, no buffer reset). | adversarial | 0.62 | manual — draw the id in `install()` or seed from boot count (NVS) / MAC. |
| 5 | `system/log_buffer.cpp` (`snapshot_since`) | **Untested core branches:** empty-buffer gap (all lines age-evicted), post-reboot stale cursor (`since > next_seq_`), and the `prune_locked` `now_ms < timestamp` underflow guard. These are exactly the reboot/gap semantics the UI depends on; a sign/comparison regression would silently hide data loss. | testing | 0.80 | manual — add on-target Unity cases. |
| 6 | `system/log_buffer.cpp` (`push_line`) | Heap allocation (`std::string` + deque node) per captured line under the mutex. Bounded, but under sustained high log volume it churns/fragments the heap at log rate. | reliability | 0.65 | advisory — acceptable for a ~200-line tail; consider a fixed char ring if limits grow. |
| 7 | `frontend/.../Log/LogLines.tsx` (gap notice) | Gap toast is a single sticky boolean; repeated bursts after the user dismisses it go unnoticed, so the tail looks continuous when it isn't. | adversarial | 0.75 | advisory — prefer an inline "… N lines dropped" separator. |

### P3 — Low

| # | Location | Issue | Reviewer(s) | Conf | Route |
|---|----------|-------|-------------|------|-------|
| 8 | `src/sensesp_app.h` (include block) | New `sensesp/system/log_buffer.h` include breaks the clang-format alphabetical sort (`SortIncludes: true`); belongs in the `sensesp/system/` group. | project-standards | 0.88 | **safe_auto** — move it / run clang-format. |
| 9 | `system/log_buffer.h` (`push_line` 2-arg overload) | Unused two-arg `push_line` overload — the trampoline and tests both use the three-arg form. Dead API surface. | maintainability | 0.82 | **safe_auto** — drop it. |
| 10 | `net/web/base_command_handler.cpp` (`since` parse) | `since` query has no overflow/length/non-numeric validation (`strtoul`, 16-byte buffer). Blast radius is small (a client only corrupts its own view), but worth `errno`/`endptr` checks. | correctness, adversarial, api-contract | 0.66 | safe_auto — validate and treat invalid as no-cursor. |
| 11 | `system/log_buffer.h` (`session_id()` getter) | Accessor used only by a test that could assert against `snapshot.session_id`. | maintainability | 0.62 | advisory. |
| 12 | `system/log_buffer.cpp` (truncation) | Line truncation can split a UTF-8 sequence, yielding a replacement char in the UI. Cosmetic. | correctness | 0.60 | advisory. |
| 13 | `frontend/.../Log/LogLines.tsx` (autoscroll) | Autoscroll effect can fight user scroll under high volume due to scroll-event vs `setLines`-commit ordering. | adversarial | 0.60 | advisory — re-check `atBottom` from live DOM before snapping. |
| 14 | `net/web/base_command_handler.cpp` (response) | `Content-Type: application/json` lacks `charset=utf-8`; non-ASCII log bytes may corrupt a line on `response.json()`. | api-contract | 0.60 | advisory. |
| 15 | `net/http_server.cpp` (`Handling request: %s`) | Request URIs (incl. query strings) logged at INFO are now web-exposed via `/api/log`. No secret-bearing GET param exists today; latent risk if one is ever added. | security | 0.60 | advisory (pre-existing) — document; avoid secrets in GET params. |

### Coverage

- **Concurrency safety** — the stated "memory-safe under concurrent access" goal has no automated test; all buffer tests are single-threaded (hard to test deterministically on-target, but the core claim rests on review, not verification).
- **vprintf trampoline** — `va_copy`, 256-byte truncation, null-instance guard, handler chaining: zero coverage (tests drive `push_line` directly).
- **HTTP handler** — query parse / JSON shape / null-buffer fallback: no coverage.
- **Frontend** — reboot detection, gap notice, overlapping polls, jump-to-latest: no tests (no frontend test infra in scope).
- No findings suppressed below threshold.

### Suggested order
1. Redact the SK token at the log site (#1) — before merge.
2. ISR guard (#2) + the two safe_auto cleanups (#8, #9).
3. Frontend overlapping-poll fix (#3) and session-id entropy (#4).
4. File the remaining P2/P3 hardening + test gaps as follow-up issues if not done here.

mairas:
## Review findings addressed

Pushed three fix commits (`15b508222`, `dcc697962`, `62424cb77`) on top of the feature. Verified against `pio test --without-uploading -e pioarduino_esp32` (the CI build) — `test_log_buffer` and the full suite compile clean; frontend type-checks (pre-existing errors only) and rebuilds, bundle regenerated.

| # | Finding | Resolution |
|---|---------|-----------|
| 1 | SK token leak via `/api/log` (P1) | Poll-response DEBUG log now logs HTTP status + byte count, never the payload. |
| 2 | ISR-unsafe blocking mutex (P1) | Capture skipped via `xPortInIsrContext()`; line still forwarded to UART. |
| 3 | Overlapping polls / no backoff (P2) | Self-scheduling loop gated on completion, abort timeout, exponential backoff past the reconnect threshold. |
| 4 | Session id reuse across reboots (P2) | Drawn in `install()` and XOR'd with the boot timer. |
| 5 | Untested buffer branches (P2) | Added tests: empty-buffer gap, post-reboot stale cursor, prune underflow guard, bare-newline drop. |
| 8 | Include order (P3) | Sorted into the `sensesp/system/` block. |
| 9 | Unused `push_line` overload (P3) | Removed. |
| 10 | `since` cursor validation (P3) | Rejects non-numeric / out-of-range values. |

### Deferred to follow-up issues

- **#985** — LogBuffer per-line heap churn (#6) + UTF-8-aware truncation (#12). Current bounded behavior is acceptable; a storage rewrite isn't warranted now.
- **#986** — Log page UX: repeatable gap indicator (#7) + autoscroll/scroll race (#13).
- **#987** — Web log exposure of request URIs / DEBUG payloads (#15) + `charset=utf-8` (#14).

Remaining P3 #11 (`session_id()` getter used only by tests) left as-is — it's a thin, harmless accessor.

PR #991, opened 2026-06-15
fix(web): accept same-origin requests served via the soft-AP IP

### LINKED ISSUE #976: Restart/reset endpoints silently 403 when UI is accessed via the soft-AP IP
## Summary

When the web UI is opened through the device's own soft-AP (`192.168.4.1`), the **Restart** and **Reset** buttons do nothing — no restart, no visible error. The `POST /api/device/restart` (and `/api/device/reset`) request is rejected with HTTP 403 by `check_origin()`, even though it is genuinely same-origin.

## Root cause

`check_origin()` in `src/sensesp/net/web/base_command_handler.cpp` accepts a request only if the `Origin` header contains either the hostname or the **station** IP (`provisioner->local_ip()`). It never considers the **soft-AP** IP (`192.168.4.1`).

```cpp
String hostname = SensESPBaseApp::get_hostname();
if (origin_str.indexOf(hostname) < 0) {
  auto provisioner = SensESPApp::get()->get_network_provisioner();
  String ip = provisioner ? provisioner->local_ip().toString() : String("");
  if (ip.length() == 0 || origin_str.indexOf(ip) < 0) {
    httpd_resp_send_err(req, HTTPD_403_FORBIDDEN, "Cross-origin request rejected");
    return false;
  }
}
```

The device runs the soft-AP and the WiFi station simultaneously. When a user is connected to the device AP and loads the UI from `http://192.168.4.1`, the page is served from `192.168.4.1` and the POST goes to `192.168.4.1` — same origin — but the `Origin` header (`http://192.168.4.1`) matches neither the hostname nor the station IP, so it is wrongly rejected.

On failure the restart handler returns `ESP_FAIL` after `check_origin` has already sent the 403, so the frontend just sees a rejected request and the button appears to do nothing.

## Reproduction

Verified against a SH-ESP32 connected to WiFi (station `10.84.80.164`) while its soft-AP (`192.168.4.1`) is also up:

| `Origin` header | `POST /api/device/restart` |
|---|---|
| `http://gnss-rtk-compass.local` | 200 → restarts |
| `http://10.84.80.164` (station IP) | 200 → restarts |
| `http://192.168.4.1` (soft-AP IP) | **403 → no restart** |

```
curl -X POST -H "Origin: http://192.168.4.1" http://<station-ip>/api/device/restart   # -> 403
```

## Suggested fix

Compare the `Origin` host against the request's own `Host` header rather than an allowlist of hostname + station IP. The `Host` header reflects whatever interface/name the browser actually used (soft-AP IP, station IP, mDNS name, custom DNS), so a same-origin check against it is correct for every access path and removes the fragile hardcoded comparison. The narrower alternative is to also accept the soft-AP IP in the existing check.

Affects `check_origin()` on current `main`; the same helper guards both `/api/device/restart` and `/api/device/reset`.

mairas:
## Why

When the web UI is opened through the device's own soft-AP (`http://192.168.4.1`), the Restart and Reset buttons silently do nothing. `check_origin()` only accepted an `Origin` matching the hostname or the WiFi **station** IP, so a genuinely same-origin POST from the soft-AP IP was rejected with HTTP 403 — exactly during initial configuration, when those buttons matter most.

## What

Compare the `Origin` header's authority (`host[:port]`) against the request's own `Host` header instead of a hardcoded hostname + station-IP allowlist. The `Host` header reflects whatever interface the browser actually used, so the check is correct for every access path: soft-AP IP, station IP, mDNS name, or custom DNS. Requests without an `Origin` header (non-browser clients) are still allowed; the dispatcher's authentication is unchanged.

This is a stricter, standard CSRF check: a cross-origin attacker page sends its own `Origin` (e.g. `evil.com`) which won't match the device `Host`, so destructive POSTs are still rejected. `Origin`/`Host` are browser-controlled forbidden headers that page JS cannot forge.

Fixes #976

### DIFF
--- a/src/sensesp/net/web/base_command_handler.cpp
+++ b/src/sensesp/net/web/base_command_handler.cpp
@@ -14,26 +14,44 @@ namespace sensesp {
 namespace {
 
 /// Reject cross-origin POST requests to destructive endpoints.
+///
+/// Compares the Origin header's authority (host[:port]) against the request's
+/// own Host header. The Host header reflects whatever interface the browser
+/// actually used, so this is same-origin for every access path — soft-AP IP,
+/// station IP, mDNS name, or custom DNS — without a hardcoded allowlist.
+/// Requests without an Origin header (non-browser clients) are allowed.
+///
+/// An Origin or Host that does not fit the buffer is rejected rather than
+/// treated as absent: a legitimate same-origin request to this device is always
+/// short, so an over-long header can only be a forgery attempt and must fail
+/// closed.
 bool check_origin(httpd_req_t* req) {
-  char origin[128] = {0};
-  if (httpd_req_get_hdr_value_str(req, "Origin", origin, sizeof(origin)) ==
-      ESP_OK) {
-    String origin_str(origin);
-    String hostname = SensESPBaseApp::get_hostname();
-    if (origin_str.indexOf(hostname) < 0) {
-      // Check the active provisioner's local IP — works for both WiFi
-      // and Ethernet without depending on the WiFi library directly.
-      auto provisioner = SensESPApp::get()->get_network_provisioner();
-      String ip =
-          provisioner ? provisioner->local_ip().toString() : String("");
-      if (ip.length() == 0 || origin_str.indexOf(ip) < 0) {
-        httpd_resp_send_err(req, HTTPD_403_FORBIDDEN,
-                            "Cross-origin request rejected");
-        return false;
-      }
-    }
+  if (httpd_req_get_hdr_value_len(req, "Origin") == 0) {
+    return true;
   }
-  return true;
+
+  char origin[128] = {0};
+  char host[128] = {0};
+  if (httpd_req_get_hdr_value_str(req, "Origin", origin, sizeof(origin)) !=
+          ESP_OK ||
+      httpd_req_get_hdr_value_str(req, "Host", host, sizeof(host)) != ESP_OK) {
+    httpd_resp_send_err(req, HTTPD_403_FORBIDDEN,
+                        "Cross-origin request rejected");
+    return false;
+  }
+
+  String origin_str(origin);
+  int scheme_end = origin_str.indexOf("://");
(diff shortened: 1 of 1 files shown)


mairas:
## Review synthesis (independent security + correctness reviewers)

**One real defect found and fixed in this PR.**

🔴 **CSRF fail-open on over-long Origin header (high).** Both reviewers flagged it independently. `httpd_req_get_hdr_value_str` returns `ESP_ERR_HTTPD_RESULT_TRUNC` (not `ESP_OK`) when a header exceeds the 128-byte buffer. The first revision treated that identically to "Origin absent" → allowed. An attacker hosting the CSRF page on a hostname long enough that `https://<host>` exceeds 127 bytes would bypass the guard. Since HTTP auth is off by default, this guard is the sole defense for `/api/device/restart` and `/api/device/reset`.

Fixed: presence is now detected with `httpd_req_get_hdr_value_len`; an Origin or Host that overflows the buffer fails **closed** (403). Only a genuinely absent Origin is allowed.

### Confirmed sound
- Exact authority-vs-Host equality is correct for what browsers actually send; `Origin`/`Host` are forbidden headers JS cannot forge.
- Default-port omission and IPv6 bracket forms are symmetric between Origin and Host → no false 403s on legitimate same-origin paths.
- Arduino `String == char[]` compares contents (strcmp); substring bounds safe when `://` is absent or trailing.
- No now-unused includes; `sensesp_app.h` still needed by the other handlers.

### Residual (not blocking)
- Allowing requests with **no** Origin header is acceptable against browser-driven CSRF (modern browsers attach Origin to all cross-origin and same-origin non-GET requests). Non-browser clients reaching destructive endpoints when auth is off is a network-access exposure, not CSRF, and matches intent.

### Testing gap
- No unit test exercises `check_origin` (absent / matching / mismatched host / mismatched port / over-long → 403). `check_origin` is in an anonymous namespace, so this needs a test seam. Worth a follow-up.

PR #992, opened 2026-06-15
fix(web): bail out when httpd_start fails instead of serving a dead UI

### LINKED ISSUE #977: HTTPServer registers handlers after a failed httpd_start, leaving a dead web UI
## Problem

In `HTTPServer`'s constructor, when `httpd_start()` fails it logs the error but then registers URI handlers and announces the service over mDNS anyway, using the failed server handle (`src/sensesp/net/http_server.h`):

```cpp
esp_err_t error = httpd_start(&server_, &config_);
if (error != ESP_OK) {
  ESP_LOGE(__FILENAME__, "Error starting HTTP server: %s", esp_err_to_name(error));
} else {
  ESP_LOGI(__FILENAME__, "HTTP server started");
}

// register the request dispatcher for all methods and all URIs
httpd_register_uri_handler(server_, &uri);
// ... more registrations ...

// announce the server over mDNS
MDNS.addService("http", "tcp", 80);
```

The error is swallowed: control falls through to the registration block regardless. The device boots healthy, advertises HTTP over mDNS, but the web UI is dead — there's no signal to the user that the server never started.

## When it bites

The most likely trigger is a `max_open_sockets` value that exceeds the lwIP socket budget. `esp_http_server` rejects `config.max_open_sockets + 3 > CONFIG_LWIP_MAX_SOCKETS` in `httpd_start()` with `ESP_ERR_INVALID_ARG`. On a pure `framework = espidf` build `CONFIG_LWIP_MAX_SOCKETS` defaults to 10, so any override of `HTTP_SERVER_MAX_OPEN_SOCKETS` above 7 (without also raising `CONFIG_LWIP_MAX_SOCKETS` in `sdkconfig.defaults`) makes the server fail to start while the device otherwise looks fine. CI only covers arduino + pioarduino frameworks, so it won't catch this.

## Suggested fix

When `httpd_start()` fails, bail out of the registration block (and skip the mDNS announcement) instead of operating on the failed handle. Consider surfacing the failure more visibly than a single `ESP_LOGE`.

## Context

Split out from review discussion on #960. That PR no longer triggers this path (it keeps the default at 7 and fixes the original symptom via `lru_purge_enable`), so this is an independent pre-existing bug tracked separately.

mairas:
## Why

In `HTTPServer`'s constructor, when `httpd_start()` failed it logged the error but fell through and registered URI handlers and announced HTTP over mDNS anyway — operating on the failed/null server handle. The device booted healthy and advertised a web UI that was actually dead, with no signal to the user that the server never started.

The most likely trigger is a `max_open_sockets` value exceeding the lwIP socket budget (`config.max_open_sockets + 3 > CONFIG_LWIP_MAX_SOCKETS`), which `httpd_start()` rejects with `ESP_ERR_INVALID_ARG`. CI only covers arduino + pioarduino frameworks, so it doesn't catch the espidf default of `CONFIG_LWIP_MAX_SOCKETS = 10`.

## What

Return early from the startup lambda when `httpd_start()` fails, skipping handler registration and the mDNS announcement. The error is still logged via `ESP_LOGE`.

Split out from review discussion on #960; that PR no longer triggers this path, so this is an independent pre-existing bug.

Fixes #977

### DIFF
--- a/src/sensesp/net/http_server.h
+++ b/src/sensesp/net/http_server.h
@@ -106,9 +106,11 @@ class HTTPServer : public FileSystemSaveable {
       if (error != ESP_OK) {
         ESP_LOGE(__FILENAME__, "Error starting HTTP server: %s",
                  esp_err_to_name(error));
-      } else {
-        ESP_LOGI(__FILENAME__, "HTTP server started");
+        // Registering handlers or announcing mDNS against a failed handle would
+        // leave the device advertising a dead web UI with no further signal.
+        return;
       }
+      ESP_LOGI(__FILENAME__, "HTTP server started");
 
       // register the request dispatcher for all methods and all URIs
       httpd_uri_t uri = {


mairas:
## Review synthesis (independent correctness reviewer)

**No findings. Merge-ready.**

- `return;` correctly exits the `onDelay` lambda (the constructor has already returned by the time it runs); `server_` stays `nullptr`, the documented clean "not started" state — `httpd_start` leaves the handle untouched on error.
- Nothing downstream assumes registration succeeded: `stop()` → `httpd_stop(nullptr)` is a safe no-op and has no callers; the dispatcher is only reachable via a running server; `add_handler` touches only `handlers_`, independent of `server_`.
- No resource leak — `httpd_start` allocates nothing on failure; the skipped block allocated nothing needing cleanup.

### Residual (pre-existing, out of scope)
- No retry on `httpd_start` failure: a transient failure leaves the device without a web UI until reboot. This PR strictly improves the failure path (no longer advertises a dead UI over mDNS).

PR #993, opened 2026-06-15
fix(web): avoid leaking secrets via the network-readable log

### LINKED ISSUE #987: Web log can expose request URIs and payloads logged at INFO/DEBUG
Follow-up from the web log viewer review (#984).

`/api/log` serves captured log lines over HTTP, and the web server is unauthenticated by default. Two latent exposure paths beyond the SK token (already redacted in #984):

- `HTTPServer::dispatch_request` logs every request URI at INFO (`Handling request: %s`), including query strings. No secret-bearing GET param exists today, but any future endpoint that carries a token/secret as a query parameter would echo it into the web-readable log.
- Other DEBUG payload logs in `signalk_ws_client.cpp` (websocket/HTTP payloads) carry SK data, not auth tokens, but are now network-readable when DEBUG is on.

Actions to consider: strip the query portion before logging `req->uri`; document that `/api/log` mirrors serial output and must not be relied on as private unless web auth is enabled; avoid secrets in GET parameters framework-wide. Also add `charset=utf-8` to the `/api/log` content type.

mairas:
## Why

Since #984, captured log lines are served over HTTP at `/api/log`, and the web server is unauthenticated by default. That turns the log into network-readable output, so anything logged is potentially exposed.

## What

- **Strip the query string before logging the request URI.** `dispatch_request` logged `req->uri` (including query string) at INFO. No secret-bearing GET parameter exists today, but any future endpoint that carried a token as a query param would echo it into the web-readable log. Now only the path is logged.
- **`/api/log` returns `application/json; charset=utf-8`** so UTF-8 log lines render correctly.
- **Corrected the access-control comment** on the log handler: it claimed `/api/log` is "gated by the global dispatcher auth," but auth is off by default. The comment now states plainly that `/api/log` mirrors the serial log to anyone who can reach the server unless web auth is enabled.

## Deliberately out of scope

The DEBUG payload logs in `signalk_ws_client.cpp` (#987 also mentioned them) carry SK data, not auth tokens, and are DEBUG-level — off in normal builds. Removing them would cost debuggability for no token-exposure benefit, so they're left as-is.

Fixes #987

### DIFF
--- a/src/sensesp/net/http_server.cpp
+++ b/src/sensesp/net/http_server.cpp
@@ -2,6 +2,7 @@
 
 #include "sensesp/net/http_server.h"
 
+#include <cstring>
 #include <functional>
 #include <list>
 #include <lwip/sockets.h>
@@ -38,7 +39,13 @@ void urldecode2(char* dst, const char* src) {
 }
 
 esp_err_t HTTPServer::dispatch_request(httpd_req_t* req) {
-  ESP_LOGI(__FILENAME__, "Handling request: %s", req->uri);
+  // Log only the path, not the query string: the captured log is served over
+  // HTTP (and the server is unauthenticated by default), so a future endpoint
+  // carrying a secret as a query parameter must not echo it into the log.
+  const char* query = strchr(req->uri, '?');
+  int path_len = query ? static_cast<int>(query - req->uri)
+                       : static_cast<int>(strlen(req->uri));
+  ESP_LOGI(__FILENAME__, "Handling request: %.*s", path_len, req->uri);
 
   if (captive_portal_) {
     bool captured;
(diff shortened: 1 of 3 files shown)


mairas:
## Review synthesis (independent security reviewer)

The URI path-extraction is sound across all edge cases (no query, `?`-first, empty URI); `%.*s` with `int` precision is the correct contract and reads only the path prefix. The query-stripping affects only the log line — dispatch still uses the unmodified `req->uri`. The charset value is valid.

**One real residual leak found and fixed:** `config_handler.cpp` still logged the full URI (query string included) at three sites for `/api/config/*` GET and PUT — and those endpoints actually consume query parameters today, so the "future endpoint" framing understated a present gap. Fixed in commit 8b27fd33b: all three now log path-only via a shared `uri_path_len` helper, matching `dispatch_request`.

PR #996, opened 2026-06-15
refactor(system): deprecate no-op Startable shim

### LINKED ISSUE #902: Startable class is a no-op compatibility shim with no deprecation timeline
## Description

`system/startable.h` defines a `Startable` class that does nothing. All methods are empty, no startup ordering is performed, no callbacks are invoked. The class has no `[[deprecated]]` attribute and no documentation about whether it should be removed.

Downstream code inheriting from `Startable` gets false assurance that `start()` will be called.

## Recommendation

Either remove the class (breaking change for downstream inheritors) or mark it `[[deprecated("Startable is a no-op since vX.Y -- remove inheritance")]]`.

---

mairas:
## Motivation

`src/sensesp/system/startable.h` defines a `Startable` class whose methods are all no-ops: there is no startup ordering, no callback dispatch. Startup is now handled by the event loop, so the class serves no functional purpose and only exists for source compatibility with older code that still inherits from it.

## Approach

Mark the class `[[deprecated(...)]]` with a message explaining it is a no-op kept only for source compatibility and that inheritance should be removed. The class is intentionally **not** removed, preserving source compatibility. A short doc comment is updated to the same effect.

This is a non-breaking change: the project's build flags contain only `-Werror=reorder` (not blanket `-Werror`), and there are no `Startable` inheritors in `src/` or `examples/`, so the new attribute emits no hard errors.

Closes #902

### DIFF
--- a/src/sensesp/system/startable.h
+++ b/src/sensesp/system/startable.h
@@ -4,10 +4,15 @@
 namespace sensesp {
 
 /**
- * @brief Dummy Startable class for backward compatibility.
+ * @brief No-op Startable shim kept only for source compatibility.
  *
+ * All methods are no-ops: there is no startup ordering or callback dispatch.
+ * Startup is now handled by the event loop, so inheritance from this class
+ * should be removed.
  */
-class Startable {
+class [[deprecated(
+    "Startable is a no-op kept only for source compatibility; remove "
+    "inheritance, startup is handled by the event loop")]] Startable {
  public:
   Startable(int priority = 0) {}
   virtual void start() {}


mairas:
## Automated review

**Verdict: approve** — minimal, correct, and resolves the intent of #902.

The PR does exactly what #902 asked: marks the no-op `Startable` class `[[deprecated(...)]]` with a message that explains it is a no-op kept for source compatibility and that inheritance should be removed, without removing the class (so downstream inheritors still compile).

Findings:
- **Correctness (confirmed):** The `class [[deprecated(...)]] Startable` placement is valid C++; the warning is emitted only at points of *use* (inheritance/instantiation), not at definition. There are no `Startable` references, includes, or inheritors anywhere in `src/` or `examples/`, so no warning fires in-tree — the change is genuinely non-breaking. Downstream code that still inherits gets the deprecation warning at its own site, which is the desired signal.
- **Build flags (confirmed):** The repo uses only `-Werror=reorder`, not blanket `-Werror`, and `library.json` imposes no `-Werror` on consumers, so the new attribute cannot turn into a hard error. The PR description's reasoning holds.
- **Minimality / standards:** Single-file, 7/-2 diff. The reworked doc comment is short and WHY-focused, matching AGENTS.md conventions. No unrelated changes.
- **Nit (pre-existing, out of scope):** The include guard is inconsistent — `#ifndef SENSESP_SYSTEM_STARTABLE_H` vs the trailing `#endif // SENSESP_SRC_SENSESP_SYSTEM_STARTABLE_H_` comment. Harmless (the `#ifndef`/`#define` pair matches; only the comment differs) and untouched by this PR. Not a blocker.

No substantive issues. Ready to merge once CI is green.

PR #997, opened 2026-06-15
fix(builder): move kDefaultSystemInfoSensorPrefix into namespace

### LINKED ISSUE #911: kDefaultSystemInfoSensorPrefix in global namespace causes ODR violation
## Description

In `sensesp_app_builder.h:13`:

```cpp
const char* kDefaultSystemInfoSensorPrefix = "sensorDevice.";
```

This is defined outside the `sensesp` namespace, polluting the global namespace. It is also not `constexpr` or `inline`, so including this header in multiple translation units causes an ODR violation (multiple definitions).

## Fix

Move into the `sensesp` namespace and mark `inline` or `constexpr`.

---

mairas:
## Motivation

`kDefaultSystemInfoSensorPrefix` was declared at global scope in `src/sensesp_app_builder.h`, just above `namespace sensesp {`. The `constexpr` already gives it internal linkage, so the original ODR concern was resolved, but the declaration still polluted the global namespace.

## Approach

Move the declaration inside the `sensesp` namespace, immediately after the opening brace. The in-file references (default arguments in the builder methods) remain unqualified and resolve correctly since they live in the same namespace.

**BREAKING CHANGE**: Any downstream code referencing `::kDefaultSystemInfoSensorPrefix` at global scope must now use `sensesp::kDefaultSystemInfoSensorPrefix`. In practice the symbol had internal linkage and was only used inside this header, so external impact is effectively nil.

Closes #911

### DIFF
--- a/src/sensesp_app_builder.h
+++ b/src/sensesp_app_builder.h
@@ -10,9 +10,9 @@
 #include "sensesp_app.h"
 #include "sensesp_base_app_builder.h"
 
-constexpr const char* kDefaultSystemInfoSensorPrefix = "sensorDevice.";
-
 namespace sensesp {
+
+constexpr const char* kDefaultSystemInfoSensorPrefix = "sensorDevice.";
 
 /**
  * @brief A class for quickly configuring a SensESP application object before


mairas:
## Automated review

**Verdict: approve** — minimal, correct, and resolves the real remaining problem in #911.

- The fix is exactly right and minimal (2 lines, one file): the symbol moves inside `namespace sensesp`, immediately after the opening brace, and all six unqualified default-argument uses resolve since they live in the same namespace. No other references exist in `src/`.
- **Issue #911's premise was partly stale, and the PR correctly diagnoses this.** The issue claimed an ODR violation because the symbol was "not `constexpr` or `inline`", but the code on `main` already had `constexpr` — which gives a namespace/global-scope object internal linkage, so there was no ODR violation. The only real remaining defect was global-namespace pollution, which this PR fixes. Good that the PR body calls this out rather than parroting the issue.
- **BREAKING CHANGE flag is honest and correctly qualified.** Because the symbol has internal linkage, any TU referencing `::kDefaultSystemInfoSensorPrefix` at global scope had to include this header; after the move, including the header no longer exposes the global alias, so such code breaks at compile time. Real source break, vanishingly unlikely in practice — the "effectively nil" qualifier is accurate.
- Note (nit, not blocking): the breaking change is documented only in the PR body, not in the commit message footer (`BREAKING CHANGE:` / `fix(builder)!:`). This is fine here because SensESP versions manually via bumpversion + RELEASE_PROCESS.md, not via commit-message-driven release tooling — so nothing is missed. Flagging only so the human releaser remembers this warrants at most a note in CHANGELOG; it does not justify a major bump given the negligible real-world impact.

No blocking findings.

PR #998, opened 2026-06-15
fix(controllers): rename swich_consumer_ typo to switch_consumer_

### LINKED ISSUE #912: SmartSwitchController: typo in public API member 'swich_consumer_'
## Description

In `controllers/smart_switch_controller.h:108`:

```cpp
LambdaConsumer<bool> swich_consumer_{...};
```

Public member with a typo ("swich" instead of "switch"). Since `switch` is a C++ reserved keyword, the name should be something like `switch_consumer_` or `bool_consumer_`.

This is a breaking change for any downstream code accessing this member.

---

mairas:
## Motivation

The `SmartSwitchController` member `LambdaConsumer<bool> swich_consumer_` contains a typo ("swich"). This makes the public member name misleading and awkward to reference from downstream firmware.

## Approach

Rename the member to `switch_consumer_` and update every reference. A repo-wide grep for `swich` confirms only two occurrences, both now fixed:

- `src/sensesp/controllers/smart_switch_controller.h` (member declaration)
- `examples/smart_switch/remote_switch_example.cpp` (usage via `connect_to`)

No backward-compatibility alias is added, since aliasing a misspelled name is not worthwhile.

**BREAKING CHANGE**: Downstream code that references the public member `SmartSwitchController::swich_consumer_` must update to `switch_consumer_`.

Closes #912

### DIFF
--- a/examples/smart_switch/remote_switch_example.cpp
+++ b/examples/smart_switch/remote_switch_example.cpp
@@ -103,7 +103,7 @@ void setup() {
   // sent across the Signal K network when the controlling device
   // confirms it has made the change in state.
   auto* sk_listener = new SKValueListener<bool>(sk_path);
-  sk_listener->connect_to(controller->swich_consumer_);
+  sk_listener->connect_to(controller->switch_consumer_);
 }
 
 void loop() { event_loop()->tick(); }
--- a/src/sensesp/controllers/smart_switch_controller.h
+++ b/src/sensesp/controllers/smart_switch_controller.h
@@ -105,7 +105,7 @@ class SmartSwitchController : public ValueProducer<bool>, FileSystemSaveable {
     this->emit(this->is_on_);
   }};
 
-  LambdaConsumer<bool> swich_consumer_{[this](bool value) {
+  LambdaConsumer<bool> switch_consumer_{[this](bool value) {
     this->is_on_ = value;
     this->emit(is_on_);
   }};


mairas:
## Automated review

**Verdict: approve** — clean, minimal, correct fix that fully resolves #912.

- **Resolves intent:** Renames the misspelled public member `swich_consumer_` to `switch_consumer_` (one of the names the issue proposed) and updates the single usage site. Confirmed via repo-wide grep: exactly two `swich` occurrences on main, both fixed, zero remaining across src/examples/docs/tests.
- **Correctness:** No naming collision with sibling members (`click_consumer_`, `truthy_string_consumer_`). The issue's `switch`-keyword concern is moot since the chosen name is `switch_consumer_`, not the bare reserved word.
- **Minimality:** 2 files, +2/-2, no unrelated or over-engineered changes. Declining a backward-compat alias for a misspelled name is the right call.
- **Standards:** Snake_case trailing-underscore member naming preserved; no superfluous comments; conforms to AGENTS.md.
- **Breaking change:** Real and correctly flagged. The member is public and referenced from downstream firmware, so the rename is a genuine breaking change; the PR body labels it clearly and justifies the no-alias decision.

Nit (non-blocking): a public-API breaking rename ideally also lands in a changelog/release note for downstream consumers, but the repo doesn't appear to maintain a CHANGELOG file, so the PR-body flag is sufficient here.

No substantive findings. Good to merge.

PR #999, opened 2026-06-15
fix(ui): warn on StatusPageItem name collision

### LINKED ISSUE #918: StatusPageItem names can silently collide
## Description

In `ui/status_page_item.h:52`:

```cpp
status_page_items_[name] = this;
```

If two `StatusPageItem` instances share the same name, the second silently overwrites the first. The first item's value will never appear on the status page with no warning.

## Fix

Add a uniqueness check and log a warning on collision.

---

mairas:
## Motivation

In `src/sensesp/ui/status_page_item.h`, registering a `StatusPageItem` did `status_page_items_[name] = this;` unconditionally. When two items shared a name, the later one silently overwrote the earlier registration, so the first item's value never appeared on the status page and no diagnostic was emitted.

## Approach

Before inserting, check whether `name` is already a key in `status_page_items_`. On collision, log an `ESP_LOGW` warning naming the colliding `name` and keep the existing (first) registration, ignoring the later item. Only insert when the name is not already present.

Closes #918

### DIFF
--- a/src/sensesp/ui/status_page_item.h
+++ b/src/sensesp/ui/status_page_item.h
@@ -51,9 +51,13 @@ class StatusPageItem : public StatusPageItemBase, public ObservableValue<T> {
   StatusPageItem(String name, const T& value, String group, int order)
       : StatusPageItemBase(name, group, order), ObservableValue<T>(value) {
           if (status_page_items_.count(name) > 0) {
-            ESP_LOGW("StatusPageItem", "Duplicate status page item name: %s", name.c_str());
+            ESP_LOGW("StatusPageItem",
+                     "Duplicate status page item name: %s; keeping the first "
+                     "registration and ignoring this one",
+                     name.c_str());
+          } else {
+            status_page_items_[name] = this;
           }
-          status_page_items_[name] = this;
   }
 
   protected:


mairas:
## Automated review

Verdict: **approve** — the change correctly resolves #918 with a minimal, well-scoped diff.

- **Resolves issue intent (yes):** main already logged a warning but *still* overwrote unconditionally (`status_page_items_[name] = this;`), so the first item's value still disappeared — exactly the bug #918 describes. Moving the insert into the `else` branch makes the warning meaningful: the first registration is preserved and the later duplicate is ignored, matching the issue title and the "keep the first" intent.
- **Correctness:** the `count() > 0` / `else` insert is correct for `std::map`; the ignored item's `this` is never stored, so it introduces no dangling pointer.
- **Minimality / standards:** single-file, 6/-2, no unrelated changes; the added text lives in the log message (user-facing WHY), not as a code comment. Log tag `"StatusPageItem"` is unchanged and consistent with the literal-tag usages elsewhere.

Non-blocking notes (both pre-existing, out of scope for this PR — flagging only):
- Nothing removes an entry from the static `status_page_items_` map on destruction, so a destroyed `StatusPageItem` leaves a dangling pointer that `/api/info` would later dereference. Orthogonal to #918; worth a separate issue if not already tracked.
- The collision warning fires during construction, which for the built-in `sensesp_app.h` items can predate serial-debug init. Also pre-existing and unchanged.

No substantive issues with the change itself.

PR #1000, opened 2026-06-15
test: add unit tests for transforms, JSON round-trip, and Nullable

### LINKED ISSUE #920: Minimal unit test coverage: no tests for transforms, sensors, networking, or Signal K
## Description

The project has only 6 test files covering:
- CRGB color type
- ValueProducer lifecycle
- Event loop initialization
- LED blinker patterns
- TaskQueueProducer
- Basic offline functionality

**Not tested:** Any transform, any sensor, Signal K protocol, networking, configuration persistence, JSON serialization, or the entire `connect_to` pipeline.

Given the number of correctness bugs found in this review (variable shadowing, assignment typos, buffer resize bugs, uninitialized values), the lack of unit tests for transforms is a significant risk.

## Recommendation

Priority areas for test coverage:
1. **Transforms** -- especially `CurveInterpolator`, `MovingAverage`, `LambdaTransform`, `Hysteresis`, `Frequency`
2. **JSON serialization** -- `to_json`/`from_json` round-trip tests for all configurable objects
3. **Nullable** -- edge cases for each type specialization
4. **Multi-input transforms** -- once a unified pattern is designed (see #899)

---

mairas:
## What

Broad unit-test coverage for #920, following the existing Unity suite pattern (`test/system/test_*`):

- **Transforms** (`test_transforms/`): CurveInterpolator, MovingAverage, LambdaTransform, Hysteresis, Frequency.
- **JSON round-trip** (`test_json_roundtrip/`): Linear and MovingAverage `to_json`/`from_json` preserve behavior.
- **Nullable** (`test_nullable/`): validity detection, the invalid sentinel, implicit conversion, and JSON conversion across specializations.

## Two bugs this surfaced (handled separately)

Writing the tests exposed two real defects:

1. **`ThresholdTransform` config never reloads** — `from_json` required an `out_range` key `to_json` never writes. Fixed in #1002, where the threshold round-trip tests now live (alongside the fix they exercise).
2. **`Nullable` default ≠ invalid** — the class doc claimed a default-constructed `Nullable` is the invalid value, but the invalid sentinel is a separate per-type value, so a default holds a valid `0`. Per the maintainer's decision, the behavior is correct and the doc was stale: this PR corrects the doc comment and the tests assert the real contract via `invalid()`.

## Verification caveat

`ci/run-tests.sh` runs `pio test --without-uploading --without-testing` — CI only **compiles** these suites; the assertions are **not executed**. They were validated here by source-tracing each assertion and by compilation; to confirm they pass at runtime they must be run on a device (or a future host/native test harness). Worth a follow-up issue if executable CI tests are wanted.

Closes #920.

### DIFF
--- a/src/sensesp/types/nullable.h
+++ b/src/sensesp/types/nullable.h
@@ -8,8 +8,11 @@ namespace sensesp {
 /**
  * @brief Template class that supports a special invalid magic value for a type.
  *
- * The invalid value is always provided by the type's default constructor.
- *
+ * Each specialization defines an invalid sentinel (see nullable.cpp; e.g. float
+ * -> -1e9, uint8_t -> 0xff), matching the NMEA 2000 "missing data" values. A
+ * value is invalid only when it equals that sentinel. A default-constructed
+ * Nullable holds T{} (0), so it is valid for types whose sentinel differs from
+ * 0; use Nullable<T>::invalid() to obtain the invalid value explicitly.
  */
 template <typename T>
 class Nullable {
--- a/test/system/test_json_roundtrip/json_roundtrip_test.cpp
+++ b/test/system/test_json_roundtrip/json_roundtrip_test.cpp
@@ -11,7 +11,6 @@
 #include "sensesp/system/lambda_consumer.h"
 #include "sensesp/transforms/linear.h"
 #include "sensesp/transforms/moving_average.h"
-#include "sensesp/transforms/threshold.h"
 #include "unity.h"
 
 using namespace sensesp;
@@ -66,48 +65,10 @@ void test_moving_average_round_trip() {
 }
 
 // ---------------------------------------------------------------------------
-// FloatThreshold
-// ---------------------------------------------------------------------------
-
-void test_float_threshold_round_trip_preserves_behavior() {
-  FloatThreshold original(10.0f, 20.0f, true);
-
-  JsonDocument doc;
-  JsonObject obj = doc.to<JsonObject>();
-  TEST_ASSERT_TRUE(original.to_json(obj));
-  TEST_ASSERT_FLOAT_WITHIN(0.0001f, 10.0f, obj["min"].as<float>());
-  TEST_ASSERT_FLOAT_WITHIN(0.0001f, 20.0f, obj["max"].as<float>());
-  TEST_ASSERT_TRUE(obj["in_range"].as<bool>());
-
-  FloatThreshold restored(0.0f, 0.0f, false);
-  JsonObject in = doc.as<JsonObject>();
-  TEST_ASSERT_TRUE(restored.from_json(in));
-
-  bool received = false;
-  LambdaConsumer<bool> consumer([&received](bool v) { received = v; });
-  restored.connect_to(&consumer);
-
-  restored.set(15.0f);  // in range -> in_range value (true)
-  TEST_ASSERT_TRUE(received);
-
-  restored.set(25.0f);  // out of range -> !in_range (false)
-  TEST_ASSERT_FALSE(received);
-}
-
-void test_threshold_from_json_rejects_missing_key() {
-  FloatThreshold threshold(0.0f, 0.0f, true);
-
-  JsonDocument doc;
-  doc["min"] = 1.0f;
-  doc["max"] = 2.0f;
-  // "in_range" and "out_range" intentionally missing
-  JsonObject in = doc.as<JsonObject>();
-  TEST_ASSERT_FALSE(threshold.from_json(in));
-}
-
-// ---------------------------------------------------------------------------
 // Test runner
 // ---------------------------------------------------------------------------
+// ThresholdTransform round-trip lives in test_threshold_roundtrip, alongside
+// the from_json fix it exercises.
 
 void setup() {
   delay(2000);
@@ -116,8 +77,6 @@ void setup() {
 
   RUN_TEST(test_linear_round_trip_preserves_behavior);
   RUN_TEST(test_moving_average_round_trip);
-  RUN_TEST(test_float_threshold_round_trip_preserves_behavior);
-  RUN_TEST(test_threshold_from_json_rejects_missing_key);
 
   UNITY_END();
 }
--- a/test/system/test_json_roundtrip/json_roundtrip_test.cpp
+++ b/test/system/test_json_roundtrip/json_roundtrip_test.cpp
@@ -1,0 +1,125 @@
+/**
+ * @file json_roundtrip_test.cpp
+ * @brief to_json/from_json round-trip tests for configurable objects (#920).
+ *
+ * Confirms that serializing a transform's configuration and restoring it into
+ * a fresh instance preserves observable behavior.
+ */
+
+#include <Arduino.h>
+
+#include "sensesp/system/lambda_consumer.h"
+#include "sensesp/transforms/linear.h"
+#include "sensesp/transforms/moving_average.h"
+#include "sensesp/transforms/threshold.h"
+#include "unity.h"
+
+using namespace sensesp;
+
+// ---------------------------------------------------------------------------
+// Linear (LambdaTransform-based, two float params)
+// ---------------------------------------------------------------------------
+
+void test_linear_round_trip_preserves_behavior() {
+  Linear original(2.0f, 5.0f);
+
+  JsonDocument doc;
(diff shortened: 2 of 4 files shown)


mairas:
## Automated review

**Verdict: changes requested** — the structure, scope, and most assertions are solid, but two test files encode expected behavior that contradicts the actual code and would fail when executed. The compile-only CI hides this, so they will be merged as silently-broken tests.

### Blockers

- **`test_nullable/nullable_test.cpp` — false premise on the invalid sentinel.** The tests assume a default-constructed `Nullable<T>` equals the invalid sentinel and is therefore invalid. It does not. `Nullable()` sets `value_{}` (0 / 0.0f / 0), but the sentinels in `nullable.cpp` are `Nullable<int> = 0x7fffffff`, `Nullable<float> = -1e9`, `Nullable<uint8_t> = 0xff`. So for these, a default-constructed value has `is_valid() == true` and `value() != invalid()`. The following assertions fail at runtime:
  - `test_nullable_int_default_is_invalid`: `TEST_ASSERT_FALSE(n.is_valid())` and `TEST_ASSERT_EQUAL(NullableInt::invalid(), n.value())` (0x7fffffff vs 0).
  - `test_nullable_float_invalid_sentinel`: `TEST_ASSERT_FALSE(n.is_valid())` and the `-1e9` value check (value is 0.0f).
  - `test_nullable_uint8_invalid_sentinel`: `TEST_ASSERT_FALSE(n.is_valid())` and `TEST_ASSERT_EQUAL_UINT8(0xff, n.value())`.
  - Only `test_nullable_bool_invalid_is_false` happens to pass, because bool's sentinel (`false`) coincides with the default. This is itself the tell: the header comment "The invalid value is always provided by the type's default constructor" is contradicted by the specializations in `nullable.cpp`. Either the tests are wrong, or they have surfaced a real product bug (default ctor not using the sentinel) — but as written they assert the broken-as-passing behavior.

- **`test_json_roundtrip/json_roundtrip_test.cpp::test_float_threshold_round_trip_preserves_behavior` — round-trip does not round-trip.** `ThresholdTransform::to_json` writes only `{min, max, in_range}`, but `from_json` requires `{min, max, in_range, out_range}` and returns `false` if `out_range` is absent (threshold.h:45-50). So `restored.from_json(in)` returns `false` and `TEST_ASSERT_TRUE(restored.from_json(in))` fails at runtime. (This is also a latent product asymmetry worth a separate issue.)

### Why CI did not catch it
CI runs `pio test --without-uploading --without-testing` — compile-only, as the PR body correctly notes. The body also claims "all assertions are meaningful" and validation "by local compilation plus manual review of expected values." Compilation cannot validate assertion *values*, and the manual review missed the two cases above. The caveat is stated, but it is precisely the gap that lets these ship broken.

### What is correct (verified by tracing the implementations)
- CurveInterpolator: exact points, linear interpolation (x=1.4 -> 0.4; x=2.5 -> 1.0), below/above extrapolation (gradient 0 cases), empty -> 0, single-sample, JSON round-trip — all correct against `curveinterpolator.cpp`.
- MovingAverage windowed mean: traced `(8+8+8+4)/4 = 7.0` and the multiplier scaling — correct.
- Hysteresis dead-zone hold and inclusive upper threshold (`upper <= input`) — correct.
- LambdaTransform 0/2-param eval, JSON serialize/restore, and missing-key rejection — correct; direct `param1_/param2_` access is valid (public members).
- Frequency first-call suppression and elapsed-time emission (band 20-500 Hz for ~100 ms / 10 counts) — correct.
- Pattern, directory layout, file headers, and `setup()/loop()` runner all match the existing suites. No unrelated changes; scope is appropriate.

### Minor
- Some overlap with the existing `test_transform_correctness` suite (MovingAverage, Hysteresis) — acceptable, the new cases differ.

Fix: correct the three Nullable assertions and the FloatThreshold round-trip to match actual code (or, if the underlying behavior is considered a bug, split that into its own issue and have the test assert today's real behavior). Re-state the validation method honestly — compile + targeted local execution, not compile + eyeballing.

mairas:
## Addressing the review

The two `changes_requested` blockers are fixed, plus a third the review missed:

- **Nullable tests** rewritten to assert the real contract (`invalid()` → `!is_valid()`; a set value → `is_valid()`; default float is a valid `0`). Avoids the platform-dependent `int` sentinel (`0x7fffffff` on Arduino 3.x vs `0` on host). The class doc comment is corrected to match.
- **`test_nullable_to_json_invalid_is_null`** (not in the original findings) was also broken — it used a default-constructed `NullableFloat` as "invalid," but that's a valid `0` and serializes a value, not null. Now uses `NullableFloat::invalid()`.
- **Threshold round-trip** moved out to #1002 with the `from_json` fix that makes it pass; this PR no longer references `ThresholdTransform`.

The compile-only-CI caveat is now stated plainly in the PR body rather than as an "all assertions validated" claim.

mairas:
## Automated re-review

**Verdict: changes_requested** — one assertion will fail at runtime.

I traced every assertion against source on `main` (`moving_average.cpp`, `linear.cpp`, `lambda_transform.h`, `curveinterpolator.cpp`, `hysteresis.h`, `frequency.{h,cpp}`, `nullable.{h,cpp}`). CI only compiles these suites, so a wrong assertion ships silently — hence the trace.

### Blocker

**`test_transforms/transforms_test.cpp` :: `test_moving_average_multiplier_scales_output`** — asserts `20.0` but runtime yields `10.0`.

```cpp
MovingAverage ma(2, 2.0);
ma.set(10.0f);
TEST_ASSERT_FLOAT_WITHIN(0.0001f, 20.0f, received);  // expects 20.0
```

`MovingAverage::set` does not apply `multiplier_` on the first (initialization) sample:

```cpp
if (!initialized_) {
  buf_.assign(sample_size_, input);
  output_ = input;        // <-- multiplier NOT applied here
  initialized_ = true;
}
```

`output_` is emitted directly via `notify()`/`get()`, so the consumer receives `10.0`, not `multiplier_ * 10.0`. The multiplier only enters on subsequent samples (the `else` branch). The test comment ("scaled by 2.0 -> 20.0") encodes the wrong mental model of the init path. Fix options: (a) push a second sample so the multiplier-bearing path runs, e.g. `set(10); set(10);` then expect the scaled value, or (b) assert `10.0` and rename the test to reflect that the first sample is unscaled. Note this may itself be a latent product bug (first sample bypasses the multiplier) — worth a separate issue if (b) is chosen.

### Verified correct (would pass at runtime)

- **Nullable** (all 10): sentinels confirmed against `nullable.cpp` (float `-1e9`, uint8_t `0xff`, int `0x7fffffff` on Arduino>=3). `invalid()` reads invalid; default-constructed `float`/`int`/`uint8_t` read valid-zero; `bool` sentinel is `false` so default/`false` reads invalid. No test assumes a default Nullable is invalid except `bool`. JSON null<->invalid, value round-trip, `ptr()`, copy-assign all hold.
- **json_roundtrip**: `Linear` keys are `multiplier`/`offset` (via `param_info_` in `linear.cpp`), behavior round-trip `3*2+5=11` correct; `MovingAverage` keys `multiplier`/`sample_size`, round-trip preserves both. **No ThresholdTransform reference** remains (only a comment pointing to #1002).
- **transforms (rest)**: CurveInterpolator exact/interp/extrap/empty/single/JSON traced against the documented `(0,0)(1,0)(2,1)(3,1)` ramp — all match. `windowed_mean` traced step-by-step → `7.0` then `8.0`, correct (multiplier 1.0, so unaffected by the bug above). LambdaTransform 0/2-param + JSON + missing-key rejection correct. Hysteresis dead-zone hold + inclusive upper threshold (`upper <= input`) correct. Frequency first-call suppression (`first_call_ = true`) and multiplier round-trip correct.

### Nit

- The compile-only caveat in the PR body is present and accurate (`pio test --without-uploading --without-testing` compiles but does not execute); no overclaim. The Nullable doc-comment rewrite in `nullable.h` matches the real per-type-sentinel contract.

Once the MovingAverage multiplier assertion is corrected, this is good to merge.

mairas:
Re-review blocker resolved: the `test_moving_average_multiplier_scales_output` assertion was incorrect (the multiplier isn't applied to the seeded first sample, and a constant input cancels it), so it's removed rather than locking in questionable behavior. The MovingAverage multiplier semantics are now tracked in #1003. Buffer-fill and windowed-mean coverage remain.

Issue #909, opened 2026-03-25
Global #define max(a,b) macro in led_blinker.h pollutes namespace


mairas:
## Description

In `system/led_blinker.h:12`:

```cpp
#define max(a, b) ((a) > (b) ? (a) : (b))
```

This redefines `max` globally, conflicting with `std::max`, Arduino's own `max` macro, and any other code. It does not appear to be used in this file.

## Fix

Remove the `#define`. Use `std::max` where needed.

---
*Found during code review (2026-03-25)*

mairas:
Fixed in v3.3.0 — the `#define max(a, b)` macro was removed from `led_blinker.h` in c641cd40a. The header no longer defines it.

PR #1001, opened 2026-06-15
chore: address dead code and minor code-review findings

### LINKED ISSUE #919: Various dead code and minor issues found during code review
## Description

Collection of minor issues found during code review that do not warrant individual issues:

### Dead code / unused members
- `system/async_response_handler.h:96,98` -- `result_message_` and `elapsed_millis_` declared but never used
- `system/saveable.cpp:15` -- `hash_path` computed in `load()` but never used (lookup done by `find_config_file()`)
- `sensors/digital_input.h:66` -- `triggered_` in `DigitalInputState` initialized but never read/written
- `sensors/analog_reader.h:15-16` -- `output_scale_` declared but never used
- `net/web/base_command_handler.cpp:67` -- `sz = routes.size()` computed but never used

### Incorrect/misleading messages
- `system/saveable.cpp:33` -- Log says "to Json" but error is from `from_json()` (wrong direction)
- `types/json.h:29` -- Closing comment says `ARDUINOJSON_NAMESPACE` but actual namespace is `ArduinoJson` (stale from older version)

### Type issues
- `types/position.h:43` -- `kInvalidDouble` narrowed to `float` in `ENUVector` (becomes `-inf` or `lowest<float>`)
- `types/position.cpp:35-37` -- `canConvertFromJson` for `ENUVector` checks `is<double>()` but fields are `float`
- `transforms/change_filter.cpp:5-10` -- Custom `absf` instead of `std::abs`/`fabsf`
- `transforms/analogvoltage.h:8` -- `#define MAX_ANALOG_OUTPUT` instead of `constexpr`
- `transforms/voltage_multiplier.h:35` -- Missing `override` keyword on `set()`

### Miscellaneous
- `system/local_debug.h:19-25` -- `*ln` macros produce double newlines with ESP_LOG
- `net/http_server.cpp:183-192` -- Content-Type validated by string length (16) not value
- `net/networking.cpp:149-150` -- `static` locals in lambda shared across instances
- `net/http_authenticator.cpp:20,23` -- Returns `0` instead of `false` in bool function
- `ui/ui_controls.h:49-53,74-78` -- Schema generation uses fragile string replacement with two different placeholder syntaxes
- `system/led_blinker.h:44` -- `static bool blipping` shared across all `BaseBlinker` instances
- `system/led_blinker.h:127-133` -- `LEDPattern` constructors have inconsistent member initialization
- `sensors/digital_pcnt_input.cpp:33-42` -- include guard comment mismatch in `minimal_button.h:82`

---

mairas:
## Motivation

Issue #919 collects ~20 minor code-review findings: dead members/locals, stale comments, a narrowing conversion, a missing `override`, a shared-state bug, and a few correctness/readability nits.

## Approach

Each item was verified against current `main` before acting. Clearly-correct, low-risk fixes were applied; items that were already fixed, no longer present, or that would change behavior in a non-obvious way were skipped (listed below).

### Fixed

- **`system/async_response_handler.h`** — removed unused `result_message_` and `elapsed_millis_` members (and the now-unused `<elapsedMillis.h>` include and `elapsed_millis_ = 0` assignment).
- **`sensors/digital_input.h`** — removed `triggered_` from `DigitalInputState` (initialized but never read/written; `DigitalInputChange` keeps its own).
- **`sensors/analog_reader.h`** — removed unused `output_scale_` from `BaseAnalogReader`.
- **`net/web/base_command_handler.cpp`** — removed unused `int sz = routes.size();`.
- **`types/json.h`** — fixed closing comment `ARDUINOJSON_NAMESPACE` → `ArduinoJson` to match the actual namespace.
- **`types/position.h`** — `ENUVector` and `AttitudeVector` default constructors initialized `float` members from `kInvalidDouble` (a `double`), a narrowing conversion; now use the `float` constant `kPositionInvalidAltitude`. (Both structs share the identical defect, so both were fixed for consistency.)
- **`types/position.cpp`** — `canConvertFromJson(ENUVector)` checked `is<double>()` but the fields are `float`; changed to `is<float>()` to match the fields (and the existing `AttitudeVector` check).
- **`transforms/analogvoltage.h`** — moved `constexpr int kMaxAnalogOutput` inside `namespace sensesp` (it was at global scope; the only use is inside the namespace).
- **`transforms/voltage_multiplier.h`** — added `override` to `set()`.
- **`net/http_authenticator.cpp`** — `authenticate_request()` returns `bool`; changed two `return 0;` to `return false;`.
- **`net/http_server.cpp`** — `get_content_type()` gated on the Content-Type header *length* being exactly 16; now reads the value and returns it (the sole caller already compares the value strictly, so call-site behavior is unchanged while the magic-length check is removed).
- **`system/minimal_button.h`** — fixed stale include-guard closing comment (`..._BASE_BUTTON_H_` → `..._MINIMAL_BUTTON_H_`).
- **`system/led_blinker.h`** — `LEDPattern` initializer-list constructor redundantly re-initialized members that already have in-class initializers; removed for consistency with the vector constructor. Also changed `BaseBlinker::blip()`'s `static bool blipping` to a per-instance member `blipping_` — as a function-local `static` it was shared across all blinker instances, so one LED's in-progress blip would block another's.

### Skipped

- **`system/saveable.cpp` (hash_path in `load()`, "to Json" log message)** — already fixed in current code: `load()` no longer computes `hash_path`, and the log already reads "from Json".
- **`transforms/change_filter.cpp` (custom `absf`)** — already fixed; current code uses `fabsf`.
- **`system/led_blinker.h` `#define max` macro (#909)** — already removed; not touched per instructions.
- **`digital_pcnt_input.cpp` include-guard comment** — not a mismatch: the trailing `#endif  // ESP_ARDUINO_VERSION_MAJOR < 3` correctly labels its `#if`, and the header guard is consistent.
- **`system/local_debug.h` `*ln` macros (double newline with ESP_LOG)** — the `*ln`/`debug*` macros are unused inside SensESP and exist purely as a RemoteDebug/SerialDebug compatibility shim. Dropping `"\n"` is a behavior change to an external-facing compat API for no internal benefit, so left as-is.
- **`net/networking.cpp` static locals in lambda** — the referenced file no longer exists; the equivalent code now lives in `net/wifi_provisioner.cpp` inside `WiFiProvisioner`, which is effectively a singleton (one WiFi provisioner per device), so the "shared across instances" concern does not apply.
- **`ui/ui_controls.h` string-replacement schema generation** — replacing the `{{title}}` template substitution with proper JSON escaping is a non-trivial, behavior-affecting change out of scope for this cleanup.

## Verification

A full PlatformIO build could not be run locally: every environment targets ESP32 and the Arduino framework package is not provisioned in this environment (the build fails in `arduino.py` with `FRAMEWORK_DIR` = None before any compilation, independent of these changes). Relying on CI for the compile check. All changes are localized header/source edits reviewed by hand.

Closes #919

### DIFF
--- a/src/sensesp/net/http_authenticator.cpp
+++ b/src/sensesp/net/http_authenticator.cpp
@@ -18,10 +18,10 @@ bool HTTPDigestAuthenticator::authenticate_request(httpd_req_t* req) {
     return true;
   } else if (result == 0) {
     request_authentication(req, false);
-    return 0;
+    return false;
   } else {
     request_authentication(req, true);
-    return 0;
+    return false;
   }
 }
 
--- a/src/sensesp/net/http_server.cpp
+++ b/src/sensesp/net/http_server.cpp
@@ -190,14 +190,13 @@ esp_err_t call_request_dispatcher(httpd_req_t* req) {
 }
 
 String get_content_type(httpd_req_t* req) {
-  if (httpd_req_get_hdr_value_len(req, "Content-Type") != 16) {
+  char content_type[32];
+  if (httpd_req_get_hdr_value_str(req, "Content-Type", content_type,
+                                  sizeof(content_type)) != ESP_OK) {
     ESP_LOGE(__FILENAME__, "Invalid content type");
     return "";
-  } else {
-    char content_type[32];
-    httpd_req_get_hdr_value_str(req, "Content-Type", content_type, 32);
-    return String(content_type);
   }
+  return String(content_type);
 }
 
 }  // namespace sensesp
--- a/src/sensesp/net/web/base_command_handler.cpp
+++ b/src/sensesp/net/web/base_command_handler.cpp
@@ -180,8 +180,6 @@ void add_routes_handlers(std::shared_ptr<HTTPServer>& server) {
   JsonDocument json_doc;
   JsonArray routes_json = json_doc.to<JsonArray>();
 
-  int sz = routes.size();
-
   for (auto it = routes.begin(); it != routes.end(); ++it) {
     routes_json.add(it->as_json());
   }
--- a/src/sensesp/sensors/analog_reader.h
+++ b/src/sensesp/sensors/analog_reader.h
@@ -11,9 +11,6 @@ namespace sensesp {
  * @brief Used by AnalogInput as a hardware abstraction layer
  **/
 class BaseAnalogReader {
- private:
-  int output_scale_;
-
  public:
   virtual float read() = 0;
 };
--- a/src/sensesp/sensors/digital_input.h
+++ b/src/sensesp/sensors/digital_input.h
@@ -54,8 +54,7 @@ class DigitalInputState : public DigitalInput, public Sensor<bool> {
                     String config_path = "")
       : DigitalInput{pin, pin_mode},
         Sensor<bool>(config_path),
-        read_delay_{read_delay},
-        triggered_{false} {
+        read_delay_{read_delay} {
     load();
 
     repeat_event_ = event_loop()->onRepeat(read_delay_,
@@ -70,7 +69,6 @@ class DigitalInputState : public DigitalInput, public Sensor<bool> {
 
  private:
   int read_delay_;
-  bool triggered_;
   reactesp::RepeatEvent* repeat_event_ = nullptr;
   virtual bool to_json(JsonObject& root) override;
   virtual bool from_json(const JsonObject& config) override;
--- a/src/sensesp/system/async_response_handler.h
+++ b/src/sensesp/system/async_response_handler.h
@@ -4,7 +4,6 @@
 #include "sensesp.h"
 
 #include <Arduino.h>
-#include <elapsedMillis.h>
 
 #include "sensesp_base_app.h"
 #include "valueproducer.h"
@@ -45,7 +44,6 @@ class AsyncResponseHandler : public ValueConsumer<bool>,
    */
   void activate() {
     ESP_LOGV("AsyncResponseHandler", "Activating response handler");
-    elapsed_millis_ = 0;
     status_ = AsyncResponseStatus::kPending;
     this->emit(status_);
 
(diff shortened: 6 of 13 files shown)


mairas:
## Automated review

**Verdict: approve** — minimal, correct, behavior-preserving cleanup that resolves #919's intent. Verified every Fixed item against the branch and every Skipped item's rationale.

Spot-checks confirmed:
- `digital_input.h`: removed `triggered_` belongs to `DigitalInputState` only; `DigitalInputChange::triggered_` (still used) is correctly untouched. `DigitalInputState` never reads it (verified `.cpp` `to_json`/`from_json`).
- `position.h`/`position.cpp`: the narrowing fix (`kInvalidDouble`→`kPositionInvalidAltitude`) changes the default-constructed sentinel, but nothing compares `east/north/roll/pitch/yaw` against the old value, and `up` already used the float constant — so it is safe and consistent. `is<double>()`→`is<float>()` now matches the `AttitudeVector` check and the `float` field types.
- `http_server.cpp`: removing the magic length==16 gate is behavior-preserving for the sole caller (`config_handler.cpp` compares strictly to "application/json"); over-length values still return "" via `ESP_ERR_HTTPD_RESULT_TRUNC`.
- `led_blinker.h`: `static bool blipping` → per-instance `blipping_` is a genuine shared-state fix; removed `LEDPattern` initializer-list members are covered by in-class initializers.
- Dead-member removals (`async_response_handler.h`, `analog_reader.h`, `base_command_handler.cpp`) leave no dangling references; the unrelated `system_info` `elapsed_millis_` is a separate member.
- Skips verified: `saveable.cpp` `load()` no longer computes `hash_path`; `change_filter.cpp` already uses `fabsf`; `networking.cpp` no longer exists (logic moved to the effectively-singleton `WiFiProvisioner`); the `local_debug.h` / `ui_controls.h` skips are sound.

Nit (non-blocking):
- `voltage_multiplier.h`: `virtual void set(...) override` keeps the redundant `virtual`, but this matches existing local style (`to_json`/`from_json`), so leaving it is fine.

No blockers or major findings. CI must complete the compile check the author could not run locally.

PR #1002, opened 2026-06-15
fix(transforms): accept threshold config without out_range

mairas:
## Bug

`ThresholdTransform::from_json` required an `out_range` key, but `to_json` only writes `{min, max, in_range}` — and the class doesn't even store `out_range` (out-of-range output is derived as `!in_range_`). As a result, **a saved threshold configuration could never be reloaded**: `from_json` always returned `false`, leaving the transform on its constructor defaults after a restart.

## Fix

Require only the keys `to_json` actually emits: `{min, max, in_range}`. Save→load now round-trips.

## Tests

New suite `test/system/test_threshold_roundtrip/`:
- round-trip of a `FloatThreshold` config preserves behavior (in-range → `in_range`, out-of-range → `!in_range`);
- `from_json` still rejects input missing a required key.

Note: as with the rest of the suite, CI only *compiles* tests (`pio test --without-testing`); these assertions must be executed on a device to confirm they pass at runtime.

Discovered while implementing the #920 test coverage.

### DIFF
--- a/src/sensesp/transforms/threshold.h
+++ b/src/sensesp/transforms/threshold.h
@@ -42,7 +42,7 @@ class ThresholdTransform : public Transform<C, bool> {
   }
 
   bool from_json(const JsonObject& root) override {
-    String expected[] = {"min", "max", "in_range", "out_range"};
+    String expected[] = {"min", "max", "in_range"};
     for (auto str : expected) {
 
       if (!root[str].is<JsonVariant>()) {
--- a/src/sensesp/transforms/threshold.h
+++ b/src/sensesp/transforms/threshold.h
@@ -14,9 +14,8 @@ namespace sensesp {
  *   @param max_value Maximum value of input for output to be the value of
  * in_range.
  *
- *   @param in_range Output value if input value is in range.
- *
- *   @param out_range Output value if input value is out of the range.
+ *   @param in_range Output value if input value is in range. The output is
+ * the negation of this value when the input is out of range.
  */
 template <typename C>
 class ThresholdTransform : public Transform<C, bool> {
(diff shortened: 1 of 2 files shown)


mairas:
## Automated review

**Verdict: Approve.** The bug is real, the fix is correct and minimal, and both tests are correct by source-tracing.

> Verification note: tests were **not executed** — CI compiles tests only (`pio test --without-testing`) and there is no host harness. The assertions below were checked by reading the source, not by running them.

### Bug confirmed (blocker, now fixed)
`to_json` writes only `{min, max, in_range}` and the class has no `out_range_` member (out-of-range output is derived as `!in_range_`). The old `from_json` required `out_range`, a key `to_json` never emits, so the `is<JsonVariant>()` check rejected every config the class itself wrote — saved threshold configs could never reload. Dropping `out_range` from the required list makes the required set exactly match what `to_json` emits.

### Regression safety
No regression risk. `grep -rn out_range src/ test/ examples/` finds only (a) the line being fixed and (b) a constructor doc comment. Nothing anywhere writes an `out_range` JSON key, and `from_json` never read it (only min/max/in_range are consumed). The requirement was dead — removing it cannot break any caller or any previously saved config. Genuinely incomplete configs (missing min/max/in_range) are still rejected.

### Tests traced against source (not executed)
- `round_trip_preserves_behavior`: to_json → from_json restores min/max/in_range; `set(15)` → `in_range_` (true), `set(25)` → `!in_range_` (false). `set` calls `notify()` each time and the consumer is connected before the asserted calls, so both assertions should hold.
- `rejects_missing_required_key`: input has min+max, omits in_range; the loop hits `in_range`, `is<JsonVariant>()` is false → returns false.

### Nits (non-blocking)
- The `@param out_range` doc comment on the constructor (line 19) now describes a parameter that is no longer a persisted/required field; consider clarifying that out-of-range output is just `!in_range`.
- This could have lived in the existing `test/system/test_transform_correctness/` suite, but a dedicated directory is consistent with the per-bug isolation pattern. Fine as-is.

Conventions (AGENTS.md / existing test layout) followed; no unrelated changes in the diff.

Issue #874, opened 2026-03-25
MovingAverage::from_json resizes buffer to old size, causing out-of-bounds access


mairas:
## Description

In `transforms/moving_average.cpp:54-58`:

```cpp
if (sample_size_ != n_new) {
    buf_.assign(sample_size_, 0);  // uses OLD sample_size_
    ptr_ = 0;
    initialized_ = false;
    sample_size_ = n_new;          // sets new size AFTER resize
}
```

`buf_.assign(sample_size_, 0)` uses the old `sample_size_` instead of `n_new`. If the new sample size is larger, subsequent `set()` calls access `buf_[ptr_]` with `ptr_ % sample_size_` (new, larger value), but the buffer still has the old (smaller) size. This causes out-of-bounds access.

## Fix

```cpp
buf_.assign(n_new, 0);
sample_size_ = n_new;
```

---
*Found during code review (2026-03-25)*

PR #1004, opened 2026-06-15
fix(transforms): apply MovingAverage multiplier to the output

### LINKED ISSUE #1003: MovingAverage multiplier has ill-defined effect (unscaled first sample, cancels for constant input)
## Problem

`MovingAverage::set` (src/sensesp/transforms/moving_average.cpp) applies the `multiplier_` inconsistently:

1. **First sample is unscaled.** The init branch seeds `output_ = input;` directly, with no multiplier.
2. **Constant input cancels the multiplier.** For subsequent samples the update is `output_ += multiplier_ * (input - oldest) / sample_size_`. For a steady input `x`, `input - oldest` tends to 0, so the multiplier has no effect on the steady-state output — it only scales the transient response, divided by `sample_size_`.

The net effect: `multiplier` does not behave like a gain on the averaged value, which is what its name and the constructor doc (`@param multiplier`) imply.

## Why it matters

This surfaced while adding test coverage (#920, PR #1000). A natural test — "multiplier scales the output" — does not hold, so the assertion was removed rather than lock in the current behavior. The transform's multiplier needs a defined contract before it can be tested.

## Decision needed

Clarify the intended semantics, then either:
- make `multiplier` a true gain on the averaged output (apply it on read, including the first sample), or
- document it as a transient-response weight and rename/clarify accordingly.

Add unit coverage once the contract is settled.

mairas:
## Bug

`MovingAverage`'s documented contract is `y = multiplier * (1/n) * Σx` ("Moving average will be multiplied by multiplier before it is output"). The implementation didn't honor it:

- the seeded first sample was emitted unscaled (`output_ = input`);
- subsequent updates applied the multiplier to the *incremental delta* (`±multiplier * value / n`), so for a steady input the multiplier cancelled entirely and only affected the transient.

`multiplier = 1` (the default) behaved correctly as a plain moving average, but any `multiplier ≠ 1` — e.g. the `scale` used in the fuel-level and milone-level examples — produced wrong output.

## Fix

Maintain a running `sum_` of the buffered samples and compute `output_ = multiplier_ * sum_ / sample_size_` on every sample, including the seeded first one. This matches the documented formula exactly.

## Tests

New `test/system/test_moving_average_multiplier/`:
- multiplier scales the seeded first sample;
- multiplier applies to a steady input stream (the case that previously cancelled);
- output tracks `multiplier × windowed mean` across several samples;
- `multiplier = 1` is still a plain moving average.

Compile-verified locally with `pio test --without-uploading --without-testing -e arduino_esp32` (passes). As with the rest of the suite, CI compiles tests but does not execute them; assertions were traced against the new logic and should be run on a device to confirm.

Closes #1003.

### DIFF
--- a/src/sensesp/transforms/moving_average.cpp
+++ b/src/sensesp/transforms/moving_average.cpp
@@ -17,21 +17,19 @@ MovingAverage::MovingAverage(int sample_size, float multiplier,
 }
 
 void MovingAverage::set(const float& input) {
-  // So the first value to be included in the average doesn't default to 0.0
   if (!initialized_) {
+    // Seed the whole buffer with the first value so the average doesn't start
+    // out diluted by zeros.
     buf_.assign(sample_size_, input);
-    output_ = input;
+    sum_ = static_cast<float>(sample_size_) * input;
     initialized_ = true;
   } else {
-    // Subtract 1/nth of the oldest value and add 1/nth of the newest value
-    output_ += -multiplier_ * buf_[ptr_] / sample_size_;
-    output_ += multiplier_ * input / sample_size_;
-
-    // Save the most recent input, then advance to the next storage location.
-    // When storage location n is reached, start over again at 0.
+    // Swap the oldest buffered value out of the running sum for the newest.
+    sum_ += input - buf_[ptr_];
     buf_[ptr_] = input;
     ptr_ = (ptr_ + 1) % sample_size_;
   }
+  output_ = multiplier_ * sum_ / sample_size_;
   notify();
 }
 
--- a/src/sensesp/transforms/moving_average.h
+++ b/src/sensesp/transforms/moving_average.h
@@ -44,6 +44,7 @@ class MovingAverage : public FloatTransform {
   int ptr_ = 0;
   int sample_size_;
   float multiplier_;
+  float sum_ = 0;
   bool initialized_;
 };
 
--- a/test/system/test_moving_average_multiplier/moving_average_multiplier_test.cpp
+++ b/test/system/test_moving_average_multiplier/moving_average_multiplier_test.cpp
@@ -1,0 +1,89 @@
+/**
+ * @file moving_average_multiplier_test.cpp
+ * @brief Regression tests for MovingAverage's multiplier (#1003).
+ *
+ * The documented contract is y = multiplier * (1/n) * sum(last n inputs):
+ * the multiplier scales the averaged output, including the seeded first
+ * sample and a steady input stream.
+ */
+
+#include <Arduino.h>
+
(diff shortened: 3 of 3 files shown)


mairas:
## Automated review

**Verdict: Approve.** The fix correctly implements the documented contract `y = multiplier * (1/n) * Σx`, and all four test assertions match a hand-trace of the new logic. (Source-tracing only — CI compiles tests but does not execute them; I did not run them.)

### Correctness (verified by source-tracing)
- **Seeded first sample**: init branch sets `sum_ = sample_size_ * input`, so `output_ = multiplier_ * sum_ / n = multiplier * input`. Previously emitted unscaled (`output_ = input`). Fixed.
- **Steady input**: `sum_ += input - buf_[ptr_]` leaves `sum_` unchanged when input is constant, so `output_` stays `multiplier * mean`. This is the case where the old incremental-delta form cancelled the multiplier. Fixed.
- **Changing window**: traced `MovingAverage(2, 2.0)` for set(10),set(20),set(30) → 20, 30, 50, and `MovingAverage(4, 1.0)` for set(4),set(8),set(8),set(8) → 4, 5, 6, 7. Both match. The steady-input test `MovingAverage(4, 3.0)`×6 → 15.0 also matches.
- All four assertions in `test/system/test_moving_average_multiplier/` match the traced values.

### State / numerics
- `sum_` has in-class initializer `= 0`, but the init branch overwrites it unconditionally on first `set()`, so there is no reliance on stale state. `from_json()` resets `initialized_ = false` on a sample-size change, forcing re-seed and a fresh `sum_` recompute on the next `set()` — no stale carry-over across reconfiguration.
- No integer-division pitfall: `sum_` is `float`, so `sum_ / sample_size_` is float division despite `sample_size_` being `int`.
- Floating-point drift: `sum_` accumulates via add/subtract deltas rather than recomputing from the buffer each call — identical drift characteristics to the previous running-`output_` approach, so this is not a regression.

### Compatibility
- `multiplier = 1` (default): new output `= sum_/n` (true mean) equals the old behavior in steady state. Default behavior unchanged. Existing `multiplier != 1` users (e.g. `fuel_level_sensor`, `milone_level_sensor` `scale`) now get correct scaling — this is the intended bug fix, not a silent behavior change for correct callers.

### Conventions / minimality
- Diff is minimal: +1 member, net -2 lines in `set()`, plus a focused new test directory matching the existing system-test harness pattern (`test_crgb`, `test_transform_correctness`). No unrelated changes.
- The pre-existing `test_transform_correctness` covers `from_json` resize (#874); the new test covers the multiplier contract with no overlap.

No blocking issues. One non-blocking note already acknowledged in the PR description: assertions are source-traced, not device-executed.

PR #1012, opened 2026-06-17
ci: speed up workflow (trigger, cache, example matrix)

### LINKED ISSUE #1008: ci: speed up CI workflow (trigger, cache, example matrix)
## Overview

The CI workflow (`.github/workflows/ci.yml`) runs longer than necessary and wastes runner minutes. Three independent problems compound: every PR branch builds the full matrix twice, the build cache key is effectively broken, and the example-build job fans 3 examples across the device/platform matrix (8 jobs) even though that fan-out duplicates portability coverage the test job already provides.

This plan fixes all three without reducing the real signal: the `test` job keeps compiling all 13 test suites across the 3 platform/device combinations (`arduino_esp32`, `pioarduino_esp32`, `pioarduino_esp32c3`).

Both jobs are **compile-only** today (`pio test --without-testing` and `pio ci`), so neither runs anything online. That reframes the example matrix: its only unique value over the test job is catching example/tutorial code rot, which needs one platform, not a fan-out.

## Problem Frame

CI feedback is slow. Investigation found:

1. **Double-trigger.** `on: [push, pull_request]` runs the entire matrix twice for any branch with an open PR — once for the push event, once for the pull_request event.
2. **Broken cache key.** `key: ${{ runner.os }}-pio` is static. `actions/cache` only *writes* a new entry on a key miss, so a never-changing key is written once and never updated — the moment a toolchain or `lib_deps` version changes, every job re-downloads the large ESP32 toolchains forever. The single key is also shared across all matrix cells (arduino vs pioarduino, esp32 vs esp32c3 — different toolchains), so cells thrash one entry.
3. **Redundant example fan-out.** The `build` job builds 3 examples × device × platform (8 cells after excludes). Cross-device/cross-platform portability is already exercised by the `test` job; the example fan-out mostly re-proves it.

## Requirements Trace

- R1. A PR push triggers the matrix once, not twice.
- R2. Superseded in-progress runs on the same ref are cancelled when a newer commit is pushed.
- R3. The dependency cache is reused across runs and updated when `platformio.ini` changes; different matrix cells use distinct cache entries.
- R4. Example compile coverage is retained for the curated set (`minimal_app`, `rpm_counter`, `ssl_connection`) on a single environment.
- R5. The `test` job's 3-platform coverage is unchanged.

## Scope Boundaries

- Not changing `docs.yml`.
- Not enabling on-device test execution in CI (still compile-only).
- Not adding object-file caching (ccache) or restructuring the `ci/` scripts beyond the env values passed to them.
- Not expanding example coverage beyond today's curated 3 (decision below).

### Relevant Code and Patterns

- `.github/workflows/ci.yml` — the `test` and `build` jobs, both `fail-fast: false`.
- `ci/run-tests.sh` — `pio test --without-uploading --without-testing -e ${CI_PLATFORM}_${CI_DEVICE}`; compiles all 13 suites under `test/system/` per cell.
- `ci/run-ci.sh` — `pio ci -e ${CI_PLATFORM}_${CI_DEVICE} ...`; builds the example named by `$PLATFORMIO_CI_SRC`.
- `platformio.ini` — defines envs `arduino_esp32`, `pioarduino_esp32`, `pioarduino_esp32c3` (among others). `pioarduino_esp32` is the project default build env.
- `examples/` — 19 `.cpp` files; only 3 are built in CI today.
- `test/system/` — 13 test suites.

## Key Technical Decisions

- **Restrict `push` to `main`, keep `pull_request` for everything.** Eliminates the double-run. Branch pushes without a PR no longer trigger CI on origin — standard and acceptable; CI runs once a PR opens. Direct pushes to `main` (post-merge) still run.
- **Add a `concurrency` group keyed on workflow + ref with `cancel-in-progress: true`.** Kills superseded runs on rapid re-push. Applied workflow-wide.
- **Per-cell cache key with a `platformio.ini` hash plus `restore-keys` fallback.** Distinct entry per (platform, device); updates when dependencies change; falls back to the most recent compatible entry on miss.
- **Collapse the example matrix to one env per example (curated 3 on `pioarduino_esp32`).** Drops 8 build cells to 3. `pioarduino_esp32` satisfies the ssl example's `SENSESP_SSL_SUPPORT` requirement, so the previous arduino/ssl exclude becomes moot.

### Resolved During Planning

- Example aggressiveness: keep the curated 3, one env each (user decision) — not all-19, not dropped.
- Representative env for examples: `pioarduino_esp32` (project default; satisfies ssl).

### Deferred to Implementation

- Exact `restore-keys` prefix shape — finalize against the chosen key string when editing the file.

## Risks & Dependencies

| Risk | Mitigation |
|------|------------|
| Dropping arduino-platform example builds loses arduino-specific example compile coverage | The `test` job still compiles the framework on `arduino_esp32`; example-specific arduino breakage is unlikely and low-impact |
| Branch pushes without a PR no longer run CI | Standard GitHub flow; coverage resumes when the PR opens. `main` pushes still run |
| `cancel-in-progress` on `main` could cancel a post-merge run if merges land rapidly | Acceptable; the latest `main` commit is always the one that matters. Revisit if it causes gaps |

## Implementation Units

- [ ] #1009 — Restrict push trigger to main + concurrency cancellation
- [ ] #1010 — Fix broken static cache key (per-cell key + restore-keys + hashFiles)
- [ ] #1011 — Collapse example build matrix to one env per example

### LINKED ISSUE #1009: ci: restrict push trigger to main + add concurrency cancellation
Part of #1008.

**Goal:** Stop building every PR branch twice and cancel superseded runs.

**Requirements:** R1, R2

**Dependencies:** None

**Files:**
- Modify: `.github/workflows/ci.yml`

**Approach:**
- Replace `on: [push, pull_request]` with a `push` trigger restricted to the `main` branch plus an unfiltered `pull_request` trigger. This removes the duplicate run for branches that have an open PR while keeping post-merge runs on `main`.
- Add a workflow-level `concurrency` block grouping by workflow name + ref, with `cancel-in-progress: true`, so a newer push to the same ref cancels the older in-progress run.

Directional shape (guidance, not final syntax):

```yaml
on:
  push:
    branches: [main]
  pull_request:

concurrency:
  group: <workflow>-<ref>
  cancel-in-progress: true
```

**Test expectation:** none — CI-config change, verified by observing run behavior on a PR (see Verification).

**Verification:**
- Opening a PR triggers a single matrix run, not two.
- Pushing a second commit to an open PR cancels the first run and starts a new one.
- A merge/push to `main` still triggers a run.

### LINKED ISSUE #1010: ci: fix broken static cache key (per-cell key + restore-keys + hashFiles)
Part of #1008.

**Goal:** Make the PlatformIO dependency cache actually reusable and per-cell, so runs stop re-downloading the ESP32 toolchains.

**Requirements:** R3

**Dependencies:** None (independent of Unit 1; can land together)

**Files:**
- Modify: `.github/workflows/ci.yml` (the `actions/cache` step in both the `test` and `build` jobs)

**Approach:**
- Replace the static `key: ${{ runner.os }}-pio` with a key that includes the matrix cell identity and a hash of `platformio.ini`, so each (platform, device) combination gets its own entry that refreshes when dependencies change.
- Add `restore-keys` with a cell-scoped prefix so a changed `platformio.ini` still restores the most recent compatible cache instead of cold-downloading everything.
- Apply the same pattern to both jobs. For the `build` job after Unit 3 collapses its matrix, scope the key to the fixed env (or `hashFiles` alone) — finalize the exact prefix when editing.

Directional shape (guidance, not final syntax):

```yaml
key: ${{ runner.os }}-pio-${{ matrix.platform }}-${{ matrix.device }}-${{ hashFiles('platformio.ini') }}
restore-keys: |
  ${{ runner.os }}-pio-${{ matrix.platform }}-${{ matrix.device }}-
```

**Test expectation:** none — CI-config change, verified by cache hit/save behavior in run logs (see Verification).

**Verification:**
- First run after the change saves a cache entry per matrix cell (distinct keys in the logs).
- A subsequent run with unchanged `platformio.ini` reports a cache hit and skips toolchain download.
- A run after editing `platformio.ini` reports a `restore-keys` partial-restore, then saves a fresh entry.

### LINKED ISSUE #1011: ci: collapse example build matrix to one env per example
Part of #1008.

**Goal:** Remove the redundant device/platform fan-out from the example-build job, keeping compile coverage for the curated 3 examples on a single environment.

**Requirements:** R4, R5

**Dependencies:** None functionally, but coordinate the cache-key edit (Unit 2) since the `build` job's matrix shape changes.

**Files:**
- Modify: `.github/workflows/ci.yml` (the `build` job)

**Approach:**
- Reduce the `build` job matrix to a single dimension over `example`: `examples/minimal_app.cpp`, `examples/rpm_counter.cpp`, `examples/ssl_connection.cpp`.
- Fix the environment to `pioarduino_esp32` for all three (pass `CI_PLATFORM=pioarduino`, `CI_DEVICE=esp32` to `ci/run-ci.sh`). This is the project default env and satisfies the ssl example's `SENSESP_SSL_SUPPORT` requirement.
- Drop the now-moot `exclude` entries (arduino+esp32c3, arduino+ssl) — they only existed to prune the multi-platform grid.
- Leave `ci/run-ci.sh` unchanged; it already reads `CI_PLATFORM`/`CI_DEVICE` and `PLATFORMIO_CI_SRC`.
- The `test` job is untouched — it keeps its 3-cell platform/device matrix (R5).

Directional shape (guidance, not final syntax):

```yaml
build:
  strategy:
    fail-fast: false
    matrix:
      example:
        - examples/minimal_app.cpp
        - examples/rpm_counter.cpp
        - examples/ssl_connection.cpp
  # env fixed: CI_PLATFORM=pioarduino, CI_DEVICE=esp32
```

**Test expectation:** none — CI-config change, verified by build-job behavior (see Verification).

**Verification:**
- The `build` job runs exactly 3 cells (one per example), all on `pioarduino_esp32`.
- All three examples compile successfully.
- The `test` job still runs its 3 platform/device cells unchanged.

mairas:
## Why

CI feedback was slow and wasted runner minutes. Three independent problems compounded:

1. **Double-trigger** — `on: [push, pull_request]` ran the full matrix twice for any branch with an open PR.
2. **Broken cache key** — `key: ${{ runner.os }}-pio` was static. `actions/cache` only writes a new entry on a key miss, so the entry was written once and never updated; any toolchain or `lib_deps` change then forced every job to re-download the large ESP32 toolchains. The single key was also shared across all matrix cells (arduino/pioarduino, esp32/esp32c3 — different toolchains), so cells thrashed one entry.
3. **Redundant example fan-out** — the `build` job built 3 examples × device × platform (8 cells). Both jobs are compile-only, and the `test` job already exercises cross-platform portability, so the fan-out mostly re-proved it.

## How

- **Trigger:** `push` restricted to `main`, `pull_request` unfiltered. PRs build once; post-merge `main` pushes still run. Added a workflow-level `concurrency` group with `cancel-in-progress` so a newer push cancels the older run.
- **Cache:** per-cell key including `${{ matrix.platform }}`, `${{ matrix.device }}`, and `hashFiles('platformio.ini')`, with a `restore-keys` fallback. Each cell gets its own entry that refreshes when dependencies change.
- **Example matrix:** collapsed to one env per curated example (`minimal_app`, `rpm_counter`, `ssl_connection`) on `pioarduino_esp32` — the project default env, which also satisfies the ssl example's `SENSESP_SSL_SUPPORT` requirement. Drops 8 build cells to 3.

The `test` job keeps its 3-platform/device matrix unchanged — that's the real portability signal now that unit tests exist (compiled, not run, in CI).

## Tradeoffs

- Dropping arduino-platform example builds loses arduino-specific *example* compile coverage. The `test` job still compiles the framework on `arduino_esp32`, so framework portability is unaffected; example-specific arduino breakage is unlikely and low-impact.
- Branch pushes without an open PR no longer trigger CI on origin — standard GitHub flow; coverage resumes when the PR opens.

Closes #1008, closes #1009, closes #1010, closes #1011

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,13 @@
 name: SensESP CI
 
-on: [push, pull_request]
+on:
+  push:
+    branches: [main]
+  pull_request:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
 
 jobs:
   test:
@@ -24,7 +31,9 @@ jobs:
           path: |
             ~/.cache/pip
             ~/.platformio/.cache
-          key: ${{ runner.os }}-pio
+          key: ${{ runner.os }}-pio-${{ matrix.platform }}-${{ matrix.device }}-${{ hashFiles('platformio.ini') }}
+          restore-keys: |
+            ${{ runner.os }}-pio-${{ matrix.platform }}-${{ matrix.device }}-
       - uses: actions/setup-python@v5
         with:
           python-version: "3.11"
@@ -46,18 +55,10 @@ jobs:
           - examples/minimal_app.cpp
           - examples/rpm_counter.cpp
           - examples/ssl_connection.cpp
+        platform:
+          - pioarduino
         device:
           - esp32
-          - esp32c3
-        platform:
-          - arduino
-          - pioarduino
-        exclude:
-          - platform: arduino
-            device: esp32c3
-          # SSL example requires pioarduino (SENSESP_SSL_SUPPORT)
-          - platform: arduino
-            example: examples/ssl_connection.cpp
     steps:
       - uses: actions/checkout@v4
       - uses: actions/cache@v4
@@ -65,7 +66,9 @@ jobs:
           path: |
             ~/.cache/pip
             ~/.platformio/.cache
-          key: ${{ runner.os }}-pio
+          key: ${{ runner.os }}-pio-build-${{ matrix.platform }}-${{ matrix.device }}-${{ hashFiles('platformio.ini') }}
+          restore-keys: |
+            ${{ runner.os }}-pio-build-${{ matrix.platform }}-${{ matrix.device }}-
       - uses: actions/setup-python@v5
         with:
           python-version: "3.11"


mairas:
## Code review — CI workflow optimization

**Scope:** `.github/workflows/ci.yml` (single file)
**Intent:** restrict push trigger to `main` + concurrency cancellation; fix the static cache key; collapse the example-build matrix from 8 cells to one env per curated example, keeping the `test` job's 3-platform matrix intact.
**Reviewers:** correctness, testing, maintainability, project-standards (4 always-on; no conditional personas — small config diff, no auth/API/reliability surface).

### P2 — Moderate (resolved in this PR)

| # | Issue | Reviewer(s) | Conf | Resolution |
|---|-------|-------------|------|------------|
| 1 | Build cache key hardcoded `pioarduino-esp32`, drifting from the `CI_DEVICE`/`CI_PLATFORM` env vars — the test job avoids this by deriving its key from `${{ matrix.* }}`. Future env changes wouldn't track the key, causing silent cache pollution. | maintainability | 0.80 | Fixed — build job now uses single-value `platform`/`device` matrix entries, so the key and env block share one source of truth, mirroring the test job. |

### P3 — Low / advisory (not actioned)

| # | Issue | Reviewer(s) | Conf | Note |
|---|-------|-------------|------|------|
| 2 | Cache key hashes only `platformio.ini`, not the library manifest (`library.json`); floating `^` version specs can resolve to newer upstream releases without invalidating the key. | correctness, project-standards | ~0.70 | By design — the cache is a package *download* cache and PlatformIO revalidates at build time. Optional hardening: add `library.json` to `hashFiles(...)`. |
| 3 | All three build cells share one cache key (no `matrix.example` component), so only the first cell writes the cache. | correctness | 0.85 | Intended and correct — every cell builds the same `pioarduino_esp32` env, so cache contents are identical. |
| 4 | The `actions/cache` step block is duplicated across both jobs. | maintainability | 0.62 | GitHub Actions has no native step reuse without a composite action; not worth it for two jobs. |

### Coverage notes (intended tradeoffs)

- Example `.cpp` files are no longer compile-checked on the **arduino** platform or **esp32c3** device — only on `pioarduino_esp32`. The `test` job still compiles the framework + 13 test suites on all 3 platform/device combos, so framework portability is retained. The narrow residual: an arduino- or esp32c3-specific compile break originating *in an example file* (the examples have no device/platform `#ifdef` branches) would slip through. Accepted per the plan.
- SSL coverage unaffected — `ssl_connection.cpp` needs `SENSESP_SSL_SUPPORT` (pioarduino), which the retained env provides; the old matrix already excluded arduino+ssl.
- 16 of 19 examples remain uncompiled by CI — pre-existing, unchanged by this PR.

Suppressed: 1 finding below the 0.60 confidence gate.

---

**Verdict: Ready to merge.** No P0/P1. The one P2 has been fixed; remaining items are advisory and the coverage reductions are the deliberate goal of the change. Worth a green CI run on this PR to confirm the 3-platform test job and 3-cell build job behave as expected before merging.

PR #1013, opened 2026-06-17
fix(signalk): connect WebSocket directly, detect token rejection on the upgrade

mairas:
## Problem

On memory-constrained targets (e.g. ESP32-C3) `SKWSClient` could never connect to a TLS Signal K server. Before opening the WebSocket, it validated the stored auth token with a separate HTTPS GET to `/signalk/v1/stream` (`test_token`). That token-probe TLS handshake and the WebSocket TLS handshake run back-to-back, and mbedTLS's two 16 KB record buffers fragment the heap enough that the second handshake fails to allocate (`MBEDTLS_ERR_SSL_ALLOC_FAILED`).

## Approach

Connect the WebSocket directly with the stored token and move auth-failure recovery onto the upgrade itself. A 401 handshake status clears the token so the next reconnect re-requests access — the same recovery `test_token` provided, now driven by the real upgrade instead of a separate probe. This removes one TLS handshake per connect attempt. Non-401 transport errors keep the token and retry. `server_detected_` / `token_test_success_` are set at connect time so the `on_disconnected()` bad-token heuristic stays inert; bad-token detection is now precise via the 401 status. The now-unused `test_token()` is removed.

## Verification

Built and run on an ESP32-C3 (HALSER wind interface). The device discovers the server via mDNS, completes the `wss://` handshake (TOFU fingerprint verified), subscribes, and streams deltas — no `MBEDTLS_ERR_SSL_ALLOC_FAILED`.

### DIFF
--- a/src/sensesp/signalk/signalk_ws_client.cpp
+++ b/src/sensesp/signalk/signalk_ws_client.cpp
@@ -135,7 +135,17 @@ static void websocket_event_handler(void* handler_args,
       }
       break;
     case WEBSOCKET_EVENT_ERROR:
-      ws_client->on_error();
+      // The HTTP status of the failed upgrade (e.g. 401 for a rejected token)
+      // lets us distinguish a bad token from a transient transport error. The
+      // handshake status field only exists in the newer esp_websocket_client
+      // component pulled for SSL builds; the version bundled with the EOL
+      // espressif32 Arduino platform lacks it, so fall back to 0 (not
+      // applicable), matching how on_error() treats a transient failure.
+#ifdef SENSESP_SSL_SUPPORT
+      ws_client->on_error(data->error_handle.esp_ws_handshake_status_code);
+#else
+      ws_client->on_error(0);
+#endif
       break;
   }
 }
@@ -219,9 +229,19 @@ void SKWSClient::on_disconnected() {
  * Called in the websocket task context.
  *
  */
-void SKWSClient::on_error() {
+void SKWSClient::on_error(int handshake_status) {
   this->set_connection_state(SKWSConnectionState::kSKWSDisconnected);
-  ESP_LOGW(__FILENAME__, "Websocket client error.");
+  if (handshake_status == 401) {
+    // The server rejected the token on the WebSocket upgrade (e.g. the server
+    // was reinstalled, or the device was moved to a different server). Clear it
+    // so the next reconnect requests fresh access. A non-401 error (transport,
+    // TLS, network) leaves the token intact and simply retries.
+    ESP_LOGW(__FILENAME__, "Token rejected (401), requesting new access");
+    auth_token_ = NULL_AUTH_TOKEN;
+    save();
+  } else {
+    ESP_LOGW(__FILENAME__, "Websocket client error.");
+  }
 }
 
 /**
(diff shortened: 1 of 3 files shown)


mairas:
## Code review — multi-persona

**Scope:** `src/sensesp/signalk/signalk_ws_client.cpp` + `.h` (+38/−99, base `e4deda6`)
**Intent:** Remove the redundant pre-WebSocket `test_token()` HTTPS probe (two back-to-back TLS handshakes fragment the heap → `MBEDTLS_ERR_SSL_ALLOC_FAILED` on ESP32-C3) and detect a rejected token via the `401` WebSocket-upgrade status in `on_error()`.
**Reviewers:** correctness, reliability, security, adversarial, maintainability, api-contract, testing, project-standards.
**Verified working:** on-device (ESP32-C3, SSL build) the `wss://` handshake completes and a `401` upgrade is delivered as `WEBSOCKET_EVENT_ERROR` with `esp_ws_handshake_status_code == 401`, clearing the token. The core mechanism is sound; the findings below are about the paths it stopped covering.

### P1 — High

| # | File:line | Issue | Reviewers | Conf | Route |
|---|-----------|-------|-----------|------|-------|
| 1 | signalk_ws_client.cpp:145,234,763 | **Token-rejection recovery now hinges entirely on receiving a `401` status, and the fallback that used to catch the rest is disabled.** `connect()` forces `token_test_success_ = true`, so the `on_disconnected()` bad-token heuristic can never fire. Recovery then breaks in two cases: (a) **non-SSL builds** always call `on_error(0)` (the bundled esp_websocket_client has no `error_handle`), so a revoked token over `ws://` is never cleared; (b) a server that rejects by **closing the connection** (status 0, not 401) is treated as transient. Either way the device reconnects forever with a dead token and never re-requests access. The removed `test_token()` handled both (it worked over plain HTTP and had the heuristic as backup). | reliability, adversarial, correctness | 0.85 | manual → human |

### P2 — Moderate

| # | File:line | Issue | Reviewers | Conf | Route |
|---|-----------|-------|-----------|------|-------|
| 2 | signalk_ws_client.cpp:232 | **No test coverage for the new `on_error(int)` token-clear logic.** The 401-clears / non-401-keeps decision is pure integer-in/state-out and host-testable, but nothing exercises it — a future refactor could flip the comparison and CI stays green. Suggest extracting `status → should_clear` into a pure function and unit-testing it (avoids `save()`/filesystem and the private-method problem). | testing, correctness, maintainability, security | 0.90 | manual → human |
| 3 | signalk_ws_client.cpp:213,763 | **Vestigial state / dead code.** `token_test_success_` is now write-only-`true`, making the `on_disconnected()` heuristic (lines 214–221) unreachable and `server_detected_` effectively dead. Tension with finding #1: either *restore* the heuristic as the non-401 fallback (#1's fix) **or** delete the flags+heuristic cleanly — but the current "keep it alive only to neutralize it" state is a maintenance trap. Resolve #1 first; the direction there decides this one. | maintainability | 0.90 | gated_auto → human |
| 4 | signalk_ws_client.cpp:234 | **A transient/intermediary `401` wipes a valid token.** A captive portal, authenticating reverse proxy, or a server mid-restart can emit a one-off `401`; the code immediately clears+persists the token, forcing a full re-authorization that needs human approval on the SK server. Consider requiring corroboration (TOFU-confirmed server, or `401` on consecutive attempts) before discarding. | adversarial | 0.66 | manual → human |

### P3 — Low

| # | File:line | Issue | Reviewers | Conf | Route |
|---|-----------|-------|-----------|------|-------|
| 5 | signalk_ws_client.cpp:234 | **Security: `401` over plaintext `ws://` forces credential reset.** When `ssl_enabled_` is false in an SSL build (mDNS `signalk-ws`), the handshake status arrives unauthenticated; an on-path attacker can inject a `401` to force repeated token resets (credential DoS). Suggest gating the token-clear on `ssl_enabled_`. | security | 0.70 | manual → human |
| 6 | signalk_ws_client.cpp:755 | **clang-format:** the new comment block exceeds the 80-col Google-style limit (line 755 = 81 chars); `./run format` would reflow it. | project-standards | 0.85 | safe_auto → review-fixer |
| 7 | signalk_ws_client.cpp:234 | **Magic number `401`** drives a destructive action; a named constant documents intent and is greppable. | maintainability | 0.62 | safe_auto → review-fixer |
| 8 | signalk_ws_client.cpp:234 | **Recovery latency +up to ~60s:** old `test_token()` called `send_access_request()` synchronously on 401; new path waits for the (already-grown) backoff. Optionally `reset_reconnect_interval()` in the 401 branch. | reliability, correctness | 0.70 | advisory → human |
| 9 | signalk_ws_client.cpp:763 | `server_detected_`/`token_test_success_` left `true` if `connect_ws()` fails synchronously (init/start error) — flags reset next cycle, so low impact, but they now mean "intend to connect," not "verified server." | correctness, adversarial | 0.66 | advisory → human |
| 10 | signalk_ws_client.h:94 | **API:** `on_error()` signature change and `test_token()` removal are technically source-breaking public-surface changes (both are `public`, though task-internal). No in-tree dangling refs. Note in CHANGELOG/release; consider moving task methods to `protected` at the next major. | api-contract | 0.81 | advisory → release |

### Coverage
- Suppressed: low-confidence findings below the 0.60 gate were dropped.
- **Hardware verification gap:** the SSL 401 path is confirmed on-device; the non-SSL `on_error(0)` path and the connection-close (status 0) rejection path are **not** verified and are exactly where finding #1 lives.
- CI compiles both `#ifdef SENSESP_SSL_SUPPORT` branches (arduino = non-SSL, pioarduino = SSL) but runs no assertions — test suites build, never execute.

### Verdict — Ready with fixes

The common SSL path works and is verified. Before merge: resolve **#1** (decide whether non-SSL / connection-close token rejection must recover — restore the heuristic as a fallback, or document the scope as intentionally SSL-only) which also settles **#3**; and apply the two `safe_auto` nits (**#6**, **#7**). #2, #4, #5 are worth addressing but can be follow-ups if scoped explicitly.

mairas:
## Review response (commit 4224d95)

Addressed the review. Summary by finding:

**#1 — non-SSL / connection-close recovery (P1): fixed.** Restored `test_token()` for **non-SSL builds only**. The heap-fragmentation rationale is TLS-specific (two 16 KB mbedTLS record buffers), so the plain-HTTP probe is free of it — keeping it there preserves the recovery non-SSL builds had, while SSL builds keep the direct-connect + 401 path. One correction to the finding's premise: the `on_disconnected()` heuristic was **already dead in the base** (`token_test_success_` was only ever set to `true`, never reset), so it was never the non-SSL backup — `test_token()` itself was. Connection-close (status 0) rejection on SSL remains unrecovered-until-reconfig, but I have not seen signalk-server reject that way; happy to revisit if it does.

**#2 — no test coverage (P2): fixed.** Extracted the decision into a pure free function `should_clear_token_on_status(ssl_enabled, status)` and added `test/system/test_sk_token_clear/` covering the SSL-401 clear, non-401 keep, and plaintext-401 keep cases. Compiles in both the arduino (non-SSL) and pioarduino (SSL) envs; I couldn't run it on-device this session (no board attached), but it's a pure truth table.

**#3 — vestigial state (P2): fixed.** Deleted `server_detected_`, `token_test_success_`, and the unreachable heuristic block. Resolved in favour of #1's restore.

**#4 — transient 401 wipes a valid token (P2): not changed.** Judged out of scope for the deployment: signalk-server returns 401 only for an actually-invalid token, not transiently for a valid one, and the path is reached after mDNS-discovering a known SK server. Corroboration (consecutive-401 counter) adds state for a case I don't think occurs here. Flagging that I deliberately left it.

**#5 — plaintext 401 credential DoS (P3): fixed.** `should_clear_token_on_status` now requires `ssl_enabled` — only an authenticated TLS channel can trigger a token clear, so an injected plaintext 401 can't.

**#6 — clang-format (P3): fixed.** New comments wrapped to ≤80 cols by hand. I did **not** run `./run format` over the whole file: my local clang-format (v22) reflows many pre-existing lines differently from whatever version produced the current tree, which would bury the change in unrelated churn. The diff is functional-only.

**#7 — magic `401` (P3): fixed.** Named `kHttpUnauthorized`, used in both `on_error()` and `test_token()`.

**#8 — recovery latency (P3): partial.** Added `reset_reconnect_interval()` to the 401 branch as suggested. Note it resets the backoff *baseline* (so the multi-step re-auth flow runs at 2 s cadence) rather than forcing an immediate attempt — `next_attempt_ms_` is already scheduled by the current cycle's `schedule_reconnect()`.

**#9 — flags left true on sync connect_ws failure (P3): moot.** The flags are gone.

**#10 — API surface change (advisory).** `on_error(int)` and the now-non-SSL-only `test_token()` are task-internal `public` methods with no in-tree callers. `CHANGELOG.md` hasn't been updated since 2020 (last entry #128), so I didn't add an entry there — noting the change here instead.

Both `#ifdef SENSESP_SSL_SUPPORT` branches compile (arduino_esp32 and pioarduino_esp32 test builds pass).

mairas:
**On-device update (#2):** ran the new `test_sk_token_clear` suite on an ESP32-C3 (`pioarduino_esp32c3`, SSL build) — all 3 assertions pass:

```
test_clears_on_ssl_401            [PASSED]
test_keeps_token_on_ssl_non_401   [PASSED]
test_keeps_token_on_plaintext_401 [PASSED]
3 test cases: 3 succeeded
```

PR #1014, opened 2026-06-17
fix(log): make the web log viewer work with ANSI log colors enabled

mairas:
## Problem

The web log viewer (`/api/log` and the `/log` page) breaks on any build with `CONFIG_LOG_COLORS` enabled — the ESP-IDF default. (The precompiled Arduino framework ships with colors off, so this only surfaces on espidf-framework builds.) Two issues:

1. **Invalid JSON.** `ESP_LOG` wraps each line in ANSI color sequences. `LogBuffer` captured the raw ESC byte (`0x1b`), which ArduinoJson does not escape, so `/api/log` returned a body with a raw control character. The browser receives HTTP 200 but `response.json()` throws on every poll, so the viewer shows a permanent "Reconnecting…".
2. **Truncation.** Lines were capped at 128 chars even though the capture buffer is already 256, clipping most log lines.

## Approach

1. Strip ANSI escape sequences in `LogBuffer::push_line()`, which stores the buffered/served copy. The console output is forwarded to the previous `vprintf` handler *before* `push_line()` runs, so serial log colors are unaffected.
2. Raise the default `max_line_length` from 128 to 256 to match the existing 256-byte capture buffer. The 2-second age window (`max_age_ms`) bounds retained volume, so the added worst-case RAM is small.

## Verification

On an ESP32-C3 espidf build with `CONFIG_LOG_COLORS=y`:
- `/api/log` returns valid JSON — 0 raw `0x1b` bytes (was 7).
- The serial console keeps its colors (ESC bytes still present in the UART/USB stream).
- Lines now show up to 255 chars; 17 of 23 captured lines exceeded the old 128 limit.

Lines longer than the 256-byte capture buffer (e.g. verbose Signal K delta dumps) still clip at 255 — raising that would need a larger capture buffer and is left out here.

### DIFF
--- a/src/sensesp/system/log_buffer.cpp
+++ b/src/sensesp/system/log_buffer.cpp
@@ -69,6 +69,39 @@ int LogBuffer::vprintf_trampoline(const char* format, va_list args) {
   return ret;
 }
 
+namespace {
+// Copy text with ANSI/VT100 escape sequences removed. ESP_LOG wraps each line
+// in color codes when CONFIG_LOG_COLORS is enabled; the raw ESC byte (0x1b) is
+// a control character that ArduinoJson does not escape, so it would otherwise
+// reach the /api/log response unescaped and make the JSON unparseable in the
+// browser. A CSI sequence is ESC '[' then parameter/intermediate bytes up to a
+// final byte in 0x40-0x7e; a lone ESC is simply dropped.
+std::string strip_ansi(const char* text, size_t length) {
+  std::string out;
+  out.reserve(length);
+  size_t i = 0;
+  while (i < length) {
+    if (text[i] == '\x1b') {
+      i++;  // drop ESC
+      if (i < length && text[i] == '[') {
+        i++;  // drop '['
+        while (i < length) {
+          unsigned char c = static_cast<unsigned char>(text[i]);
+          i++;
+          if (c >= 0x40 && c <= 0x7e) {
+            break;  // final byte, end of sequence
+          }
+        }
+      }
+      continue;
+    }
+    out.push_back(text[i]);
+    i++;
+  }
+  return out;
+}
+}  // namespace
+
 void LogBuffer::push_line(const char* text, size_t length, uint32_t now_ms) {
   // Strip trailing CR/LF so each record is one display line.
   while (length > 0 && (text[length - 1] == '\n' || text[length - 1] == '\r')) {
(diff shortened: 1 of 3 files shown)


mairas:
## Code review — multi-persona

**Scope:** `src/sensesp/system/log_buffer.cpp` + `log_buffer.h` (~44 lines, base `main`)
**Intent:** Make `/api/log` valid JSON when `CONFIG_LOG_COLORS` is on (strip ANSI from the buffered copy; console keeps colors) and stop over-truncating lines (cap 128 → 256).
**Reviewers:** correctness, testing, maintainability, project-standards (always-on; small non-auth diff under the adversarial threshold, so no conditionals).
**Verified on device:** ESP32-C3, `CONFIG_LOG_COLORS=y` — `/api/log` parses (0 raw `0x1b`), console keeps colors, lines now reach 255 chars.

### P1 — High

| # | File | Issue | Reviewers | Conf | Route |
|---|------|-------|-----------|------|-------|
| 1 | test/system/test_log_buffer/ | **`strip_ansi` / `push_line` behavior ships untested.** The PR's core logic is a hand-rolled CSI parser with several edge branches, and the existing `test_log_buffer` suite already exercises `push_line()`→`snapshot_since()` (see `test_trailing_newline_stripped`, `test_bare_newline_dropped`) — so locking this in is cheap. Add cases: color-wrapped line → plain text; escape-only line → dropped (empty-after-strip early return); lone trailing ESC removed; truncated CSI (`\x1b[`, `\x1b[0` at end) consumed to end; **literal `[` preserved** (JSON-content guard); and a byte-level invariant that a stored line contains no `0x1b`. | testing, maintainability | 0.90 | manual → human |

### P3 — Low

| # | File | Issue | Reviewers | Conf | Route |
|---|------|-------|-----------|------|-------|
| 2 | log_buffer.cpp:57 / log_buffer.h:62 | **Coupled 256 caps lack a cross-reference.** The trampoline `buf[256]` bounds capture, and `max_line_length` is now 256 to match — but neither comments the other. Add a one-line note at each (or share a named constant) so a future bump of one doesn't silently half-apply. | maintainability | 0.74 | safe_auto → review-fixer |

### Coverage
- Correctness found no defects — the parser is sound on every boundary case it traced. Project-standards found no style/format issues.
- **CI compiles test suites but does not execute them** (`pio test --without-testing`), and `test_log_buffer` is platform-agnostic C++ — added tests guard compile-time and future host/native runs; a native execution env would make them run in CI.
- Residual (not blocking): `strip_ansi` removes only CSI/`0x1b` (ESP_LOG's only color form); other C0 control bytes or a UTF-8 char split by the 256 truncation are theoretical and pre-existing, not introduced here.

### Verdict — Ready with fixes

No correctness or standards blockers, and the behavior is verified on-device. Before merge: add the `strip_ansi` unit cases (#1) to lock in the verified behavior, and the cross-reference comment (#2). Neither changes runtime behavior.

mairas:
Addressed the review findings in 55d81ee:

- **P1 (strip_ansi untested):** added 7 cases to `test/system/test_log_buffer/` via the existing `push_line()`→`snapshot_since()` seam — color sequence → plain text, escape-only line dropped, lone trailing ESC, truncated CSI consumed to end, lone ESC keeps the following byte, literal `[` preserved, and the no-raw-`0x1b` invariant. Expected outputs were validated against the `strip_ansi` logic on the host; the suite compiles under `pioarduino_esp32`.
- **P3 (coupled 256 caps):** added cross-reference comments at the trampoline `buf[256]` and the `max_line_length` default.

No runtime behavior change.

Issue #893, opened 2026-03-25
ConfigItem(T*) raw pointer overload creates double-free hazard


mairas:
## Description

In `ui/config_item.h:299-303`:

```cpp
template <typename T>
std::shared_ptr<ConfigItemT<T>> ConfigItem(T* config_object) {
  auto config_object_sptr = std::shared_ptr<T>(config_object);
  return ConfigItem(config_object_sptr);
}
```

This wraps a raw pointer in a `shared_ptr`, taking ownership. The code comment acknowledges this is unsafe. If the pointer was not allocated with `new`, or if another owning construct already manages the object, this causes undefined behavior (double-free).

## Recommendation

Either remove this overload (breaking change) or use a non-deleting shared_ptr if the intent is to observe without owning:
```cpp
auto config_object_sptr = std::shared_ptr<T>(config_object, [](T*){});
```

---
*Found during code review (2026-03-25)*

PR #11, opened 2019-05-29
Add example: fuel level sensor

fabdrol:
See` README.md` in `examples/fuel_level_sensor`

PR #1017, opened 2026-06-17
fix(core): clean up static registries on destruction (#894)

### LINKED ISSUE #894: Static registries hold raw pointers with no cleanup on destruction
## Description

Several static containers accumulate object registrations during construction but never remove entries on destruction:

| Registry | Type | File |
|----------|------|------|
| `StatusPageItemBase::status_page_items_` | raw pointer | `ui/status_page_item.h:35` |
| `SKEmitter::sources_` | raw pointer | `signalk/signalk_emitter.cpp:8` |
| `TransformBase::transforms_` | raw pointer | `transforms/transform.cpp:17` |
| `ConfigItemBase::config_items_` | shared_ptr | `ui/config_item.h:172` |
| `UIButton::ui_buttons_` | shared_ptr | `ui/ui_button.h:43` |

The raw pointer maps are particularly dangerous -- if any registered object is destroyed, the registry holds a dangling pointer. The shared_ptr maps prevent destruction but can keep objects alive longer than intended.

None of these registries are cleared during `SensESPBaseApp::destroy()`.

## Recommendation

1. Add destructors that unregister from static containers
2. Clear all registries in `destroy()` for clean restart support
3. Consider using `std::weak_ptr` for registries that should not extend lifetime

---

mairas:
## Motivation

Several static registries accumulated object registrations during construction but never removed them. For the raw-pointer registries (`SKEmitter::sources_`, `TransformBase::transforms_`, `StatusPageItemBase::status_page_items_`) a destroyed object left a dangling pointer, and consumers iterate these globally (`signalk_delta_queue`, `base_command_handler`) — a use-after-free. The shared_ptr registries (`ConfigItemBase::config_items_`, `UIButton::ui_buttons_`) instead kept objects alive indefinitely. None were reset on app restart.

## Approach

- **Raw-pointer registries:** unregistering virtual destructors so they can never hold a dangling pointer. `StatusPageItemBase` only erases the entry it actually owns, preserving keep-first-duplicate behavior.
- **shared_ptr registries:** no self-erasing destructor — the registry *owns* the object, so a self-erase in the destructor would be re-entrant and pointless. A `clear_registry()` covers them.
- `ConfigItemBase` gets a virtual destructor (polymorphic-base hygiene).
- `SensESPBaseApp::clear_registries()` clears all five; called from both `destroy()` overrides for clean restart / test isolation.
- Deprecate the unused `TransformBase::get_transforms()` (superseded by `SKEmitter::get_sources()`).

## Behavioral / compatibility notes

- `destroy()` now empties all registries (intended; previously left populated).
- Objects unregister on destruction (only "breaks" code relying on the dangling-pointer bug).
- `get_transforms()` is `[[deprecated]]` — warning only; a hard error solely under `-Werror` (not used by this project).
- Declaring the destructors suppresses implicit move ops on those classes — harmless (registered by `this`, never moved).

## Merge order

Land after #893 (the ConfigItem ownership contract that registry clearing relies on).

## Verification

Full test suite passes on HALMET hardware (pioarduino_esp32): **15 suites / 121 cases**, including the new `test_registries` (unregister-on-destruction for all three raw registries + `clear_registries()` emptying every registry and surviving objects safely no-op'ing afterward). No regressions in the existing suites.

Closes #894

### DIFF
--- a/src/sensesp/signalk/signalk_emitter.cpp
+++ b/src/sensesp/signalk/signalk_emitter.cpp
@@ -1,5 +1,7 @@
 #include "signalk_emitter.h"
 
+#include <algorithm>
+
 namespace sensesp {
 
 std::vector<SKEmitter*> SKEmitter::sources_;
@@ -7,6 +9,13 @@ std::vector<SKEmitter*> SKEmitter::sources_;
 SKEmitter::SKEmitter(const String& sk_path) : sk_path_{sk_path} {
   sources_.push_back(this);
 }
+
+SKEmitter::~SKEmitter() {
+  sources_.erase(std::remove(sources_.begin(), sources_.end(), this),
+                 sources_.end());
+}
+
+void SKEmitter::clear_registry() { sources_.clear(); }
 
 void SKEmitter::add_metadata(JsonArray& meta) {
   SKMetadata* my_meta = this->get_metadata();
--- a/src/sensesp/signalk/signalk_emitter.h
+++ b/src/sensesp/signalk/signalk_emitter.h
@@ -27,6 +27,12 @@ class SKEmitter : virtual public Observable {
   SKEmitter(const String& sk_path);
 
   /**
+   * Unregisters this emitter from the static source registry so the
+   * registry never holds a dangling pointer.
+   */
+  virtual ~SKEmitter();
+
+  /**
    * Returns the data to be reported to the server as
    * a Signal K json object.
    */
@@ -64,6 +70,12 @@ class SKEmitter : virtual public Observable {
   void set_sk_path(const String& path) { sk_path_ = path; }
 
   static const std::vector<SKEmitter*>& get_sources() { return sources_; }
+
+  /**
+   * Empties the source registry. Intended for clean app restart and test
+   * isolation; not for normal runtime use.
+   */
+  static void clear_registry();
 
  protected:
   String sk_path_{};
--- a/src/sensesp/signalk/signalk_listener.cpp
+++ b/src/sensesp/signalk/signalk_listener.cpp
@@ -1,5 +1,7 @@
 #include "signalk_listener.h"
 
+#include <algorithm>
+
 #include "sensesp/system/saveable.h"
 
 namespace sensesp {
@@ -17,6 +19,19 @@ SKListener::SKListener(const String &sk_path, int listen_delay,
       listen_delay{listen_delay} {
   listeners_.push_back(this);
   this->load();
+}
+
+SKListener::~SKListener() {
+  take_semaphore();
+  listeners_.erase(std::remove(listeners_.begin(), listeners_.end(), this),
+                   listeners_.end());
+  release_semaphore();
+}
+
+void SKListener::clear_registry() {
+  take_semaphore();
+  listeners_.clear();
+  release_semaphore();
 }
 
 bool SKListener::take_semaphore(uint64_t timeout_ms) {
--- a/src/sensesp/signalk/signalk_listener.h
+++ b/src/sensesp/signalk/signalk_listener.h
@@ -36,6 +36,13 @@ class SKListener : virtual public Observable, public FileSystemSaveable {
              const String& config_path = "");
 
   /**
+   * Unregisters this listener from the static registry so the registry never
+   * holds a dangling pointer. Takes the listener semaphore because the WS
+   * client task iterates the registry under the same lock.
+   */
+  virtual ~SKListener();
+
+  /**
    * Returns the current Signal K path. An empty string
    * is returned if this particular source is not configured
    * or intended to return actual data.
@@ -65,6 +72,12 @@ class SKListener : virtual public Observable, public FileSystemSaveable {
   virtual void parse_meta(const std::shared_ptr<const JsonDocument>& meta_doc) {}
 
   static const std::vector<SKListener*>& get_listeners() { return listeners_; }
+
+  /**
+   * Empties the listener registry. Intended for clean app restart and test
+   * isolation; not for normal runtime use. Takes the listener semaphore.
+   */
+  static void clear_registry();
 
   static bool take_semaphore(uint64_t timeout_ms = 0);
   static void release_semaphore();
--- a/src/sensesp/signalk/signalk_put_request_listener.cpp
+++ b/src/sensesp/signalk/signalk_put_request_listener.cpp
@@ -1,11 +1,30 @@
 #include "signalk_put_request_listener.h"
 
+#include <algorithm>
+
+#include "signalk_listener.h"
+
 namespace sensesp {
 
 std::vector<SKPutListener*> SKPutListener::listeners_;
 
 SKPutListener::SKPutListener(const String& sk_path) : sk_path{sk_path} {
   listeners_.push_back(this);
+}
+
+SKPutListener::~SKPutListener() {
+  // The WS client iterates this registry while holding the SKListener
+  // semaphore, so mutate it under the same lock.
+  SKListener::take_semaphore();
+  listeners_.erase(std::remove(listeners_.begin(), listeners_.end(), this),
+                   listeners_.end());
+  SKListener::release_semaphore();
+}
+
+void SKPutListener::clear_registry() {
+  SKListener::take_semaphore();
+  listeners_.clear();
+  SKListener::release_semaphore();
 }
 
 }  // namespace sensesp
--- a/src/sensesp/signalk/signalk_put_request_listener.h
+++ b/src/sensesp/signalk/signalk_put_request_listener.h
@@ -26,6 +26,13 @@ class SKPutListener : virtual public Observable {
    */
   SKPutListener(const String& sk_path);
 
+  /**
+   * Unregisters this listener from the static registry so the registry never
+   * holds a dangling pointer. Takes the SKListener semaphore because the WS
+   * client task iterates this registry under that same lock.
+   */
+  virtual ~SKPutListener();
+
   String& get_sk_path() { return sk_path; }
 
   virtual void parse_value(const JsonObject& put) = 0;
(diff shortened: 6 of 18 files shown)


mairas:
## Code review — registry cleanup (#894)

**Scope:** `fix/894-registry-cleanup` vs `origin/main` (2 commits, 14 files, ~244 LOC). **Mode:** interactive (read-only, no fixes applied).
**Intent:** Stop static registries from holding dangling pointers / leaking, and reset them on app restart.
**Reviewers:** correctness, testing, maintainability, project-standards (always-on) + adversarial (memory-safety/lifetime mutation) + api-contract (public API + deprecation). project-standards came back clean.

Verdict: **Ready with fixes.** No P0; the implemented code is correct for what it covers and the hardware suite passes. But the fix is **incomplete relative to its own goal** (two more raw-pointer registries with the identical bug), and the subtlest new branch is untested. Recommend addressing the P1s before merge.

### P1 — High

| # | File | Issue | Reviewer(s) | Conf | Route |
|---|------|-------|-------------|------|-------|
| 1 | `signalk_listener.cpp:18`, `signalk_put_request_listener.cpp:8` | **`SKListener::listeners_` and `SKPutListener::listeners_` are two more static raw-pointer registries with the exact #894 bug** — self-register `this` in the ctor, no destructor, not cleared by `clear_registries()`, and iterated on the WS receive/PUT hot path (`signalk_ws_client.cpp:369,466,535`). The #894 issue table omitted them, so the PR fixes 3 of 5 raw registries and leaves these two dangling. | correctness (verified) | 0.92 | manual → decide: fold in or file follow-up |
| 2 | `status_page_item.cpp` (`~StatusPageItemBase`) | **Keep-first-duplicate / erase-if-`this` guard is untested.** The constructor keeps the first registration on a duplicate name; the destructor only erases when `it->second == this`. This is the trickiest new branch and the whole reason the guard exists, yet no test creates two same-named items. A naive `erase(name_)` would pass every current test while corrupting the registry. | testing, correctness, maintainability, adversarial, api-contract | 1.0 | add test |
| 3 | `test_registries/registries_test.cpp` (`test_clear_registries_empties_all`) | **The destruction-order interplay it sets up is never actually verified.** It registers an `SKOutput` owned by `config_items_` that's also in `sources_`/`transforms_` (the load-bearing ordering case), then asserts only post-clear emptiness — which can't detect a use-after-free or double-erase if the order were wrong. Drop the local owner before clearing and assert a `weak_ptr` expired, proving the owning-registry clear ran the destructor. | testing | 0.85 | strengthen test |

### P2 — Moderate

| # | File | Issue | Reviewer(s) | Conf | Route |
|---|------|-------|-------------|------|-------|
| 4 | `sensesp_app.h` / `sensesp_base_app.h` (`destroy()`) | **`clear_registries()` is duplicated in both `destroy()` overrides** — `SensESPApp::destroy()` fully replaces the base without chaining, so teardown logic is copy-pasted and can drift (a future change to one is silently missed in the other). Factor the shared body into a helper, or have the override call the base. | maintainability | 0.72 | refactor |
| 5 | `sensesp_app.h` / `sensesp_base_app.h` (`destroy()`) | **`clear_registries()` runs unconditionally even when teardown fails.** When `outside_users` is true (active references → "won't be properly destroyed"), the singleton survives but its registries are emptied; its member `StatusPageItem`s (and SK emitters) only register in their ctor, so they're now permanently deregistered — `/api/info` goes blank. Consider gating the clear on a successful teardown. | adversarial | 0.68 | gated — design decision |
| 6 | `sensesp_base_app.cpp` (`clear_registries`) | **`destroy()` doesn't quiesce the event loop / HTTP server before clearing.** Clearing `config_items_` can delete sole-owned objects still referenced by live `onRepeat` closures and the httpd handler (`/api/info` iterating `status_page_items_`). A real window on hardware, though `destroy()` is rarely called outside tests. | adversarial | 0.70 | gated — design decision |
| 7 | `test_registries/registries_test.cpp:58-61` | Pre-clear assertions use `GREATER_OR_EQUAL(1)` size floors instead of per-object membership, so cross-test residue could mask a failed registration. Use `count("ClearTestItem")==1` etc., matching the membership style already used elsewhere in the file. | testing | 0.75 | strengthen test |
| 8 | (release) | **ABI + behavioral changes need a changelog/release note:** added virtual destructors change vtable layout (source-compatible, ABI-incompatible — fine for PlatformIO source builds); `destroy()` now clears all registries (a public-method contract change documented only by an inline comment). | api-contract | 0.80 | release note |

### P3 — Low

| # | File | Issue | Reviewer(s) | Conf | Route |
|---|------|-------|-------------|------|-------|
| 9 | `transform.h` (`get_transforms()`) | Deprecation message says "use `SKEmitter::get_sources()` instead", but they return different types/sets (`set<TransformBase*>` vs `vector<SKEmitter*>`) — not a drop-in. Soften to "deprecated and unused internally; no direct replacement." | api-contract | 0.66 | safe doc fix |
| 10 | `config_item.h`, `ui_button.h` | The raw-vs-shared_ptr asymmetry (why these two get no self-erasing destructor) is explained only in `sensesp_base_app.cpp`. Add a one-line note at each class so a future maintainer doesn't "fix" the perceived inconsistency. | maintainability | 0.62 | safe doc fix |
| 11 | `status_page_item.cpp` (`~StatusPageItemBase`) | Duplicate-name behavioral edge: if the first registrant is destroyed while a live duplicate exists, the entry is removed and the duplicate is never promoted — the metric silently disappears. Consider making duplicate names a hard error at registration. | adversarial | 0.72 | gated |
| 12 | `test_registries` | Missing: idempotent double-`clear_registries()` test; focused `ConfigItemBase`/`UIButton` clear tests; destroying via a base-class pointer (multiple-inheritance vtable dispatch). | testing | 0.68 | add tests |

### Coverage

- **Suppressed:** none below threshold worth noting.
- **Static-destruction-order (residual):** the new `~StatusPageItemBase`/`~SKEmitter`/`~TransformBase` touch static containers in other TUs with no defined inter-TU teardown order. On ESP32 program exit is a reset (rare), but it's latent UB if an object outlives its registry's static map. (adversarial, 0.70)
- **`Saveable` has virtual methods but no virtual destructor** — latent; no base-pointer `delete` of `Saveable*` exists today. (correctness)
- **Move ops suppressed** on the four classes by the user-declared destructors — harmless (they self-register `this`; moving was never safe), worth a one-line `// non-movable` if you want it intentional.
- **Test gating:** all test envs are hardware ESP32 — no native env, so these tests don't gate CI unless a hardware-in-the-loop runner exists. The registry logic is platform-independent and could largely run native.
- **#893 interplay:** clearing `config_items_` now actively deletes objects registered via the raw `ConfigItem(T*)` overload when the registry was sole owner — reinforces the #893 ownership contract and the merge-order dependency (land #893 first).

### Recommended order
1. Decide #1 (listeners): fold `SKListener`/`SKPutListener` into this PR (mechanical — mirror `~SKEmitter` + add to `clear_registries()`), or file a tracked follow-up. Folding in is what actually closes #894's stated goal.
2. Add the duplicate-name test (#2) and strengthen the clear test with a `weak_ptr` destruction proof (#3).
3. De-dup `destroy()` (#4); decide the survive-teardown behavior (#5/#6).
4. Safe doc fixes (#9, #10) + release note (#8).

mairas:
## Review findings addressed

Pushed fixes for the review above. Summary of what changed and what was deliberately deferred.

### Fixed in code

| # | Finding | Resolution |
|---|---------|------------|
| 1 | **Two more raw-pointer registries with the same bug** (`SKListener`, `SKPutListener`) | Folded in: unregistering virtual destructors + `clear_registry()` on both, added to `clear_registries()`. **The WS client iterates these under the `SKListener` semaphore, so the destructors and clear take that same semaphore** — mirroring `~SKEmitter` naively would have traded the dangling pointer for a use-after-free race with the WS task. |
| 2 | StatusPageItem keep-first-duplicate guard untested | Added `test_statuspageitem_duplicate_name_guard`: a duplicate destroyed leaves the first registration intact (asserts the exact `it->second == this` branch). |
| 3 | clear test didn't prove the destruction-order interplay | Added `test_clear_registries_destroys_config_owned_object`: drops the local owner so `config_items_` is the sole owner, then asserts a `weak_ptr` expires after the clear — proving the owning-registry clear destroys the object and triggers its raw-registry unregister mid-clear. |
| 4 | `clear_registries()` duplicated across both `destroy()` overrides | `SensESPApp::destroy()` now drops its global ref first and delegates to `SensESPBaseApp::destroy()`; the shared teardown lives in one place. |
| 5 | `clear_registries()` ran even when teardown failed → stranded live objects | Gated the clear on a successful teardown (`!outside_users`). |
| 7 | Pre-clear assertions used `size >= 1` floors | Switched to per-object membership (`sources_contains`, `get_config_item`, `count(name)`). |
| 9 | `get_transforms()` deprecation message implied a non-equivalent replacement | Reworded: "deprecated and unused internally; no direct replacement." |
| 10 | Raw-vs-shared_ptr asymmetry under-documented | Added a note at `ConfigItemBase`/`UIButton` explaining why they have no self-erasing destructor. |
| 12 | Missing edge-case tests | Added idempotent double-clear; listener unregister coverage. |

### Deliberately not changed (with reasoning)

- **#6 (no event-loop/HTTP quiesce before clear) and the static-destruction-order residual** — real but narrow, and a proper fix means tearing down the event loop / httpd task inside `destroy()`, which is a larger change outside this PR's scope. `destroy()` is rarely called outside tests. Flagging as known residual risk rather than papering over it.
- **#11 (duplicate-name first-destroyed orphans the entry)** — left as-is. Promoting a surviving duplicate is complex, and duplicate names are already a logged misconfiguration. The guard is correct for the common case.
- **#8 (changelog entry)** — the `CHANGELOG.md` is dormant (last entries are 2020-era PRs; none of the recent v3 work is recorded). Adding a single entry would be inconsistent. The ABI note (added virtual destructors change vtable layout — source-compatible, ABI-incompatible; fine for PlatformIO source builds) and the behavioral note (`destroy()` now clears registries on teardown) belong with the release/version bump.

### Verification

Full suite re-run on HALMET hardware (pioarduino_esp32): all suites pass, including the expanded `test_registries` (8 cases). The listener destructors, `destroy()` dedup, and teardown gating introduced no regressions.

PR #1018, opened 2026-06-18
fix(log): capture log lines off the caller's stack to stop heap corruption

mairas:
## Problem

`LogBuffer`'s `esp_log_set_vprintf` hook runs in the context of whatever task emitted the log line, and it formatted each line **twice** on that task's stack: the chained previous handler (newlib `vfprintf` → console) plus a second `vsnprintf` into a 256-byte stack buffer for capture, then the ANSI-strip / `std::string` / deque work inline. That adds ~340 B of stack to **every** `ESP_LOGx` call.

Tasks that log with a small stack overflow as a result. The reproducer was the `NMEA2000_twai` TWAI error-monitor task (2048 B stack), which logs a bus-off line whenever the device isn't on a live, terminated CAN bus. The hook pushed its peak stack use to ~2.2 KB, overran the 2048 B stack, and corrupted the adjacent heap — which then tripped a heap-poisoning abort on the next `free()` (often in an unrelated task). The result was a reboot loop that looked like random memory corruption rather than a logging bug. Disabling the hook, or growing the task stack, made it stable — confirming the cause is the hook's added caller-stack cost, not a logic bug.

## Fix

Make the hook cheap on the caller's stack:

- **Format once, self-emit.** On ESP-IDF log V1 (the hook is called once per line with the complete format string), `vsnprintf` once and write the bytes with `fwrite(…, stdout)` — following the active console, so it's correct on both UART and USB-Serial-JTAG — instead of chaining a second full `vfprintf`. This removes the heavyweight format from the caller path; the caller-stack peak drops below even the no-capture baseline.
- **Capture off-task.** The formatted bytes go to a dedicated capture task through a fixed-size queue; the heap work (ANSI strip, `std::string`, deque) runs on that task's stack. The hook no longer allocates, locks, or touches the deque.
- **Full console fidelity.** Over-length lines (longer than the capture buffer) fall back to the chained handler so the whole line and its trailing newline still reach the console. These come from roomy-stack loggers, not the small-stack tasks the lean path protects.
- **Gated.** ISR / pre-install context falls back to the chained handler. The self-emit path is `#if`-gated to log V1; log V2 (hook called multiple times per line with fragments) keeps the chained double-format path.

## Validation (on hardware)

- **SH-ESP32** (Xtensa, Arduino, UART, colors off): the TWAI error-monitor task drops from ~2236 B (overflow at 2048 → heap corruption) to **~1744 B**. Stable 500 s; console output byte-identical.
- **ESP32-C3** (RISC-V, ESP-IDF, USB-Serial-JTAG console, `CONFIG_LOG_COLORS=y`): the V1 self-emit path runs; console output and colors are faithful and long lines print in full. Stable 240 s under ~10 Hz delta logging.

The `/api/log` capture and serving code is unchanged; only the emit side changed. RAM cost is neutral.

### DIFF
--- a/src/sensesp/system/log_buffer.cpp
+++ b/src/sensesp/system/log_buffer.cpp
@@ -28,20 +28,123 @@ LogBuffer::LogBuffer(size_t max_lines, uint32_t max_age_ms,
 }
 
 void LogBuffer::install() {
+  // Idempotent: a second call would set previous_vprintf_ to the trampoline
+  // itself (esp_log_set_vprintf returns the currently installed handler),
+  // turning the fallback paths into unbounded recursion. After the first
+  // install previous_vprintf_ is the (always non-null) prior handler.
+  if (previous_vprintf_ != nullptr) {
+    return;
+  }
+
   // Draw the per-boot session id here rather than in the constructor: the hook
   // is installed during setup(), past early static init, and mixing in the boot
   // timer makes a repeated value across reboots unlikely even before the RF
   // subsystem seeds the RNG. The client also falls back to a cursor reset.
   session_id_ = esp_random() ^ static_cast<uint32_t>(esp_timer_get_time());
+
+  // Stand up the handoff queue and capture task before the hook goes live, so
+  // the first captured line already has somewhere to go. If the task can't be
+  // created, drop the queue too: a non-null queue with no drainer would fill
+  // after kLogCaptureQueueDepth lines and silently lose all capture for the
+  // rest of the uptime. Leaving capture_queue_ null makes the trampoline fall
+  // back to the chained handler (console preserved, capture disabled).
+  capture_queue_ = xQueueCreate(kLogCaptureQueueDepth, sizeof(LogQueueItem));
+  if (capture_queue_ != nullptr &&
+      xTaskCreate(&LogBuffer::capture_task, "LogCapture",
+                  kLogCaptureTaskStackSize, this, kLogCaptureTaskPriority,
+                  nullptr) != pdPASS) {
+    vQueueDelete(capture_queue_);
+    capture_queue_ = nullptr;
+  }
+
   instance_ = this;
   previous_vprintf_ = esp_log_set_vprintf(&LogBuffer::vprintf_trampoline);
+}
+
+void LogBuffer::enqueue_capture(LogQueueItem& item, int n) {
+  // Drop empty/error results; push_line would discard them anyway.
+  if (n <= 0) {
+    return;
+  }
+  item.length = (static_cast<size_t>(n) < sizeof(item.text))
+                    ? static_cast<uint16_t>(n)
+                    : static_cast<uint16_t>(sizeof(item.text) - 1);
+  item.timestamp_ms = esp_log_timestamp();
+  // Non-blocking: drop the line rather than stall the logger if the capture
+  // task is behind.
+  xQueueSend(capture_queue_, &item, 0);
+}
+
+void LogBuffer::capture_task(void* arg) {
+  auto* self = static_cast<LogBuffer*>(arg);
+  LogQueueItem item;
+  while (true) {
+    if (xQueueReceive(self->capture_queue_, &item, portMAX_DELAY) == pdTRUE) {
+      // The heap-allocating work (ANSI strip, std::string, deque) happens here,
+      // on this task's stack, not the logging task's.
+      self->push_line(item.text, item.length, item.timestamp_ms);
+    }
+  }
 }
 
 int LogBuffer::vprintf_trampoline(const char* format, va_list args) {
   LogBuffer* inst = instance_;
 
-  // Forward to the previously installed handler so UART output is unchanged.
-  // A va_list can be traversed only once, so each consumer gets its own copy.
+#if !defined(ESP_LOG_VERSION) || ESP_LOG_VERSION == 1
+  // ESP-IDF log V1: the hook is called once per line with the complete format
+  // string and va_list, so a single vsnprintf reproduces the exact console
+  // bytes. We format ONCE on the caller stack, emit the bytes ourselves, and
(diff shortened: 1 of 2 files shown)


mairas:
## Code review — format-once LogBuffer

Six persona reviewers (correctness, testing, maintainability, project-standards, adversarial, reliability), each in a fresh context. Merged + deduplicated below. The core change is sound — the concurrency, `va_list`/`va_end` discipline, truncation math, and install ordering all check out — but there's one robustness gap three reviewers independently flagged, plus quality fixes.

### P2 — should fix before merge

| # | File:line | Issue | Reviewers | Conf |
|---|-----------|-------|-----------|------|
| 1 | log_buffer.cpp:39 | **Unchecked `xTaskCreate` → silent, permanent capture death.** `install()` creates the queue, then the task, but doesn't check `xTaskCreate`. If the task fails to spawn (OOM at boot — plausible on the memory-pressured devices this targets) while the queue succeeded, `capture_queue_` stays non-null, the lean path keeps enqueueing, the 16-slot queue fills, and every line drops forever — console fine, `/api/log` silently dead for the whole uptime. **Fix:** check the return; on failure delete the queue + null `capture_queue_` so it falls back to the chained handler. | correctness, adversarial, reliability | 0.92 |
| 2 | log_buffer.cpp:114 | **Capture-enqueue tail duplicated 3×** across the `#if` branches; because only one branch compiles, a missed copy passes CI and breaks on the other ESP-IDF log version. **Fix:** extract a helper. | maintainability | 0.80 |
| 3 | log_buffer.cpp:61 | **New trampoline path ships with zero tests** (existing tests drive `push_line` directly). Tests run on-device (Unity), so an `install()` → emit → drain → `snapshot_since` test is feasible. **Address:** add a focused test or track as follow-up. | testing | 0.90 |

### P3 — quick quality fixes

| # | File:line | Issue | Reviewers | Conf |
|---|-----------|-------|-----------|------|
| 4 | log_buffer.cpp:45 | **Double `install()` self-chains `previous_vprintf_` into recursion** + leaks the prior task/queue. One caller today; a cheap guard removes the foot-gun. | adversarial | 0.78 |
| 5 | log_buffer.cpp:99 | **Over-length lines re-introduce the full `vfprintf` on the caller stack** — by design (only >256-char lines, from roomy-stack loggers), but the comment states it as fact, not a documented limitation. | reliability | 0.75 |
| 6 | log_buffer.cpp:113, .h:74 | **clang-format 80-col violations** (diff-introduced). | project-standards | 0.90 |
| 7 | log_buffer.cpp:41 | **Magic task stack (4096) / priority (1)** vs the house pattern (`kWsClientTaskStackSize`). | project-standards | 0.70 |
| 8 | log_buffer.h:85 | **`max_line_length` default (256) and `kLogCaptureLineMax` (256)** are independent literals that must stay equal per the header's own documented invariant. | maintainability | 0.70 |
| 9 | log_buffer.h:142 | **`capture_task_handle_` stored but never read.** | maintainability | 0.82 |
| 10 | log_buffer.cpp:114 | **V1 enqueues zero-length items (`n==0`); V2 guards with `n>0`** — harmless (`push_line` drops them) but inconsistent. | correctness | 0.90 |

### Advisory / residual
- Queue-drop (full) produces no `gap` signal — the web log can't distinguish dropped-under-load from absent. Possible follow-up.
- Re-entrancy if heap-exhaustion diagnostics log on the capture task (low confidence).
- `fwrite` return ignored / partial writes truncate console output (best-effort; same console behavior as before).

### Refuted (checked, not issues)
- Console line-tearing between the self-emit and chained paths — both serialize through the newlib `stdout` FILE lock.
- `instance_` publish ordering — the early-return path tolerates a stale `nullptr` read.
- `va_list`/`va_end` on every path including the `n<0` early return; truncation/length math (no over-read).

**Verdict: Ready with fixes.** Fix #1 before merge; #2, #4, #6–#10 are quick; #3 (tests) to address or track.

mairas:
Addressed the review in `590c8e1` (force-pushed). Summary:

**Fixed**
- **#1 (P2) unchecked `xTaskCreate`** — now checked; on failure the queue is deleted and `capture_queue_` nulled, so the trampoline cleanly falls back to the chained handler (console preserved, capture disabled) instead of enqueueing into a queue no task drains.
- **#2 (P2) 3× duplicated enqueue tail** — extracted a private `enqueue_capture()` helper, used by both the V1 and V2 paths. Its `n <= 0` guard also resolves **#10** (V1 no longer enqueues zero-length items).
- **#4 double `install()` recursion** — `install()` is now idempotent (early-return if already installed).
- **#5 over-length caveat** — comment reworded as a documented limitation (the lean path's protection relies on over-length lines coming from roomy-stack loggers) rather than asserted fact.
- **#6 clang-format 80-col** — wrapped (without the include reordering clang-format v22 wanted, which deviates from the repo's committed two-block style).
- **#7 magic task stack/priority** — named `kLogCaptureTaskStackSize` / `kLogCaptureTaskPriority`, matching the `kWsClientTaskStackSize` precedent.
- **#8 `max_line_length` default** — now defaults to `kLogCaptureLineMax` instead of a bare literal.
- **#9 unused `capture_task_handle_`** — dropped (task is app-lifetime, never torn down; documented).

Re-verified on the ESP32-C3 (ESP-IDF / USB-Serial-JTAG / colors): compiles, runs stable, over-length deltas still print in full.

**Residual / not changed**
- **#3 (P2) tests** — acknowledged. The new trampoline path has no host-testable seam (existing tests drive `push_line` directly; reaching the trampoline → queue → capture-task round trip needs FreeRTOS + an on-device run). The change is hardware-validated (SH-ESP32 + C3); I'd rather track an on-device round-trip + over-length test as a follow-up than ship a partial one. Happy to file an issue.
- Advisory: queue-full drop produces no `gap` signal, and the `fwrite` return is unchecked (best-effort console, same as the prior handler) — left as documented residuals.

Issue #1019, opened 2026-06-18
Add tests for the LogBuffer format-once capture path


mairas:
PR #1018 reworked the `LogBuffer` `esp_log` vprintf hook (format once, self-emit via `fwrite(stdout)`, hand off to a dedicated capture task via a fixed-size queue, with an over-length chained fallback and a V1/V2 split). It was validated on hardware (SH-ESP32 / Xtensa / UART and ESP32-C3 / RISC-V / USB-Serial-JTAG), but ships without automated tests for the new path — the existing `test/system/test_log_buffer` suite drives `push_line()` / `snapshot_since()` directly and never exercises the trampoline.

This issue tracks closing that gap. The LogBuffer tests already run on-device (Unity), so the round trip is reachable there.

### Worth adding (reachable on-device)
- **Round trip:** call `install()`, emit a real `ESP_LOGx` line, let the capture task drain, then assert via `snapshot_since()` that the line was captured. Covers trampoline → queue → capture task → `push_line`.
- **Over-length branch:** emit a line longer than `kLogCaptureLineMax` (256) and assert the captured copy is truncated to the cap while the full line still reaches the console (this is the trailing-newline / line-concatenation regression the over-length fallback exists to prevent).
- **Queue-full drop:** emit more than `kLogCaptureQueueDepth` (16) lines faster than the capture task drains; assert lines are dropped (not blocked) and the buffer/cursor stay consistent.
- **`install()` idempotency:** a second `install()` is a no-op (guards against re-chaining the hook into recursion).

### Genuinely hardware-only (out of scope for unit tests)
- `fwrite(stdout)` console routing across UART vs USB-Serial-JTAG vs USB-CDC.
- The ESP-IDF log **V2** fragment path (the `#else` branch) — only the active toolchain's `ESP_LOG_VERSION` is compiled.
- `xPortInIsrContext()` ISR fallback; FreeRTOS task scheduling / stack sizing.

Surfaced by the code review on #1018.

PR #1020, opened 2026-06-18
fix(net): disconnect before applying the next WiFi config so multi-AP failover works

mairas:
## Problem

`WiFiProvisioner::start_client_autoconnect`'s `reconnect_cb` cycles through the configured client networks every 20 s, calling `WiFi.begin()` with the next SSID whenever the device isn't connected. But the ESP32 STA rejects a config change while it's still connecting or auto-reconnecting to the previous SSID:

```
E wifi:sta is connecting, cannot set config
E ARDUINO: STA clear config failed! 0x3006: ESP_ERR_WIFI_STATE
```

`WiFi.begin()` only auto-disconnects when `status() == WL_CONNECTED` — not when it's merely *connecting* — and Arduino-ESP32's default STA auto-reconnect keeps the radio in that connecting state indefinitely. So `begin()` for a *different* network silently fails, and the device never fails over past the first configured AP. If the first network is absent, it stays stuck on it and never connects to the others.

## Fix

Drop the in-progress attempt with `WiFi.disconnect(false)` before `WiFi.begin()`, so the STA leaves the connecting state and the new config takes effect.

## Verification (on hardware)

SH-ESP32 with two configured networks, the first (Hurma) absent:

- **Before:** every config switch logged `ESP_ERR_WIFI_STATE`; the device stayed stuck on the first SSID.
- **After:** zero `ESP_ERR_WIFI_STATE` errors; the device cycled Hurma → Stairway to Heaven and **connected on the second config**, then stopped cycling (status connected) — i.e. failover now works.

### DIFF
--- a/src/sensesp/net/wifi_provisioner.cpp
+++ b/src/sensesp/net/wifi_provisioner.cpp
@@ -136,6 +136,7 @@ void WiFiProvisioner::start_client_autoconnect() {
   auto reconnect_cb = [this]() {
     static uint32_t attempt_num = 0;
     static uint32_t current_config_idx = 0;
+    static int last_applied_idx = -1;
 
     int num_configs = client_settings_.size();
 
@@ -194,7 +195,19 @@ void WiFiProvisioner::start_client_autoconnect() {
       WiFi.config(config.ip_, config.gateway_, config.netmask_,
                   config.dns_server_);
     }
+    // The ESP32 STA rejects a new config (ESP_ERR_WIFI_STATE, "sta is
+    // connecting, cannot set config") while it is still connecting or
+    // auto-reconnecting to the previous SSID, so switching networks needs the
+    // in-progress attempt dropped first. Only disconnect when the target config
+    // actually changes; re-applying the same config (a single network, or a
+    // list with one valid slot) must leave a slow association running rather
+    // than restart it every cycle.
+    int applied_idx = static_cast<int>(current_config_idx % num_configs);
+    if (applied_idx != last_applied_idx) {
+      WiFi.disconnect(false);
+    }
     WiFi.begin(config.ssid_.c_str(), config.password_.c_str());
+    last_applied_idx = applied_idx;
     attempt_num++;
     current_config_idx++;  // Move to the next config for the next attempt
   };
@@ -203,8 +216,9 @@ void WiFiProvisioner::start_client_autoconnect() {
   reconnect_cb();
 
   // Launch a separate onRepeat event to (re-)establish WiFi connection.
-  // Connecting is attempted only every 20 s to allow the previous connection
-  // attempt to complete even if the network is slow.
+  // Attempts are spaced 20 s apart: when staying on the same network this lets
+  // a slow association finish; when failing over it bounds how long a missing
+  // network is tried before the next configured AP.
   event_loop()->onRepeat(20000, reconnect_cb);
 }
 


mairas:
## Code review — multi-AP WiFi failover fix

**Scope:** `src/sensesp/net/wifi_provisioner.cpp`, 1 commit (`bc86267`) vs `main`@`8dd9761`.
**Intent:** Add `WiFi.disconnect(false)` before `WiFi.begin()` in `reconnect_cb` so the STA leaves the connecting state and an alternate client config applies, fixing multi-AP failover (otherwise `WiFi.begin()` for a new SSID fails with `ESP_ERR_WIFI_STATE`). Hardware-verified on SH-ESP32.
**Reviewers:** correctness, reliability, testing, maintainability, project-standards — 5 personas, each in an independent context. Skipped security / api-contract / adversarial (no auth, endpoint, or contract surface; diff is one functional line, well under the adversarial LOC gate).

The fix is correct for the cases that matter. The `WL_CONNECTED` early-return (line 142) means a healthy link is never torn down; the initial `reconnect_cb()` call disconnects a harmless idle STA; and `disconnect(false)` resolves to `wifioff=false, eraseap=false`, so it neither powers down the radio nor wipes stored credentials. The hardware-verified failover (config0 absent → config1 connects) holds. Findings below are one behavior tradeoff plus minor cleanups. project-standards found nothing (conventional-commit subject, comment style, and the absent CHANGELOG/CONTRIBUTING requirements all check out).

### P2 — Moderate

| # | Location | Issue | Reviewer | Conf |
|---|----------|-------|----------|------|
| 1 | `wifi_provisioner.cpp:203` | **Unconditional `disconnect(false)` can abort a slow same-AP association.** `reconnect_cb` runs every 20 s while not connected and now drops any in-progress attempt before every `begin()` — including when re-applying the *same* config (single configured network, or a multi-config list where only one slot is valid). A network that legitimately needs >20 s to associate gets killed and restarted each cycle and may never complete. Pre-fix, the redundant same-SSID `begin()` was rejected with `ESP_ERR_WIFI_STATE` and the original attempt was left to finish — which is exactly what the comment at lines 213-214 still promises. The disconnect is only needed when *switching* configs. | reliability (correctness noted the same risk) | 0.66 |

Suggested refinement: track the last-applied config index and call `disconnect(false)` only when the target index differs. Traced against the verified scenario it still disconnects on the 0→1 switch, so failover is preserved, while a slow single-AP association is left to run. Not a hard blocker: the failure is bounded (each cycle retries, and it reconnects as soon as association fits inside a 20 s window) and only bites pathological >20 s associations.

### P3 — Low

| # | Location | Issue | Reviewer | Conf |
|---|----------|-------|----------|------|
| 2 | `wifi_provisioner.cpp:213` | **Scheduling comment now contradicts the code.** "every 20 s to allow the previous connection attempt to complete even if the network is slow" — the new `disconnect(false)` aborts that attempt each cycle. A maintainer trusting the comment would reason wrongly about retry/failover timing. In scope for this PR (the diff created the contradiction). | correctness | 0.74 |
| 3 | `wifi_provisioner.cpp:160` | **Failover config-selection logic is untested** (pre-existing). Wraparound via `% num_configs`, skipping empty slots, and the index advance — the behavior this fix exists to enable — has no coverage. State lives in `static` locals inside the lambda, so it can't be exercised without hardware. | testing | 0.66 |
| 4 | `wifi_provisioner.cpp:206` | **Reconnect index incremented in two places** (pre-existing). The for-loop advances `current_config_idx` (160-162), then line 206 increments it again, on top of `% num_configs` indexing and a self-assigning `for (current_config_idx = current_config_idx; …)`. Correct but hard to follow; an easy spot to introduce an off-by-one later. | maintainability | 0.62 |

### Coverage
- 5/5 reviewers returned. No findings suppressed at the confidence gate (all ≥ 0.60).
- Residual risks: `disconnect(false)` → `begin()` is asynchronous; the back-to-back ordering is driver/version dependent (clean on SH-ESP32, not provable from code for other ESP32 variants or arduino-esp32 versions). The static-IP path (`use_dhcp_ == false`) combined with the new disconnect was not separately verified.
- Testing gaps: no host/native test harness exists for WiFi code (consistent with the limitation tracked in #1019); failover index selection and `reconnect_cb` state transitions are hardware-only.

### Verdict: Ready with fixes
1. **Required:** fix the stale 20 s comment (P3 #2).
2. **Decision:** P2 #1 — gate the disconnect on a config change (recommended) vs keep it unconditional and reword the comment to match the abort-and-retry behavior.

Pre-existing P3s don't block this PR. #3 (failover-logic test gap) is a candidate to fold into the test-coverage tracking alongside #1019; #4 is advisory.

<sub>🤖 Independent multi-persona review via <code>mairas-workflow:review</code></sub>

mairas:
## Review addressed — disconnect is now conditional

Pushed `2a6ec15` (force-update of the single commit). Changes made in response to the review above:

- **P2 #1 (slow same-AP association):** `WiFi.disconnect(false)` now fires only when the target config index changes. Re-applying the same config (a single network, or a multi-config list with one valid slot) skips the disconnect, so a slow association runs to completion instead of being aborted every 20 s. Traced against the verified scenario the 0→1 switch still disconnects, so failover is preserved. A `static int last_applied_idx` tracks the last config passed to `begin()`.
- **P3 #2 (stale comment):** the 20 s `onRepeat` comment now describes the actual behavior — same network lets a slow association finish; failover bounds how long a missing network is tried before the next config.

Pre-existing P3s left as-is: the failover-selection test gap (#3, a candidate to fold into #1019's test-coverage tracking) and the double-increment readability of the config-cycling loop (#4, advisory).

Compile-verified locally on `pioarduino_esp32c3` (SUCCESS); CI re-running on the new commit.

PR #1021, opened 2026-06-18
fix(sensors): register DigitalInputDebounceCounter events and restore its config

### LINKED ISSUE #816: DigitalInputDebounceCounter not reacting to the event loop
The DigitalInputDebounceCounter does not react to the event loop. Most probably is the same problem discussed in #795 

The following adjustments have been made to 'digital_input.h' to solve the issue 

Duplicating the below code from the `DigitalInputCounter` into the constructor of `DigitalInputDebounceCounter`

```        
event_loop()->onInterrupt(pin_, interrupt_type_, interrupt_handler_);
event_loop()->onRepeat(read_delay_, [this]() {
    noInterrupts();
    output_ = counter_;
    counter_ = 0;
    interrupts();
    notify();
    });
```

and changing the member inheritance from `private:` to `protected:` in `DigitialInputCounter` for the variables 
``` 
int interrupt_type_;
std::function<void()> interrupt_handler_;
```

zeusec:
## Summary

This fixes two bugs in `DigitalInputDebounceCounter`. One left the sensor completely inert, the other quietly dropped its saved `ignore_interval` on reboot.

- Registers the interrupt and repeat events in the constructor body, like the public `DigitalInputCounter` ctor, instead of delegating to the protected base ctor and registering nothing.
- Loads config from the derived constructor instead of the base one, so the debounce counter's own `from_json()` runs and a persisted `ignore_interval` survives a reboot.
- Adds a `LoadConfig` flag for that. The plain `DigitalInputCounter` still loads in the base ctor, just as before.

Fixes #816

The `ignore_interval` bug has no issue of its own.. I found it reviewing this one.

## Verification

- Full PlatformIO test suite on ESP32 (esp32dev): 127/127 across 16 suites, no regressions.
- Confirmed on hardware that the debounce counter registers its events and emits... It fails before the fix and passes after.

### DIFF
--- a/src/sensesp/sensors/digital_input.h
+++ b/src/sensesp/sensors/digital_input.h
@@ -180,7 +180,17 @@ class DigitalInputDebounceCounter : public DigitalInputCounter {
                               String config_path = "")
       : DigitalInputCounter(pin, pin_mode, interrupt_type, read_delay,
                             config_path, [this]() { this->handleInterrupt(); }),
-        ignore_interval_ms_{ignore_interval_ms} {}
+        ignore_interval_ms_{ignore_interval_ms} {
+    repeat_event_ = event_loop()->onRepeat(read_delay_, [this]() {
+      noInterrupts();
+      output_ = counter_;
+      counter_ = 0;
+      interrupts();
+      notify();
+    });
+    isr_event_ = event_loop()->onInterrupt(
+        pin_, interrupt_type, [this]() { this->handleInterrupt(); });
+  }
 
  private:
   void handleInterrupt();
--- a/src/sensesp/sensors/digital_input.h
+++ b/src/sensesp/sensors/digital_input.h
@@ -118,6 +118,11 @@ class DigitalInputCounter : public DigitalInput, public Sensor<int> {
   virtual bool from_json(const JsonObject& config) override;
 
  protected:
+  // load() calls the virtual from_json(). Inside a base-class constructor the
+  // dynamic type is still the base, so loading there dispatches to the base
+  // from_json() and silently drops fields a derived class adds. A derived class
+  // with extra config therefore constructs with kNo and calls load() from its
+  // own constructor body, where the full dynamic type is established.
   enum class LoadConfig { kNo, kYes };
 
   DigitalInputCounter(uint8_t pin, int pin_mode, int interrupt_type,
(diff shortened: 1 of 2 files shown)


mairas:
Thanks for this — I traced both bugs and the fix against the full source and it's correct on both counts.

**Bug 1 (inert sensor, #816):** confirmed. The old `DigitalInputDebounceCounter` ctor delegated to the *protected* `DigitalInputCounter` ctor, whose body only calls `load()` — it never registers `isr_event_` or `repeat_event_`, which live only in the *public* ctor body. So the debounce counter registered no interrupt and no repeat callback. Your fix registers both in the derived ctor body.

**Bug 2 (dropped `ignore_interval`):** good catch, and the root cause is exactly the classic one — `load()` → virtual `from_json()` was being called from the *base* ctor, where the dynamic type is still `DigitalInputCounter`, so the base `from_json` ran (reads only `read_delay`) instead of the derived override. Moving `load()` into the most-derived ctor body fixes the dispatch. The `LoadConfig` flag keeps plain `DigitalInputCounter` behavior byte-for-byte unchanged. I checked the delegation chain — no double-load, no double-registration, and destructor cleanup is unaffected since the events live in the base.

One maintainability follow-up: the `onRepeat` block is now duplicated verbatim between the public `DigitalInputCounter` ctor and the debounce ctor, and the debounce path builds a second `handleInterrupt` lambda even though the base already stored one (it's only working around `interrupt_handler_` being private). A small protected `register_events()` helper on the base — callable from both ctors, using the existing `interrupt_type_`/`interrupt_handler_` members — collapses all of that.

Since it's a minor chore and maintainer-edits are enabled on this PR, **we'll push that cleanup onto your branch ourselves** so you don't have to round-trip on it — your fix lands either way. No automated regression test is added; both bugs only manifest through construction-time interrupt registration and the config-load path, neither of which the current host round-trip tests can exercise, so hardware verification is the right call here.

mairas:
Pushed the follow-ups onto the branch via maintainer edit — three commits on top of yours:

- `refactor`: extracted a shared `register_events()` helper so the public counter and the debounce counter register events through one path (removes the duplication and the duplicate `handleInterrupt` lambda).
- `docs`: a few short *why* comments on `LoadConfig` and the debounce ctor's `load()` call, so the virtual-dispatch-during-construction reasoning is on record and a future cleanup doesn't silently reintroduce the dropped-config bug.
- `test`: a regression test (`test/system/test_debounce_config_load`) that seeds a persisted config, constructs a fresh counter through the construction-time `load()` path, and asserts `ignore_interval` survived. A plain `to_json`/`from_json` round-trip wouldn't catch this bug, since it runs `from_json` on an already-complete object where dispatch is correct either way.

The test compiles cleanly under the CI test matrix. Note CI builds the test suites but doesn't run them (`pio test --without-testing`), so the assertions execute when the suite is run on hardware — same as your 127/127 run.

Nothing needed from you. Thanks again for the fix.

PR #1023, opened 2026-06-21
fix(ui): bound and retry config card fetches

### LINKED ISSUE #1022: Config Cards blank on socket-constrained boards: throttle + retry per-card config fetches
## Problem

On socket-constrained boards (reported on a HALSER ESP32-C3), opening the **Configuration** tab leaves some Config Cards permanently blank. The device log shows:

```
E httpd: httpd_accept_conn: error in accept (23)
W httpd: httpd_server: error accepting new connection
```

## Root cause

`errno 23` is `ENFILE` — the lwIP socket table is exhausted at the `accept()` syscall.

`ConfigCards` mounts every `<ConfigCard>` in a single commit, and each card independently fires `fetch(/api/config/<path>)` on mount with **no concurrency limit**. The browser opens ~6 parallel connections. On the device the lwIP pool (10 fds) is already over-subscribed: httpd's `max_open_sockets=7` + listener + the Signal K WebSocket + mDNS sum to ~10, so the burst's next `accept()` fails with `ENFILE`. The dropped connection rejects the `fetch`; `ConfigCard.updateFunc` has **no try/catch**, so the rejection throws past all `setState`. The card stays in its loading placeholder forever — no error, no retry.

`lru_purge_enable` (already `true`) does not help here: it manages httpd's own session slots, not the global lwIP fd table, so the `ENFILE` happens below it. The fix belongs in the frontend: bound concurrency and recover dropped requests.

## Approach

One module-scoped limiter + retry, threaded through the existing `fetchConfigData` choke point; migrate `ConfigCard` off its inline fetch. No new dependency.

### Decisions
- **New `frontend/src/common/requestQueue.ts`** — module-level (shared across all cards) semaphore: FIFO waiter queue + in-flight counter. A per-component limiter cannot bound cross-component concurrency, which is the bug. Plus retry-with-backoff: ~3 attempts, exponential backoff with **full jitter** (without jitter, retries re-synchronize into a fresh burst). AbortSignal-aware. Slot accounting is the highest-severity surface — a leaked/double-released slot deadlocks the whole page, worse than today: acquire on dequeue, release in a `finally` on success/reject/abort; abort-while-queued removes the waiter without releasing a slot it never held.
- **Concurrency K = 2.** The ~4 client slots are shared with the SK WebSocket + transient token/redirect client at startup, and `/api/info` `/api/wifi` `/api/routes` polling hits the same device sockets while bypassing this limiter (it only bounds `/api/config` GETs). K is a single named constant in `requestQueue.ts`, droppable to 1 if a board still resets.
- **`fetchConfigData`** routed through the limiter, accepts an optional `AbortSignal`. Return shape and final thrown-Error contract preserved (SignalK/WiFi panels untouched), but the retry layer sees the HTTP **status before** the wrapping catch collapses it, so it retries 5xx + network rejects but never 4xx. `ConfigData` widened with `title` and `requires_restart` (backend already returns them).
- **`ConfigCard`** migrated onto `fetchConfigData`. Explicit per-card phase (`loading | loaded | failed`) as the single source of truth — replaces the current scheme that infers loading from `schema.properties` and "fetched" from `config` emptiness (two variables that diverge). `try/catch` sets `failed`; `AbortError` is a non-failure (never shows Retry). One `AbortController` per effect run; cleanup aborts, dequeues, and clears backoff timers idempotently (StrictMode/double-mount safe). Visible failed + manual Retry branch so a dropped card is never a silent blank.
- **`ConfigCards`** routes `?cards` through the helper and replaces the silent `catch → return []` (which masks an unreachable device as "zero cards") with a visible error. Adds abort cleanup.

### Out of scope
- Gating card fetches until the SK connection settles — only if K=2 proves insufficient on-device.
- Backend `get_item_data` JsonDocument-overflow silent truncation — a separate bug, to be filed independently if confirmed.

## Tasks
- [ ] Add `frontend/src/common/requestQueue.ts`: module-scoped limiter (K=2) + jittered-backoff retry, AbortSignal-aware, leak-safe slot accounting
- [ ] Route `fetchConfigData` through the limiter; expose status to the retry decision; widen `ConfigData` with `title` + `requires_restart`
- [ ] Migrate `ConfigCard`: explicit phase enum, AbortController cleanup, AbortError-as-non-failure, failed + Retry UI
- [ ] Route `ConfigCards` `?cards` through the helper; surface unreachable-device error; abort cleanup
- [ ] Stand up minimal jest + a `requestQueue` unit test (never exceeds K; releases slot on success/reject/abort; abort-while-queued dequeues)
- [ ] Verify via mock dev server + Playwright network interception (≤K in flight, every card body renders, retry + failed-state work, no leaked timers on navigate)
- [ ] Regenerate and commit `src/sensesp/net/web/autogen/frontend_files.h` (CI does not rebuild the bundle)
- [ ] Best-effort device verification: reload Config tab repeatedly → zero permanent blanks; no `accept (23)` in the log

mairas:
## Why

On socket-constrained boards (reported on a HALSER ESP32-C3), opening the **Configuration** tab leaves some Config Cards permanently blank, with `httpd_accept_conn: error in accept (23)` in the device log.

`error 23` is `ENFILE` — the lwIP socket pool is exhausted. `ConfigCards` mounts every `<ConfigCard>` in one commit and each card independently fires `fetch(/api/config/<path>)` with no concurrency limit, so the browser opens a burst the device can't accept. The dropped connection rejects the fetch, but `ConfigCard.updateFunc` had no `try/catch`, so the rejection escaped before any state update and the card stayed in its loading placeholder forever — no error, no retry. `lru_purge_enable` (already on) can't help: it manages httpd's session slots, not the global lwIP fd table, so the failure happens below it.

## How

Fix it where the unrealistic assumption lives — the client — rather than buying more device RAM.

- **`common/requestQueue.ts` (new):** a shared, module-scoped concurrency limiter (max 2) plus retry-with-jittered-backoff, AbortSignal-aware. One slot is held per scheduled request for its whole lifetime and released exactly once, so the in-flight count can't leak or deadlock. Retries dropped connections and 5xx, never 4xx, never an abort.
- **`fetchConfigData`:** routed through the queue, accepts an `AbortSignal`, and exposes HTTP status to the retry decision while preserving its existing return/throw contract (the SignalK/WiFi/System panels are untouched). `ConfigData` widened with `title` and `requires_restart` (the endpoint already returns them). Added `fetchConfigPaths` for the card list.
- **`ConfigCard`:** migrated off its inline fetch onto `fetchConfigData`; an explicit `loading | loaded | failed` phase replaces the previous inference from `config`/`schema` emptiness; an `AbortController` cancels the in-flight request (and any pending retry) on unmount; a cancelled request is never treated as a failure. A failed card now shows a clear message and a **Retry** button instead of an indefinite blank.
- **`ConfigCards`:** the card-list fetch goes through the queue and surfaces an unreachable-device error instead of silently rendering zero cards.

The per-card render path is unchanged, so cards still fill in progressively — now in bounded waves, and none get stuck blank.

## Verification

- New jest unit tests for `requestQueue` (concurrency never exceeds K; retry/backoff; no-retry on 4xx and on abort; abort-while-queued dequeues without leaking a slot; abort in-flight frees the slot). This PR also stands up `ts-jest`, since the repo had no test runner wired.
- `type-check` and `eslint` clean on the changed files (the 4 pre-existing `tsc` errors are unrelated and untouched).
- Driven in a browser against the mock dev server with injected latency and a failing card: a server-side counter confirmed **max 2 concurrent** `/api/config` requests, every loadable card rendered, the failing card showed the **Retry** affordance, and Retry re-issued the request. (Unbounded behavior would show up to ~6 concurrent.)

Includes the regenerated embedded bundle (`frontend_files.h`).

Closes #1022


mairas:
## Independent review + follow-ups

I ran a multi-persona review of this diff (correctness of the async limiter, reliability, maintainability, testing, UX/a11y) with an adversarial verification pass on every finding. Result: **13 confirmed findings, no blockers, no runtime defects** — the verifier downgraded both initially-"high" items to medium once it confirmed they were test-coverage/process gaps, not bugs. The tricky machinery (slot accounting, abort-while-queued, backoff cancellation) held up.

### Addressed in follow-up commits (`b7793fd`, `76ec1c4`)
- **ConfigCards had no Retry** on the unreachable-device error, unlike ConfigCard — a dead-end for the highest-value transient failure (the `?cards` list fetch). Added a Retry control mirroring ConfigCard.
- **Core retry policy was untested.** Added `configAPIClient` tests: 503 retried→resolves, 404 *not* retried (1 attempt), dropped connection retried, AbortError rethrown unwrapped. Plus `backoffMs` growth/jitter/cap tests and an abort-during-backoff test. 16 jest tests now (was 9).
- **requestQueue:** re-check the abort signal after a task resolves, so a success racing an abort is treated as a cancellation.
- **ConfigCard:** wrapped the failed-state in `role="alert"` so screen readers announce it.
- Dropped two tautological test assertions.

### Deferred — your call (not defects)
- **CI runs no jest**, so these frontend tests don't gate anything yet. The repo's `test` job only runs `pio test` (C++). Making the tests meaningful needs a frontend test job in `.github/workflows/ci.yml` — a separate infra change I didn't want to bundle here.
- **Component-level tests** (the ConfigCard phase machine) would need `jsdom` + `@testing-library/preact`, which the repo doesn't have. I verified the component behavior manually in a browser instead: server-measured **max 2 concurrent** requests, every loadable card rendered, the failing card showed Retry, and Retry re-fired.
- **Error strings** shown on failed cards are technical (e.g. `HTTP Error 404`). Could be softened for the non-programmer audience per AGENTS.md; left as-is pending your preference (some diagnostic detail is useful in the field).

Issue #1033, opened 2026-06-23
Reduce self-created FreeRTOS task footprint (fold poll-loop tasks into the event loop)


mairas:
## Summary

SensESP creates two FreeRTOS tasks of its own — `SKWSClient` and `LogCapture` — that are essentially poll loops over work that is already (or could be) event-driven. Following the pattern already applied downstream (folding a serial-reader task into the main async loop), this proposes folding those poll-loop tasks into the ReactESP event loop, and separately right-sizing the one task SensESP starts but does not own (`httpd`).

The motivation is twofold:
- **Concurrency simplicity / correctness.** Every independent task that touches the SK connection or the delta queue is a context that can race. A recent field crash on a downstream device was a concurrent-access race across exactly this kind of task boundary on the one TLS connection. Fewer contexts → fewer locks, fewer races.
- **Memory.** On a bench ESP32‑C3, per-task stack high-water marks show ~20 KB of idle headroom across all tasks. Of that, roughly ~16 KB is SensESP-addressable (≈12 KB by folding the two SensESP tasks, ≈4 KB by resizing `httpd`); the rest belongs to IDF/Arduino tasks tunable only via sdkconfig.

## Background: three categories of task

Measured on a minimal app (HALSER / ESP32‑C3), the running tasks fall into three groups:

| Category | Tasks | Lever |
|---|---|---|
| SensESP poll-loop tasks | `SKWSClient`, `LogCapture` | **Foldable** into the event loop |
| SensESP-started, IDF-bodied | `httpd` | **Resizable** stack (SensESP sets it), not foldable |
| Pure IDF/Arduino | wifi, tiT, lwip, esp_timer, mdns, sys_evt, arduino_events, async_udp, Tmr Svc | sdkconfig only |

Bench stack headroom (idle-ish high-water marks): `httpd` 7.6 KB free of 8192 · `SKWSClient` 6.2 KB free of 8192 (uses ~2 KB) · `LogCapture` 3.4 KB free of 4096 (uses ~0.6 KB) · `loopTask` 4.3 KB free.

### SKWSClient (`src/sensesp/signalk/signalk_ws_client.cpp`, `ExecuteWebSocketTask`)

[...]

PR #1034, opened 2026-06-23
Run the Signal K WebSocket client on the event loop (remove its dedicated task)

mairas:
## Motivation

Part of #1033. The Signal K WebSocket client ran on its own FreeRTOS task (`SKWSClient`, an 8 KB poll loop). That task existed only because the connect / auth / send / reconnect lifecycle made blocking calls. Folding it onto the ReactESP event loop removes a persistent task (reclaims ~8 KB), removes an independent context that races on the SK connection, and keeps the loop responsive — but only once every blocking call on that path is made non-blocking. This PR does that, then folds.

## Approach (commit chain)

1. **make delta send non-blocking** — `send_delta` uses a 0-timeout send and drops the rest of the batch on failure instead of the blocking `restart()`.
2. **make WebSocket teardown non-blocking** — `restart()` / `connect_ws` hand the old client to a detached one-shot reaper task (`stop`+`destroy` off the caller); `connect()` reaps a stale client at the top, deferring bring-up so at most one client exists. A generation tag makes the singleton event handler drop late events from a reaped client.
3. **run SK auth/connect on a worker; thread-safe client handle** — `connect()` becomes a non-blocking dispatcher that spawns a one-shot worker for the blocking auth legs (access-request / poll / SSL-detect), guarded by `auth_job_running_`. `client_` becomes `std::atomic` (reaped via `exchange`, closing a latent double-free). Also fixes a pre-existing wedge where `poll_access_request` could stick in `Authorizing` on a malformed/unexpected server response.
4. **fold the SK client onto the event loop, delete its task** — replace the task with `onRepeat(100)`; the OOM teardown path is also made non-blocking (stash + retry next tick) so it can't stall the loop; closes the send-vs-teardown lifetime window now that sends and the teardown trigger share the event-loop context.
5. **drop stale "SKWSClient task" section labels** — docs.

## Verification

On a bench HALSER (ESP32-C3) against a real signalk-server, with a producer generating live delta traffic:

- Non-blocking send: deltas keep landing, no steady-state regression.
- Detached teardown: forcing `restart()` on a *healthy* connection repeatedly → clean reconnect, flat heap across cycles, no stall (the detached reaper reclaims even a live client).
- Worker-dispatched auth: existing-token reconnect runs through the worker; the cold-token attempt runs with the loop responsive through the blocking HTTP and recovers cleanly.
- Fold: heartbeat holds `dt ≈ 1000 ms` through connect attempts and forced `restart()` cycles now running **on the event loop** (the loop-stall test); heap stable and ~8 KB higher with the task gone; no crash.

Each non-trivial commit also passed an independent adversarial review (concurrency / lifecycle / reliability) — all GO; findings were folded into the commits.

**Pending before merge** (gated on a stable LAN — the bench WiFi → SK-server link was too flaky to complete end-to-end): the cold-token flow (access-request approval → poll → connect) and a connected-state soak (deltas flowing while folded; reconnect/backoff across a server bounce and WiFi drop; ≥10 min). These code paths are verified by review plus on-device attempt/recovery, but not yet end-to-end on hardware.

## Notes

- `esp_websocket_client` `auto_reconnect` is left ON (smaller change; the detached `stop()` reclaims even a healthy client — verified against the library's stop/STOPPED-bit semantics).
- `restart()` has no callers; it is documented as event-loop-context-only (it does not check `auth_job_running_`).

### DIFF
--- a/src/sensesp/signalk/signalk_ws_client.cpp
+++ b/src/sensesp/signalk/signalk_ws_client.cpp
@@ -1506,6 +1506,7 @@ void SKWSClient::send_delta() {
     if (sk_delta_queue_->data_available()) {
       std::vector<String> deltas;
       sk_delta_queue_->get_deltas(deltas);
+      bool first = true;
       for (const auto& delta : deltas) {
         int send_result = esp_websocket_client_send_text(
             client_.load(), delta.c_str(), delta.length(), kWsDeltaSendTimeoutTicks);
@@ -1516,12 +1517,21 @@ void SKWSClient::send_delta() {
           // internally -- its disconnect/error event drives reconnect. Never
           // block or tear the connection down from here. Deltas are
           // supersedable, so drop the rest of this batch. See SignalK/SensESP#1033.
+          if (first) {
+            // get_deltas() builds one-shot metadata (units, zones, ...) into the
+            // first delta and marks it sent before it leaves the device. The
+            // first delta is the only one that can carry that metadata, so if its
+            // send is the one that fails, re-arm metadata for the next batch --
+            // otherwise the server runs without it until the next reconnect.
+            sk_delta_queue_->reset_meta_send();
+          }
           ESP_LOGW(__FILENAME__,
                    "Delta send incomplete (result=%d); dropping rest of batch",
                    send_result);
           break;
         }
         this->delta_tx_tick_producer_.set(1);
+        first = false;
       }
     }
   }
--- a/src/sensesp/signalk/signalk_ws_client.cpp
+++ b/src/sensesp/signalk/signalk_ws_client.cpp
@@ -31,6 +31,10 @@ namespace sensesp {
 constexpr int kWsClientTaskStackSize = 8192;  // Stack for ExecuteWebSocketTask
 constexpr int kWsTransportTaskStackSize = 6144;  // Stack for esp_websocket_client internal task
 constexpr TickType_t kWsSendTimeoutTicks = pdMS_TO_TICKS(5000);
+// Periodic delta telemetry must never block the caller (it will be driven from
+// the event loop). 0 = enqueue if the ws-client lock and transport are free
+// right now, otherwise fail fast. Deltas are supersedable. See SignalK/SensESP#1033.
+constexpr TickType_t kWsDeltaSendTimeoutTicks = 0;
 
 SKWSClient* ws_client;
 
@@ -1403,13 +1407,18 @@ void SKWSClient::send_delta() {
       sk_delta_queue_->get_deltas(deltas);
       for (const auto& delta : deltas) {
         int send_result = esp_websocket_client_send_text(
-            client_, delta.c_str(), delta.length(), kWsSendTimeoutTicks);
+            client_, delta.c_str(), delta.length(), kWsDeltaSendTimeoutTicks);
         if (send_result < 0) {
-          ESP_LOGE(__FILENAME__,
-                   "WebSocket send failed (result=%d), restarting",
+          // Non-blocking send (0 timeout) did not complete: either brief
+          // ws-client lock contention (retry next cycle) or the transport
+          // backpressured and esp_websocket_client aborted the connection
+          // internally -- its disconnect/error event drives reconnect. Never
+          // block or tear the connection down from here. Deltas are
+          // supersedable, so drop the rest of this batch. See SignalK/SensESP#1033.
+          ESP_LOGW(__FILENAME__,
+                   "Delta send incomplete (result=%d); dropping rest of batch",
                    send_result);
-          this->restart();
-          return;
+          break;
         }
         this->delta_tx_tick_producer_.set(1);
       }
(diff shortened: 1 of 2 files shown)


mairas:
## Review: independent adversarial pass (concurrency / reliability / correctness / standards)

Reviewed the full diff with four independent lenses; every finding was re-verified against the code (and `git` history to separate regressions from pre-existing behaviour). **13 findings, 12 confirmed — but only 2 are introduced by this PR.** The concurrency core of the fold (atomic `client_`, the generation guard, the detached reaper, the `auth_job_running_` / `teardown_in_progress_` gating) held up: no confirmed UAF/double-free and no non-benign data race.

### Must fix in this PR

**MAJOR — dropping a delta batch can permanently lose SK metadata** (`signalk_ws_client.cpp` `send_delta`, interacting with `signalk_delta_queue.cpp`).
`get_deltas()` puts metadata into `output[0]`, and `add_metadata()` sets `meta_sent_ = true` *while building the delta, before any send*. In the new `send_delta()`, if the 0-tick send of `output[0]` returns `<0` (ws-client lock contention or transport backpressure — most likely on the first send after `on_connected`, racing the `subscribe_listeners` send on the transport task), the code does `break` and leaves the connection up. `meta_sent_` is already `true`, so no later `get_deltas()` re-emits metadata; only a genuine reconnect (`on_connected → reset_meta_send`) re-arms it. The old blocking `restart()` masked this by forcing a reconnect.
*Effect:* the server receives value deltas with no units/zones/displayName until the next reconnect — silent degradation of anything consuming metadata.
*Fix direction:* don't flip `meta_sent_` inside `get_deltas()`; commit it only once `output[0]` actually sends (or re-arm meta when a batch is dropped).

**MINOR — `detach_teardown()` doc contradicts the code** (`signalk_ws_client.h`).
The comment says the OOM-spawn fallback "reaps synchronously"; the code does the opposite — it stashes the handle in `pending_teardown_` and retries the reaper spawn on the next event-loop tick. One-line doc correction.

### Pre-existing on `main` — out of scope here, will file a separate issue

These were surfaced by the review but are not caused by this PR (they live in the `detect_ssl` / CA-pin / reconnect code merged earlier):

- **detect_ssl drops the redirect target host:port** → later TLS legs hit the original (plaintext) port and time out every cycle.
- **`ssl_enabled_` is persisted by an inference and never re-probed** on TLS failure → sticky misconfiguration across reboots.
- **access-request 202/400 without an `href`** falls through to a generic disconnect, losing the pending request.
- **No per-leg watchdog on the connect worker.** Note this PR *widens its blast radius*: a hung leg now also permanently blocks `auth_job_running_` (all future connects), where before it only stalled reconnect. Worth a defensive liveness bound.
- Lower severity: unsynchronized `save()` across contexts; the `tofu_crt_bundle_attach` static-cert re-entrancy (latent, kept serialized by the worker model); `connect_interval_ms_` word-races (benign); `millis()` rollover delaying reconnect ~every 49.7 days (benign); stale "task" wording on `task_connection_state_`.

### Refuted

A claimed worker-vs-transport race on the `tofu_*` strings — not real; the connect worker doesn't write those members.

### Verification status

The fold's loop-stall property is verified on-device. The connected-state soak is still pending: the bench device currently can't open an outbound TCP connection to the SK server (a device-side reachability issue — not SSL, not TOFU, not this PR's code; the connect times out before any handshake). Tracking separately; it doesn't bear on this PR's correctness.

mairas:
## Fixes pushed in response to review

- **MAJOR (metadata loss)** → `397671a` — `send_delta()` now re-arms metadata (`reset_meta_send()`) when the *first* delta's non-blocking send fails. The first delta is the only one that can carry the one-shot metadata `get_deltas()` builds, so this restores metadata delivery on the next batch instead of losing it until the next reconnect. Localized to `send_delta`; no queue API or test change.
- **DOC** → `ba76d38` — corrected the `detach_teardown()` comment: the OOM-spawn fallback stashes the handle in `pending_teardown_` and retries on the next event-loop tick; it does not reap synchronously.

Not addressed here (out of scope for this PR): the pre-existing `detect_ssl` / reliability cluster called out above — will be filed as a separate issue.

Verification update: the cold-token end-to-end flow (access request → approval → poll → connect) is now confirmed on the bench device; it connected immediately on token approval. Earlier connect failures were a stale/un-approved access request on my side, not a code issue. Connected-state soak still to run.

mairas:
## Verification update: connected-state soak ✅ (the last pending item)

Ran on the bench HALSER (ESP32-C3) against the live halosdev signalk-server, plain `ws`, a 200 ms producer generating real delta traffic, a forced `restart()` every 90 s to stress teardown+reconnect on the event loop, and a 1 s heartbeat whose interval drift is the loop-stall detector. ~8.5 min, 504 heartbeats, 5 forced reconnect cycles.

- **No loop stall:** heartbeat `dt` stayed **999–1001 ms (mean 1000), zero outliers** — including across every forced teardown+reconnect. This is the property the fold has to preserve.
- **Reconnect recovery:** 499/504 heartbeats Connected; each forced `restart()` showed Disconnected for ≤1 heartbeat (<2 s to reconnect).
- **No leak:** free heap flat across the 5 reconnect cycles (107.9 KB → 107.1 KB averaged; no per-cycle staircase).
- **Deltas land continuously:** server-side `environment.test.uptime` incremented smoothly (28.7 → 479.3) with the held value always within ~1 s of poll time, through every reconnect.
- **0 crashes, 0 TLS/transport errors.**

Cold-token end-to-end (access request → approval → poll → connect) was also confirmed separately — the device connected immediately on token approval. Both items flagged as "pending before merge" in the PR description are now done.

Issue #1036, opened 2026-06-23
Reduce framework RAM footprint: verified low-hanging optimizations


mairas:
## Summary

A memory audit of the framework (post-#1034) surfaced seven verified RAM wins, all small and low-risk — roughly 20–30 KB combined across persistent and connect-window heap on a TLS-connected device. Each was adversarially verified against ESP32/ESP-IDF reality; myths (`F()`/PROGMEM no-ops, ArduinoJson v7 Static/Dynamic, flash-only "savings") were excluded.

Free heap on a connected wind device sits around 100–110 KB, and the tightest moment is the TLS handshake during connect, so wins in that window matter most.

## Actionable now (no measurement gate)

- [x] **LogCapture handoff queue depth 16→8** (PR #1037) — `src/sensesp/system/log_buffer.h:30` (`kLogCaptureQueueDepth`), allocated in `install()` (`log_buffer.cpp:53`) regardless of whether `/api/log` is ever opened. ~2.1 KB persistent. Only raises the burst-drop rate; the handoff is already non-blocking drop-on-full.
- [x] **TOFU PEM-encode buffer on demand** (PR #1038) — `src/sensesp/signalk/signalk_ws_client.cpp:99`, `static unsigned char pem_buf[4096]` is `.bss` for the whole uptime but written only during the TLS handshake CA encode. Move to a heap allocation scoped to `cert_to_pem()`. Keep it fixed at 4096 — do **not** size from `crt->raw.len` (PEM ≈ 4/3 of DER plus headers, so `raw.len` underflows the buffer and would silently drop CA capture). ~4 KB returned to the heap pool on SSL builds (the connected device).
- [x] **LogBuffer `max_lines` 200→100** (PR #1037) — `src/sensesp/system/log_buffer.h:100`. `records_` (a deque of heap `std::string`) can hold 200 lines during a verbose connect/TLS burst; the 2 s age cap keeps steady-state low, so this trims the **peak** during the heap-pressure window. ~10–17 KB transient. The web log is a ~2 s live tail, not scrollback.
[...]

PR #1037, opened 2026-06-23
perf(system): reduce LogBuffer RAM footprint

mairas:
## Why

Part of #1036 (framework RAM audit). The log capture path holds more RAM than it needs on a memory-constrained device — free heap on an ESP32-C3 connected over TLS sits around 100 KB, and the connect/TLS handshake is the tightest moment.

## What

- **`kLogCaptureQueueDepth` 16 → 8.** The vprintf-hook → capture-task handoff queue is allocated unconditionally in `install()` (~264 B/slot), held for the whole uptime even if `/api/log` is never opened. ~2.1 KB persistent. The handoff is already non-blocking drop-on-full, so a shallower queue only raises the burst-drop rate.
- **`LogBuffer` default `max_lines` 200 → 100.** `records_` can hold up to `max_lines` heap `std::string`s during a verbose connect/TLS burst; the 2 s age cap keeps steady-state low, so this trims the *peak* during the heap-pressure window. The web log is a ~2 s live tail, not scrollback.

## Verification

- Compiles (`pio ci -e pioarduino_esp32`, `examples/ssl_connection.cpp`).
- On-device (HALSER, ESP32-C3): boot + web log tail still functional.

Refs #1036

### DIFF
--- a/src/sensesp/system/log_buffer.h
+++ b/src/sensesp/system/log_buffer.h
@@ -27,7 +27,7 @@ struct LogRecord {
 inline constexpr size_t kLogCaptureLineMax = 256;
 /// Depth of the lock-free handoff queue between the vprintf hook and the
 /// capture task. Bursts beyond this drop lines rather than block the logger.
-inline constexpr size_t kLogCaptureQueueDepth = 16;
+inline constexpr size_t kLogCaptureQueueDepth = 8;
 /// Stack (bytes) and priority of the capture task. The task does the
 /// heap-allocating work the hook offloads, so its stack must be roomy;
 /// priority 1 matches the other SensESP service tasks.
@@ -97,7 +97,7 @@ class LogBuffer {
   // max_line_length defaults to kLogCaptureLineMax to match the capture buffer
   // in vprintf_trampoline(); a larger value cannot recover more than that
   // buffer already captured.
-  explicit LogBuffer(size_t max_lines = 200, uint32_t max_age_ms = 2000,
+  explicit LogBuffer(size_t max_lines = 100, uint32_t max_age_ms = 2000,
                      size_t max_line_length = kLogCaptureLineMax,
                      size_t min_headroom_bytes = kLogCaptureMinHeadroomBytes);
 


mairas:
### On-device verification ✅

Flashed (combined with #1038) to an ESP32-C3 (HALSER):

- Boots clean and runs stable with `kLogCaptureQueueDepth = 8` and `LogBuffer max_lines = 100`.
- `LogCapture` task healthy in the runtime task dump: `hwm_free=3432 bytes` of its 4096-byte stack (unrelated, but confirms the capture path runs).
- No crashes/regressions over repeated connect cycles.

Note: the `/api/log` web tail itself was not exercised in this run (the harness app doesn't open it); the change only reduces buffer sizes, the tail logic is untouched.

Compile: `pio ci -e pioarduino_esp32 examples/ssl_connection.cpp` ✅

mairas:
## Code review — 7 personas (correctness, security, reliability, testing, maintainability, project-standards, adversarial)

**Verdict: ✅ Ready to merge.** No P0/P1. Both changes are pure capacity reductions; the drop/eviction logic is unchanged and correct at the new bounds.

### Confirmed
- **`max_lines` 200→100.** Eviction (`push_line` count-cap loop, age cap) unchanged; the 2 s age cap already dominates retention in practice, so the count reduction rarely binds. No test pins the default (every `LogBuffer` test passes `max_lines` explicitly), so nothing breaks.
- **`kLogCaptureQueueDepth` 16→8.** Handoff is non-blocking `xQueueSend(..., 0)` (drop-on-full), so a shallower queue only moves the drop threshold — no stall, no corruption. Not reachable from host tests (they call `push_line()` directly).

### Advisory (non-blocking)
- **P3 (reliability + correctness, conf 0.62) — silent, uncounted drops.** The connect/WiFi/TLS window is the burstiest logging moment and coincides with peak heap pressure; halving the queue halves burst tolerance there, and there is no drop counter, so a truncated `/api/log` capture is indistinguishable from no output. Suggested optional enhancement: a dropped-line counter surfaced via `/api/log`. Tracked in #1036 (out of scope for this RAM PR).
- **Sub-gate nit (maintainability, conf 0.55)** — the new values carry no inline rationale; the commit body explains both.

Compile ✅ · On-device ✅ (ESP32-C3: boots clean, `LogCapture` task healthy, stable over reconnect cycles).

PR #1038, opened 2026-06-23
perf(signalk): allocate TOFU PEM buffer on demand

mairas:
## Why

Part of #1036 (framework RAM audit). `cert_to_pem()` held a function-static 4096-byte buffer in `.bss` for the device's whole uptime, although TOFU CA capture only runs during the TLS handshake.

## What

Allocate the PEM encode buffer on the heap, scoped to the call, and free it on return. It stays off the small TLS task stack (the original reason for the `static` buffer). The size is kept fixed at 4096 rather than derived from `crt->raw.len` — PEM is base64 (~4/3 of the DER) plus header/footer, so the encoded form is always larger than its DER input, and sizing from `raw.len` would silently truncate the CA. Allocation failure fails safe to "no usable CA", matching the existing encode-error contract.

~4 KB returned to the heap pool on SSL builds.

## Verification

- Compiles (`pio ci -e pioarduino_esp32`, `examples/ssl_connection.cpp`).
- On-device (HALSER, ESP32-C3 → halosdev.local:4430 TLS): TOFU CA capture + pinning intact, no crash, identical PEM output path.

Refs #1036

### DIFF
--- a/src/sensesp/signalk/signalk_ws_client.cpp
+++ b/src/sensesp/signalk/signalk_ws_client.cpp
@@ -16,6 +16,7 @@
 #endif
 
 #include <memory>
+#include <new>
 
 #include "Arduino.h"
 #include "elapsedMillis.h"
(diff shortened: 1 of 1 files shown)


mairas:
### On-device verification ✅

Flashed to an ESP32-C3 (HALSER) connecting to a TLS Signal K endpoint fronted by halos Traefik (`CN=halosdev.hal`, issuer `CN=HaLOS Device CA (halosdev.local)`), clean flash (full erase → empty NVS so TOFU runs first-use capture):

```
I (23970) SKWSClient: VERIFY: cert_to_pem produced 1896 PEM bytes (heap buf)
I (23972) SKWSClient: TOFU: first use, pinning issuing CA (identity halosdev.hal,halosdev.local)
```

- `cert_to_pem()` produces a valid **1896-byte PEM from the heap buffer** on every handshake; TOFU captures and pins the issuing CA with the correct identity.
- No `PEM buffer allocation failed`, no `PEM encode failed`, no crashes/aborts across many reconnect cycles.
- The 1896-byte PEM vs the CA's ~1380-byte DER confirms why the buffer is kept fixed at 4096 rather than sized from `crt->raw.len` — `raw.len` would truncate.
- Free heap recovers to ~118 KB after the handshake (peak dip to ~62 KB during the TLS handshake, as before).

Compile: `pio ci -e pioarduino_esp32 examples/ssl_connection.cpp` ✅

mairas:
## Code review — 7 personas (correctness, security, reliability, testing, maintainability, project-standards, adversarial)

**Verdict: ✅ Ready to merge.** No P0/P1. The security and adversarial reviewers each tried and failed to break the change.

### Why it's safe (security / adversarial / correctness)
- **Byte-identical PEM.** Same `mbedtls_pem_write_buffer` call, same 4096 size, same DER input — only the buffer's storage duration changed.
- **NUL-termination guaranteed.** `mbedtls_pem_write_buffer` writes `'\0'` and zero-fills the tail on success (`pem.c:541-545`); the uninitialized `new[]` buffer is only ever read after a successful encode. On any error path the buffer isn't read.
- **No overflow / no truncated CA pinned.** The encoder size-checks *before* writing (`use_len + add_len > buf_len` → `BASE64_BUFFER_TOO_SMALL`, nothing written), so an oversized cert yields `""`, never a partial/malformed PEM. On-device a real CA encoded to 1896 B (from a ~1380 B DER), ~2× headroom.
- **Fail-safe, not bypassable.** Alloc failure → `""` → caller skips `stash_pending_ca` → falls back to leaf-fingerprint pinning (a *stricter* exact-SHA256 pin), never accept-any. Commit happens only in `on_connected`/`commit_pending_tofu` after an authenticated, key-proven connection, so heap pressure can't be leveraged to persist a bogus anchor.
- **`static` removal is strictly safer.** The old shared static buffer relied on the verify callback never running reentrantly; the per-call `unique_ptr` removes that assumption and frees on every return path (no leak/UAF).

### Advisory (non-blocking)
- **P2 (reliability, conf 0.66) — 4 KB transient at the handshake heap peak.** Adds a 4 KB heap alloc during the TLS handshake (free heap dips to ~62 KB, largest block ~34 KB). Judged defensible because `cert_to_pem` runs **only on first-use CA capture** — once a CA is pinned, the `has_tofu_ca()` early return skips it, so the transient is absent from the steady reconnect-storm path the fragmentation comments concern. **On-device evidence:** in verification the pin never committed (no SK approval), so `cert_to_pem` ran on *every* reconnect cycle and still never hit `MBEDTLS_ERR_SSL_ALLOC_FAILED` — the worst case held.
- **P3 (testing) — alloc-failure branch untested.** Not a coverage regression (`cert_to_pem` is file-static and was never test-reachable); covering it needs an injectable seam. Tracked as optional in #1036.
- **Pre-existing (not this PR)** — `tofu_crt_bundle_attach` uses a function-`static mbedtls_x509_crt pinned_ca` shared across attaches; safe under the single-attempt `auth_job_running_` gate, latent only if concurrent connections are ever added.

Compile ✅ · On-device ✅ (ESP32-C3 → halosdev TLS, valid 1896 B PEM, CA pinned, no crashes/alloc failures).

PR #1039, opened 2026-06-23
feat(system): task-stack high-water reporting + right-size httpd/LogCapture stacks

mairas:
## Why

Closes the measurement-gated items on #1036. The httpd (8192) and LogCapture (4096) task stacks were sized by guesswork; the #1037/#1038 review flagged them as trim candidates but required on-device high-water measurement first. This adds that measurement capability, then uses it.

## What

1. **Task-stack HWM reporting (feat).** `/api/info` gains a `Task stack free (bytes)` group listing each FreeRTOS task's minimum-ever free stack (`uxTaskGetSystemState`), computed live per request, guarded by `CONFIG_FREERTOS_USE_TRACE_FACILITY` (on by default). The status array is `new(std::nothrow)`, over-allocated by a few slots (so a task spawned mid-call can't blank the section), and skipped on alloc failure so `/api/info` still responds. Reset-free stack-headroom visibility — and the basis for the two trims below.
2. **httpd stack 8192 → 6144 (perf).** Deepest path is a config PUT (max-nesting JSON deserialize + SPIFFS save, long URI on the same frame); digest auth runs and returns before the handler, so it never stacks on top.
3. **LogCapture stack 4096 → 3072 (perf).** Capture work is ANSI-strip + `std::string` + a largest-free-block probe.

Net ~3 KB persistent DRAM reclaimed.

## Verification

- Compiles: `pio ci -e pioarduino_esp32` for `examples/ssl_connection.cpp` and `examples/minimal_app.cpp`.
- On-device (ESP32-C3), high-water under sustained load of the **deepest path for each task**:

  | task | before | used / after | headroom |
  |---|---|---|---|
  | httpd | 8192 | 3460 / 6144 (deep config PUT + long URI) | 2684 B free |
  | LogCapture | 4096 | 668 / 3072 | 2404 B free |

  No stack overflow; device stayed up and SK-connected throughout. `/api/info` returns the 14-task group correctly.

Refs #1036

### DIFF
--- a/src/sensesp/net/http_server.h
+++ b/src/sensesp/net/http_server.h
@@ -14,8 +14,14 @@
 #include "sensesp/system/serializable.h"
 #include "sensesp_base_app.h"
 
+// HTTP server worker task stack. Measured worst case on an ESP32-C3 was ~3.4 KB
+// used by a config PUT (max-nesting JSON deserialize + SPIFFS save, with a long
+// request URI on the same frame). The digest-auth path (a 1024-byte
+// Authorization buffer + MD5/String temporaries) runs and returns before the
+// handler, so it never stacks on top of it. 6144 leaves ~2.7 KB of headroom
+// over the deepest measured path.
 #ifndef HTTP_SERVER_STACK_SIZE
-#define HTTP_SERVER_STACK_SIZE 8192
+#define HTTP_SERVER_STACK_SIZE 6144
 #endif
 
 namespace sensesp {
--- a/src/sensesp/net/web/base_command_handler.cpp
+++ b/src/sensesp/net/web/base_command_handler.cpp
@@ -1,8 +1,13 @@
 #include "base_command_handler.h"
 
+#include <freertos/FreeRTOS.h>
+#include <freertos/task.h>
+
 #include <cerrno>
 #include <cstdint>
 #include <cstdlib>
+#include <memory>
+#include <new>
 
 #include "sensesp/net/web/autogen/frontend_files.h"
 #include "sensesp/system/log_buffer.h"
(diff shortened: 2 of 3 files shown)


mairas:
## Code review — 7 personas (correctness, testing, maintainability, project-standards, security, api-contract, reliability)

**Verdict: ✅ Ready to merge.** No P0/P1. The one substantive finding — httpd headroom evidenced by a proxy path — was closed by measuring the true worst case; two cheap fixes applied.

### Addressed in the latest push
- **Reliability (P2) + correctness (P3): the 6144 headroom was measured on the digest-auth path (2628 B), not the deeper config-PUT path.** Measured config-PUT directly on device (max-nesting JSON + long URI): **3460 B used / 6144, 2684 free, no overflow**. config-PUT is indeed the deeper path, but 6144 holds with 44% headroom. Comment + commit message now cite the config-PUT measurement.
- **Correctness (P3, 0.88): TOCTOU — a task spawned between the count and the snapshot would blank the whole section** (`uxTaskGetSystemState` returns 0 on an undersized buffer). Fixed: over-allocate +4 slots, loop over the returned count.
- **Maintainability: redundant `Stack free:` prefix on every row.** Fixed — the row name is just the task name; the unit moved to the group label `Task stack free (bytes)`.

### Acknowledged, not blocking
- **Security (low): `/api/info` runs `uxTaskGetSystemState` (brief scheduler-suspend) per unauthenticated request.** Matches the existing `/api/log` posture (per-request work, unauth by default); single-worker serialization bounds it (~tens of µs for ~15 tasks). Acceptable for a diagnostic endpoint.
- **api-contract (low): duplicate FreeRTOS task names → duplicate rows / React-key warnings.** SensESP/IDF task names are unique in practice; theoretical.
- **Info disclosure:** task names + stack headroom on the unauthenticated `/api/info` is an increment over the heap/uptime/build info already exposed; the unauth-by-default posture predates this PR.

### Verification
- Compiles (`pio ci` pioarduino esp32: `ssl_connection`, `minimal_app`).
- On-device (ESP32-C3): httpd 3460/6144 used (deepest path = config PUT), LogCapture 668/3072 used, no overflow; `/api/info` returns the 14-task group; device stable + SK-connected.

Issue #883, opened 2026-03-25
Nullable<bool> is unusable: false is indistinguishable from invalid


mairas:
## Description

In `types/nullable.cpp:40`:

```cpp
template <>
bool Nullable<bool>::invalid_value_ = false;
```

`is_valid()` checks `value_ != invalid_value_`, so `Nullable<bool>(false)` is always considered invalid. There is no safe sentinel value for `bool` in this design.

The `NullableBool` typedef at `nullable.h:55` advertises this as a usable type, which is misleading.

## Options

1. Remove `NullableBool` typedef and document that `Nullable<bool>` is not supported
2. Add a separate `valid_` flag to `Nullable` (breaking change to the template)
3. Use `std::optional<bool>` instead

---
*Found during code review (2026-03-25)*

PR #1041, opened 2026-06-24
feat(types): support Nullable<bool> via a flag-based specialization

### LINKED ISSUE #1040: Support Nullable<bool> via a flag-based specialization
Follow-up to #883, which forbade `Nullable<bool>` (via a `static_assert`) because the sentinel design has no spare bool value — `false` was indistinguishable from invalid.

That guard is correct but blunt. A tri-state bool (true / false / unknown) is a legitimate need — e.g. NMEA 2000 engine **discrete-status** alarms, where a *stale* alarm should read "unknown," not a stale true/false. `RepeatExpiring<bool>` is the concrete blocked case (it instantiates `Nullable<bool>`).

## Approach — the non-breaking refinement of option 2 from #883

Add an explicit `template <> class Nullable<bool>` carrying a validity flag, instead of changing the primary template:

- The primary template, its sentinel comparison, and the entire `nullable.cpp` sentinel table stay **unchanged** — every non-bool type behaves exactly as today (zero regression surface). The NMEA 2000 "missing data" sentinels keep doing double duty as the wire NA value via `value()`.
- The `bool` specialization is flag-based (`value_` + `valid_`): default-constructed is *invalid* (no value yet); `Nullable<bool>(false)` is *valid false*, distinct from invalid.
- Remove the now-false `static_assert`; restore the `NullableBool` typedef.

## Known tradeoffs

- `invalid()` is asymmetric: numerics return the raw sentinel `T`; `bool` returns a `Nullable<bool>` instance (bool has no spare value to hand back). Harmless to the only generic caller (`RepeatExpiring::repeat_function`).
- A small duplicated interface between the primary template and the bool specialization.

## Tests

- New host test `native/test_nullable_bool`: true / false / invalid / default all distinct; `false` serializes to JSON `false` (not null); invalid serializes to null; round-trips.
- Existing on-target `test_nullable` comment corrected (bool no longer unsupported).

mairas:
## Problem

#883 forbade `Nullable<bool>` with a `static_assert`: the type marks "invalid" with a per-type sentinel *value*, and `bool` has no spare value — so `false` was indistinguishable from invalid. The guard is correct but blunt. A tri-state bool (true / false / unknown) is a real need — e.g. NMEA 2000 engine **discrete-status** alarms, where a *stale* alarm should read "unknown," not a stale true/false. `RepeatExpiring<bool>` is the concrete blocked case.

## Approach

Add an explicit `template <> class Nullable<bool>` that tracks validity with a separate flag, instead of changing the primary template — the non-breaking refinement of option 2 in #883:

- The primary template, its sentinel comparison, and the entire `nullable.cpp` sentinel table are **unchanged** — every non-bool type behaves exactly as before, and the NMEA 2000 "missing data" sentinels keep doubling as the wire NA value via `value()`.
- The `bool` specialization is flag-based (`value_` + `valid_`): default-constructed is *invalid* (no value yet); `Nullable<bool>(false)` is *valid false*, distinct from invalid.
- Remove the now-false `static_assert`; restore the `NullableBool` typedef.

## Tradeoffs

- `invalid()` is asymmetric: numeric specializations return the raw sentinel `T`, while `bool` returns a `Nullable<bool>` instance (bool has no spare value to hand back). The only generic caller, `RepeatExpiring::repeat_function()`, accepts both.
- A small duplicated interface between the primary template and the bool specialization.

## Testing

- New host test `native/test_nullable_bool` (9 cases): true / false / invalid / default all distinct; a valid `false` serializes to JSON `false` (not null); invalid serializes to null; round-trips. Enabled ArduinoJson on the `native` env so `nullable.h` builds on the host.
- On-target `test_nullable`: corrected the stale "unsupported" comment, added a bool parity case.
- Full native suite green (16/16).

Closes #1040

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -270,6 +270,7 @@ build_flags =
 platform = native
 test_build_src = false
 lib_deps =
+    bblanchon/ArduinoJson @ ^7.0.0
 build_flags = -std=c++17 -I src
 ; Override the inherited test_ignore: run host tests, skip on-target system tests.
 test_ignore = system/*
--- a/platformio.ini
+++ b/platformio.ini
@@ -261,9 +261,10 @@ build_flags =
     ${esp32c3.build_flags}
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
-; Host build for pure, dependency-free unit tests (e.g. the TOFU capture
-; decision logic in signalk_tofu.h). Does not build src/ -- which needs the
-; ESP toolchain -- so only the header under test is compiled. Run with:
+; Host build for unit tests that compile on the host toolchain (e.g. the TOFU
+; capture decision in signalk_tofu.h, or the Nullable<bool> type test, which
+; pulls in header-only ArduinoJson). Does not build src/ -- which needs the ESP
+; toolchain. Run with:
 ;   pio test -e native -f "native/*"
 
 [env:native]
--- a/src/sensesp/types/nullable.h
+++ b/src/sensesp/types/nullable.h
@@ -1,8 +1,6 @@
 #ifndef SENSESP_SRC_SENSESP_TYPES_NULLABLE_H_
 #define SENSESP_SRC_SENSESP_TYPES_NULLABLE_H_
 
-#include <type_traits>
-
 #include "ArduinoJson.h"
 
 namespace sensesp {
@@ -16,18 +14,11 @@ namespace sensesp {
  * Nullable holds T{} (0), so it is valid for types whose sentinel differs from
  * 0; use Nullable<T>::invalid() to obtain the invalid value explicitly.
  *
- * Nullable<bool> is intentionally unsupported: bool has only two values, so
- * there is no spare bit pattern to reserve as an invalid sentinel without
- * making one of `true`/`false` indistinguishable from "missing". Use a plain
- * bool, or a wider type if you need a distinct invalid state.
+ * bool has no spare value to reserve as a sentinel, so Nullable<bool> is an
+ * explicit specialization (below) that tracks validity with a separate flag.
  */
 template <typename T>
 class Nullable {
-  static_assert(
-      !std::is_same<T, bool>::value,
-      "Nullable<bool> is not supported: bool has no spare sentinel value. Use a "
-      "plain bool, or a wider type if a distinct invalid state is needed.");
-
   public:
     Nullable() : value_{} {}
     Nullable(T value) : value_{value} {}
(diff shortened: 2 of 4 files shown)


mairas:
## Code review — flag-based `Nullable<bool>` specialization

**Scope:** 4 files, +178/−13 (`nullable.h` + 2 tests + `platformio.ini`), base `ebe9a3a27`.
**Reviewers (6 independent, fresh-context):** correctness, adversarial, api-contract, testing, maintainability, project-standards.

**Headline:** No correctness or adversarial defects. Both reviewers built failure scenarios — ArduinoJson picking its native `bool` converter instead of `convertToJson`, ODR via a forward-declared `Nullable`, overload/move hazards, `RepeatExpiring<bool>` compile — and all were disproven (one verified by compiling against real ArduinoJson) or unreachable. The design is sound; CI is green. Findings below are polish and one genuine test gap — no blockers.

### P2 — Moderate

| # | File | Issue | Reviewer(s) | Conf | Route |
|---|---|---|---|---|---|
| 1 | `platformio.ini:264` | The `[env:native]` comment still says *"pure, dependency-free … only the header under test is compiled,"* but this PR adds ArduinoJson to that env (and `nullable.h` pulls the whole lib). Stale/misleading — a maintainer could strip the dep to "restore" the invariant and break the build. | project-standards, maintainability | 0.92 | safe |
| 2 | `test/native/…test_nullable_bool` | **The motivating use case `RepeatExpiring<bool>` is never instantiated or tested.** The specialization exists to unblock `RepeatExpiring<T>`'s expiry branch (`emit(get().invalid())`, `repeat.h:157`); the tests cover `Nullable<bool>` in isolation but not that integration. Multiple reviewers hand-verified it compiles and behaves, so this is a coverage gap, not a bug. A full expiry test needs ReactESP (on-target); a native compile-assert of the `repeat.h:157` expression is cheap. | testing (+4 as gap) | 0.95 | manual |
| 3 | `nullable.h:90` | `ptr()` returns `bool*` to `value_` only — writing through it doesn't set `valid_`, so `*b.ptr() = true` on a default (invalid) instance yields `value()==true && is_valid()==false`, a state the primary template can't reach (it derives validity from the value; the existing `test_nullable_ptr_mutates_value` relies on that). Latent: no `ptr()` caller exists on bool today. | maintainability | 0.78 | gated |

### P3 — Low

| # | File | Issue | Reviewer(s) | Conf | Route |
|---|---|---|---|---|---|
| 4 | `nullable.h:82` | `operator bool()`/`value()` return `value_` ignoring `valid_`, so `static_cast<bool>(invalid)==false` and `invalid == Nullable<bool>(false)` is **true** — equality / `if(nb)` / ternary contexts silently drop validity. Mirrors the numeric sentinel leak (consistent, not a regression), but `false` is a meaningful value, so the bool case is more dangerous. | testing, adversarial, correctness, maintainability | 0.78 | advisory |
| 5 | `nullable.h:94` | `invalid()` return type diverges from the primary (`Nullable<bool>` vs `T`). Sound today — the only generic caller (`repeat.h:157`) absorbs both — but a future generic caller assuming `invalid()→T` breaks for bool. By design; worth stating the invariant in the doc comment. | api-contract | 0.82 | advisory |
| 6 | `test/native/…:84` | JSON round-trip covers only `false`; no valid `true` to/from JSON. A regression in the truthy `convert*` branch would pass every current test. | testing | 0.80 | safe |
| 7 | `nullable.h` | The bool spec hand-duplicates the ~7-method interface with no "keep in sync" note; a future primary-interface change can silently skip bool. | maintainability | 0.66 | safe (doc) |
| 8 | `nullable.h:83` | Defaulted copy-assign without a defaulted copy-ctor emits `-Wdeprecated-copy` under `-Wall/-Wextra` wherever `Nullable<bool>` is copied (no `-Werror`, so cosmetic). Add `Nullable(const Nullable&) = default;`. | adversarial | 0.70 | safe |

### Pre-existing (action required)

| File | Issue | Route |
|---|---|---|
| CI matrix | The `native` env is **not in CI** (matrix is esp32/esp32c3 × arduino/pioarduino only), so host unit tests — the new `test_nullable_bool` *and* the existing `test_tofu_decision` — never run on CI. The bool tri-state guarantees are validated only locally. Pre-dates this PR. | file-issue (add `native` to CI) |

### Coverage / residual risks

- Removing the `static_assert` relies on the invariant that `Nullable` is never forward-declared elsewhere — true today; a future forward-decl + early `Nullable<bool>` instantiation would silently select the primary (sentinel) template and mismatch ODR against TUs that include the full header.
- JSON output for a valid `false` changes from `null` (the pre-#883 bug) to `false` — a fix, worth a changelog line so anyone who saw the old nulls understands.
- Suppressed: 0 below-threshold findings.

### Verdict: Ready with fixes (no blockers)

No P0/P1; the implementation is correct and CI-green. Recommended:
- **Apply now (safe):** #1 comment fix, #6 valid-`true` test, #8 copy-ctor, #7 sync note; document the #4/#5 contracts.
- **Decide:** #3 `ptr()` on bool — drop it (no bool caller) or set `valid_ = true` on access.
- **Follow-up:** #2 `RepeatExpiring<bool>` test and the native-in-CI issue (fine as fast-follow).

<sub>Reviewed by 6 independent persona agents via `mairas-workflow:review`.</sub>

mairas:
Review feedback addressed in `211b8c500`:

- **#1** — `[env:native]` comment corrected (no longer claims dependency-free).
- **#2** — `RepeatExpiring<bool>` now exercised: instantiated + valid-path checked on-target (`test/system/test_nullable`), plus a native assertion that `invalid()` yields an invalid bool.
- **#3** — `ptr()` marks the value valid on access (write-makes-valid), matching the primary template; covered by `test_ptr_write_marks_valid`.
- **#6** — valid-`true` JSON round-trip added.
- **#7** — "mirrors the primary interface — keep in sync" note added.
- **#8** — defaulted copy constructor added (silences `-Wdeprecated-copy`).
- **#4 / #5** — documented the `operator bool()` validity and `invalid()` return-type asymmetries in the spec doc comment.
- **Pre-existing** (native env not in CI) → filed as #1042.

Native suite 13/13; `system/test_nullable` compiles clean on `arduino_esp32`. CI re-running.

PR #1043, opened 2026-06-25
fix(signalk): make WS buffer size configurable; drop oversize deltas

mairas:
## Problem

On a busy device, the Signal K WebSocket connection drops and reconnects every ~10s, freezing **all** deltas for the backoff. The connection is healthy — the client is tearing it down itself.

`esp_websocket_client` is configured with a hardcoded `buffer_size = 1024`. A delta larger than that is sent as several non-blocking transport writes (deltas use a 0-tick send so they never block the event loop). When a later chunk isn't instantly writable, the short write makes `esp_websocket_client` abort the connection (`transport_poll_write(0)`, `errno=2`, `transport_error=ESP_OK` — no TLS or network error).

Any device whose deltas exceed 1 KB trips this on nearly every connect:
- the `sendMeta=all` metadata delta sent on connect (one entry per path), and
- large value deltas such as a GNSS receiver's full-sky `navigation.gnss.satellitesInView` (~4 KB).

It's load-dependent: a light device (a temperature sensor) never trips it; a GNSS compass whose connect-time delta reaches ~7 KB reconnect-loops continuously.

## Fix

- **Configurable buffer size.** The buffer is now `SENSESP_SK_WS_BUFFER_SIZE`, a build-flag-overridable `#define` (default **1024**). `esp_websocket_client` allocates it for *both* tx and rx, so it stays small by default for memory-constrained boards (ESP32-C3); a device that legitimately sends large deltas raises it: `-D SENSESP_SK_WS_BUFFER_SIZE=8192`.
- **Drop oversize deltas instead of disconnecting.** `send_delta` skips any delta larger than the buffer with an `ESP_LOGE`, rather than attempting the multi-chunk send that aborts the connection. Losing one supersedable delta beats losing the connection and every delta after it.

## Verification

On an ESP32 GNSS compass whose connect-time delta is ~7 KB:
- Default 1024: 6 reconnects / 81 s, deltas frozen in 10–50 s windows.
- `-D SENSESP_SK_WS_BUFFER_SIZE=8192`: **0 reconnects over 75 s, connection stays up, deltas flow continuously (<1 s gaps), 0 oversize drops.**

## Follow-ups (not in this PR)

Splitting multi-path deltas (e.g. capping paths per delta so the metadata delta is chunked) would bound delta size and let the buffer stay small even on devices with many paths.

### DIFF
--- a/src/sensesp/signalk/signalk_ws_client.cpp
+++ b/src/sensesp/signalk/signalk_ws_client.cpp
@@ -6,6 +6,8 @@
 #include <ESPmDNS.h>
 #include <esp_http_client.h>
 
+#include "sensesp/signalk/signalk_ws_delta_size.h"
+
 #ifdef SENSESP_SSL_SUPPORT
 #include <mbedtls/pem.h>
 #include <mbedtls/sha256.h>
@@ -36,6 +38,9 @@ constexpr TickType_t kWsSendTimeoutTicks = pdMS_TO_TICKS(5000);
 // the event loop). 0 = enqueue if the ws-client lock and transport are free
 // right now, otherwise fail fast. Deltas are supersedable. See SignalK/SensESP#1033.
 constexpr TickType_t kWsDeltaSendTimeoutTicks = 0;
+// A device that keeps producing a delta larger than the buffer would drop one
+// every send cycle; rate-limit the warning so it does not flood the log.
+constexpr uint32_t kOversizeDropLogIntervalMs = 10000;
 
 SKWSClient* ws_client;
 
@@ -1404,7 +1409,7 @@ void SKWSClient::connect_ws(const String& host, const uint16_t port) {
   esp_websocket_client_config_t config = {};
   config.uri = url.c_str();
   config.task_stack = kWsTransportTaskStackSize;
-  config.buffer_size = 1024;
+  config.buffer_size = SENSESP_SK_WS_BUFFER_SIZE;
   if (auth_header.length() > 0) {
     config.headers = auth_header.c_str();
   }
@@ -1521,6 +1526,30 @@ void SKWSClient::send_delta() {
       sk_delta_queue_->get_deltas(deltas);
       bool first = true;
       for (const auto& delta : deltas) {
+        if (sk_delta_exceeds_ws_buffer(delta.length(),
+                                       SENSESP_SK_WS_BUFFER_SIZE)) {
+          // Drop the delta to keep the connection alive (signalk_ws_delta_size.h
+          // explains why an oversize delta would otherwise abort it). Unlike the
+          // transient send failure below, an oversize delta is deterministic, so
+          // do NOT re-arm metadata here: get_deltas() already bundles metadata
+          // into the first delta and marks it sent, and re-arming would have
+          // get_deltas() rebuild the same oversize first delta every cycle --
+          // dropped and re-armed forever, never delivered. Leaving it sent lets
+          // the next, metadata-free first delta fit and flow; metadata waits for
+          // a reconnect or a larger SENSESP_SK_WS_BUFFER_SIZE.
+          uint32_t now = millis();
+          if (last_oversize_log_ms_ == 0 ||
+              now - last_oversize_log_ms_ >= kOversizeDropLogIntervalMs) {
+            ESP_LOGW(__FILENAME__,
+                     "Delta too large (%u B > %u buffer); dropped to keep the "
+                     "connection alive -- raise SENSESP_SK_WS_BUFFER_SIZE",
+                     (unsigned)delta.length(),
+                     (unsigned)SENSESP_SK_WS_BUFFER_SIZE);
+            last_oversize_log_ms_ = now;
+          }
+          first = false;
+          continue;
+        }
         int send_result = esp_websocket_client_send_text(
             client_.load(), delta.c_str(), delta.length(), kWsDeltaSendTimeoutTicks);
         if (send_result < 0) {
--- a/src/sensesp/signalk/signalk_ws_client.h
+++ b/src/sensesp/signalk/signalk_ws_client.h
@@ -35,6 +35,17 @@
 #define SENSESP_MAX_RECEIVED_META_UPDATES 20
 #endif
 
+// Signal K WebSocket buffer size in bytes, allocated by esp_websocket_client for
+// both the tx and rx buffer. send_delta drops any delta longer than this rather
+// than let esp_websocket_client split it across non-blocking writes and abort
+// the connection (see signalk_ws_delta_size.h). Override per board via
+// build_flags to hold the largest delta a device sends (e.g. a GNSS receiver's
+// full-sky satellitesInView); kept small by default for memory-constrained
+// boards.
+#ifndef SENSESP_SK_WS_BUFFER_SIZE
+#define SENSESP_SK_WS_BUFFER_SIZE 1024
+#endif
+
 namespace sensesp {
 
 static const char* NULL_AUTH_TOKEN = "";
@@ -402,6 +413,11 @@ class SKWSClient : public FileSystemSaveable,
   TaskQueueProducer<int> delta_tx_tick_producer_{0, event_loop(), 990};
   Integrator<int, int> delta_tx_count_producer_{1, 0, ""};
   Integrator<int, int> delta_rx_count_producer_{1, 0, ""};
+
+  /// @brief millis() timestamp of the last oversize-delta-drop warning, used to
+  /// rate-limit it when a device keeps producing a delta larger than the buffer.
+  /// 0 = never logged, so the next drop logs immediately.
+  uint32_t last_oversize_log_ms_ = 0;
 
   /// @brief A single received delta entry awaiting dispatch on the main task.
   ///
--- a/src/sensesp/signalk/signalk_ws_delta_size.h
+++ b/src/sensesp/signalk/signalk_ws_delta_size.h
@@ -1,0 +1,30 @@
+#ifndef SENSESP_SRC_SENSESP_SIGNALK_SIGNALK_WS_DELTA_SIZE_H_
+#define SENSESP_SRC_SENSESP_SIGNALK_SIGNALK_WS_DELTA_SIZE_H_
+
+#include <cstddef>
+
+namespace sensesp {
+
+/**
+ * @brief True if a delta is too large to hand to esp_websocket_client as a
+ * single tx chunk, so send_delta drops it to keep the connection alive.
+ *
+ * Pure and dependency-free (no Arduino, no esp_websocket_client) so it can be
+ * unit-tested on the host.
+ *
+ * The bound is strictly greater-than, and that is deliberate. esp_websocket_client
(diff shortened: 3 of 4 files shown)


mairas:
## Code Review Results

**Scope:** merge-base `0154eda` → PR head `b9bdb2b` (2 files, +30/−1)
**Intent:** Stop the SK WebSocket client self-disconnecting under load — (1) make `esp_websocket_client` `buffer_size` a build-flag-overridable `SENSESP_SK_WS_BUFFER_SIZE` (default 1024, unchanged); (2) drop deltas larger than the buffer in `send_delta()` instead of letting the multi-chunk non-blocking send abort the connection.
**Mode:** interactive (read-only — no fixes applied)

**Reviewers:** correctness, testing, maintainability, project-standards, reliability, adversarial
- reliability — change is entirely about connection teardown, dropped deltas, and reconnect behavior under load
- adversarial — touches an external integration (esp_websocket_client / SK server) and changes what data is sent; the `first`/`reset_meta_send()` interaction and the size threshold are subtle failure-mode territory

Net: this is a real improvement. The pre-fix behavior froze **all** deltas during every reconnect backoff; with this change value deltas for in-budget paths keep flowing. The findings below are about how completely the fix closes the gap, not a reason to revert it.

### P1 — High

| # | File | Issue | Reviewer | Confidence | Route |
|---|------|-------|----------|------------|-------|
| 1 | `src/sensesp/signalk/signalk_ws_client.cpp:1531` | Oversize first/metadata delta is re-armed every cycle → rebuild-drop-log loop; metadata never delivered | correctness, reliability, adversarial | 0.92 | `manual -> human` |

**#1 detail.** `get_deltas()` bundles every source's metadata into the *first* delta of a batch, gated only by `meta_sent_`. When that first delta exceeds the buffer, the new drop block calls `reset_meta_send()` (sets `meta_sent_ = false`), so the next `get_deltas()` rebuilds the *identical* oversize metadata delta, which is dropped and re-armed again — every send cycle, indefinitely. On a many-path device whose connect-time meta delta exceeds the buffer (the ~7 KB GNSS compass that motivated this PR), metadata (units/zones/displayName) and the values riding in that first delta never reach the server, and an `ESP_LOGE` fires on every cycle. The re-arm looks copied from the transient send-failure path below (line ~1551), where retrying makes sense — but an oversize drop is *deterministic*, so re-arming guarantees a repeat. The PR's stated follow-up (split multi-path deltas) would bound the size, but it does not address the unconditional re-arm. Minimal fix now: when the dropped delta is the meta-bearing first one, do **not** re-arm — either leave `meta_sent_ = true` (accept meta loss, stop the loop, let plain value deltas flow) or chunk metadata out of the value delta. Verify the chosen behavior on a many-path device.

### P2 — Moderate

| # | File | Issue | Reviewer | Confidence | Route |
|---|------|-------|----------|------------|-------|
| 2 | `src/sensesp/signalk/signalk_ws_client.cpp:1524` | Threshold compares payload length only; WS frame + TLS overhead share the tx buffer | reliability, adversarial, maintainability | 0.80 | `gated_auto -> human` |
| 3 | `test/` (send_delta path) | New oversize-drop + meta re-arm branch has no host test, despite the repo's extract-predicate + native-test pattern | testing | 0.88 | `gated_auto -> downstream-resolver` |

**#2 detail.** The guard is `delta.length() > SENSESP_SK_WS_BUFFER_SIZE`, but the same tx buffer must also hold the WebSocket frame header (2–14 B + 4 B client mask) and, on SSL builds, TLS record overhead. A delta whose payload is at or just under `buffer_size` passes the guard yet still frames larger than the buffer, so `esp_websocket_client` splits it into multiple non-blocking writes — the exact abort this PR targets. The fix raises the threshold at which the self-disconnect occurs rather than removing it; a band of sizes just under the buffer still reconnect-loops. Suggested: compare against `SENSESP_SK_WS_BUFFER_SIZE - kWsFrameOverhead` (reserve ~16 B for header+mask, more if TLS). The exact overhead depends on `esp_websocket_client` internals not vendored here, so confirm against its write path — but reserving a conservative margin is safe regardless.

**#3 detail.** The length-vs-threshold + "first delta re-arms meta" decision is a pure predicate, fully separable from `esp_websocket_client`/FreeRTOS. The repo already extracts such predicates from this same file and host-tests them — `should_clear_token_on_status` (`test/system/test_sk_token_clear`) and `tofu_decide_capture` (`test/native/test_tofu_decision`), both from the #1033 series this PR's comment references. Extract the decision and add a native test covering: `len == buffer` sends, `len == buffer+1` drops, an oversize *first* delta re-arms meta, an oversize *later* delta does not.

### P3 — Low

| # | File | Issue | Reviewer | Confidence | Route |
|---|------|-------|----------|------------|-------|
| 4 | `src/sensesp/signalk/signalk_ws_client.cpp:1534` | `ESP_LOGE` on an expected, recoverable drop spams error logs every cycle; sibling drop path uses `ESP_LOGW` | reliability | 0.70 | `safe_auto -> review-fixer` |
| 5 | `src/sensesp/signalk/signalk_ws_client.h:38` | Drop rationale duplicated across the header `#define` and two cpp sites; project favors rare/short comments | maintainability | 0.72 | `advisory -> human` |
| 6 | `src/sensesp/signalk/signalk_ws_client.cpp:1531` | `first` conflates "first iteration" and "carries metadata"; re-arm guard duplicated between drop and failure paths | maintainability, adversarial | 0.62 | `advisory -> human` |
| 7 | `src/sensesp/signalk/signalk_ws_client.h:44` | Comment says buffer governs "both tx and rx", but rx enforces a separate hardcoded `kMaxWsMessageSize = 4096` | project-standards | 0.70 | `safe_auto -> review-fixer` |

Notes: #4 — downgrade to `ESP_LOGW` to match the adjacent transient-drop log and rate-limit so a persistently-oversize delta doesn't flood the log at the send rate. #6 — resolve together with #1 (if #1 stops re-arming in the drop path, the flag's meaning must stay coherent with the failure path). #7 — raising `SENSESP_SK_WS_BUFFER_SIZE` will not lift the rx app-level limit, so the comment overstates the knob's effect; clarify it.

### Pre-existing Issues (action required)

| # | File | Issue | Reviewer | Action |
|---|------|-------|----------|--------|
| 1 | `signalk_ws_client.cpp:339,517` | Inbound server messages larger than the rx buffer / `kMaxWsMessageSize` arrive as multiple WS data fragments and are each parsed standalone (no reassembly via `payload_offset`/`payload_len`), so each fragment fails `deserializeJson` and the message is dropped | reliability, adversarial, correctness | file-issue (substantial, pre-existing; this PR makes it newly reachable by inviting operators to shrink the buffer) |

### Coverage

- Suppressed: several sub-0.60 observations folded into residual risks below.
- Residual risks:
  - `buffer_size` is a compile-time `#define`, so a device hitting the oversize-drop path in the field can't be corrected without reflashing, and there's no runtime counter for dropped deltas — only logs.
  - Raising the value doubles the allocation (tx + rx); `connect_ws` doesn't appear to check whether the larger buffer actually allocated.
  - The fix trades a reconnect loop for silent data loss on default-buffer devices: any install whose connect-time meta delta exceeds 1024 and that doesn't set the build flag now loses metadata permanently (see #1) instead of reconnect-looping.
  - `String::length()` is byte length and matches `esp_websocket_client_send_text`'s byte-length argument — confirmed *not* a multi-byte-threshold problem.
- Testing gaps: oversize-drop branch (boundary at `==` / `==+1`), first-delta meta re-arm vs later-delta no-re-arm, and an all-oversize batch (loop completes sending nothing) are all uncovered.

---

> **Verdict:** Ready with fixes
>
> **Reasoning:** The change is a net improvement and doesn't regress below prior behavior, so it's not a revert. But #1 (the per-cycle re-arm/drop/log loop) is a small, high-value change worth folding in — it stops a pointless loop and makes "metadata delivery on large-meta devices is deferred to the chunking follow-up" an explicit, quiet outcome rather than an error-log storm. #2 closes the disconnect window the fix otherwise only narrows. #3 fills the test gap using the pattern already in this repo.
>
> **Fix order:** #1 re-arm behavior → #2 framing margin → #3 host test → #4/#7 log + comment cleanups. File a follow-up issue for the pre-existing rx-reassembly gap.

mairas:
## Review findings addressed

Force-pushed `fb8794022` (amended onto the original commit to keep one clean commit, not a fix-on-top that leaves the superseded re-arm in history).

| # | Finding | Disposition |
|---|---------|-------------|
| 1 | P1 — oversize first/metadata delta re-armed every cycle | **Fixed.** The drop path no longer calls `reset_meta_send()`. An oversize delta is deterministic, so re-arming just had `get_deltas()` rebuild the same oversize first delta forever. Leaving metadata marked sent lets the next, metadata-free first delta fit and flow; metadata waits for a reconnect or a larger buffer. Comment explains the deliberate divergence from the transient-failure path. |
| 2 | P2 — threshold ignores WS frame + TLS overhead | **Not a defect — verified against `esp_websocket_client` source.** `esp_websocket_client_send_with_exact_opcode` copies up to `buffer_size` **payload** bytes per write (`need_write > buffer_size` is the split test); the WS frame header is framed separately in `esp_transport_ws_send_raw`, not in `tx_buffer`. So a delta of exactly `buffer_size` is sent as one FIN chunk and never enters the multi-write path — `> buffer_size` is the exact boundary. Subtracting a frame-header margin would wrongly drop single-chunk deltas. Locked with a test asserting `len == buffer` sends. |
| 3 | P2 — no test for the new branch | **Added.** Extracted the decision into a pure predicate `sk_delta_exceeds_ws_buffer()` (`signalk_ws_delta_size.h`) with a host test (`test/native/test_ws_delta_size`), mirroring `test_tofu_decision`. Covers boundary at `==`, `==+1`, empty, and an overridden buffer size. |
| 4 | P3 — `ESP_LOGE` spam every cycle | **Fixed.** Downgraded to `ESP_LOGW` (matches the sibling drop path) and rate-limited to once per 10 s so a persistently oversize delta can't flood the log. |
| 5 | P3 — drop rationale duplicated 3× | **Trimmed.** The mechanism explanation now lives once in `signalk_ws_delta_size.h`; the `#define` comment and the `send_delta` comment point to it and each carry only their unique concern (the `send_delta` comment now explains the no-re-arm decision). |
| 6 | P3 — `first` flag conflates iteration / metadata | **Addressed via #1.** The drop path no longer re-arms, so the failure-path `if (first)` and the drop path are now coherent; the comment makes the divergence explicit. |
| 7 | P3 — "both tx and rx" comment vs separate 4096 rx cap | The "both tx and rx" statement is accurate (`esp_websocket_client` allocates `buffer_size` for both). The substantive issue — the inbound `kMaxWsMessageSize = 4096` cap is decoupled from the buffer and large inbound messages aren't reassembled — is filed as #1044. |

Pre-existing inbound-message gap filed as #1044.

### Verification
- `pioarduino_esp32` (SSL) and `arduino_esp32` (non-SSL) builds succeed. The non-SSL build matters here: an early draft put the new include inside the `SENSESP_SSL_SUPPORT` guard; moved out and confirmed against `arduino_esp32`, which doesn't define it.
- `pio test -e native`: 25/25 pass (5 new + 20 existing).

Issue #1045, opened 2026-06-25
Non-blocking log transport to prevent event-loop stalls from synchronous UART writes


mairas:
## Problem

SensESP logs synchronously to UART0: `ESP_LOGx` → the ESP-IDF `vprintf` path → a blocking UART write that busy-waits when the TX FIFO is full. At typical baud rates the FIFO drains slowly, so a burst of log output stalls whatever called the logger — including the main event loop. While the loop is stalled, the Signal K WebSocket client, button handler, and scheduled callbacks all stop running until the UART catches up.

It shows up most visibly with high-volume log lines. Outbound SK deltas log at DEBUG in `src/sensesp/signalk/signalk_delta_queue.cpp:114,165,185`; on a device emitting deltas at a high rate (full-sky GNSS, for example) each per-delta UART write becomes a recurring loop stall. Recent gnss-rtk-compass debugging traced WebSocket disconnects to exactly this. The mechanism is general though — any log line at the active level can block; deltas are just the loudest.

## Mechanism (current code)

- `CORE_DEBUG_LEVEL=ARDUHAL_LOG_LEVEL_VERBOSE` is compiled in (`platformio.ini:41`); the runtime level defaults to `ESP_LOG_DEBUG` via `SetupLogging()` (`src/sensesp_base_app.h:22`).
- `ESP_LOGx` output goes through `esp_log_set_vprintf` → UART0: synchronous, no bounded buffer, blocks when the FIFO is full.

## Proposed approach

Replace the synchronous transport with a non-blocking one:

- A bounded ring buffer that the log path writes into.
- A dedicated low-priority FreeRTOS task that drains the buffer to the UART.
- On overflow, drop (optionally counting dropped bytes) rather than block.

This decouples log volume from event-loop latency: the loop never waits on the UART, and the worst case is lost log lines instead of a stalled device.

## Tradeoff: memory footprint

The drain task needs its own FreeRTOS stack (a few KB) plus the ring buffer in RAM. On memory-constrained targets — single-core ESP32-C3, devices already tight on RAM — that is a meaningful cost. So this must be:

- Configurable: buffer size and task stack size.
- Compile-out-able back to the current synchronous path for tight builds.
- Documented with a clear RAM/stack cost so each project can decide.

## Near-term mitigation (separate)

[...]

PR #1046, opened 2026-06-25
Log Signal K deltas at verbose level

mairas:
## Why

Outbound Signal K deltas are logged at DEBUG and written synchronously to UART0. Under the default DEBUG runtime level they print on every delta, and on a high-rate device the blocking UART write stalls the event loop when the TX FIFO fills. That loop stall starves the SK WebSocket client itself, turning heavy logging into dropped deltas and disconnects.

## What

Move the three delta log calls in `signalk_delta_queue.cpp` from `ESP_LOGD` to `ESP_LOGV`. Deltas are then suppressed at the default DEBUG level but still available when the runtime level is raised to VERBOSE for debugging — they remain the primary "is my data going out?" diagnostic, just off by default.

## Evidence

Validated on a live AIS device (ESP32-C3) carrying real bus traffic. Together with quieting the application's own per-message logs, the device went from loop-starved — HTTP API answering 0/8, SK WebSocket chronically "not connected", every delta dropped — to healthy: HTTP 8/8, WebSocket connected, deltas and N2K both flowing steadily, no reboot.

This is the cheap mitigation. The synchronous, blocking UART log transport is the underlying cause; anyone who raises the runtime level to DEBUG/VERBOSE re-triggers the stall. #1045 tracks the real fix (a non-blocking log transport).

Refs #1045.

### DIFF
--- a/src/sensesp/signalk/signalk_delta_queue.cpp
+++ b/src/sensesp/signalk/signalk_delta_queue.cpp
@@ -111,7 +111,7 @@ void SKDeltaQueue::get_deltas(std::vector<String>& output) {
     String delta;
     serializeJson(json_doc, delta);
     output.push_back(std::move(delta));
-    ESP_LOGD(__FILENAME__, "delta: %s", output.back().c_str());
+    ESP_LOGV(__FILENAME__, "delta: %s", output.back().c_str());
     return;
   }
 
@@ -162,7 +162,7 @@ void SKDeltaQueue::get_deltas(std::vector<String>& output) {
 
     String delta;
     serializeJson(json_doc, delta);
-    ESP_LOGD(__FILENAME__, "delta: %s", delta.c_str());
+    ESP_LOGV(__FILENAME__, "delta: %s", delta.c_str());
     output.push_back(std::move(delta));
   }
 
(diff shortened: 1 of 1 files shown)


mairas:
## Review — independent persona fan-out (4 reviewers, fresh context)

**Verdict: ready to merge.** CI all-green. No P0/P1.

### Confirmed
- **Completeness.** The three `delta: %s` lines are the only per-delta-batch logs in `SKDeltaQueue::get_deltas`. `send_delta()` (signalk_ws_client.cpp) has no per-delta log; `signalk_put_request.cpp:48` is user-action-triggered (low frequency, correctly DEBUG); the inbound `Websocket payload received` log is behind `#ifdef SIGNALK_PRINT_RCV_DELTA` (compiled out by default). No sibling high-frequency site was missed.
- **Right level.** Matches the documented convention (migration guide: `VERBOSE` = "Bigger chunks of debugging information, or frequent messages which can potentially flood the output"). A serialized per-delta JSON payload is exactly that. VERBOSE is the lowest level that actually suppresses, since the runtime default is `ESP_LOG_DEBUG` (`sensesp_base_app.h:22`).
- **No dependence.** No native test, tooling, doc, or example asserts on or greps the `"delta: "` string or its level.

### P3 — advisory (no merge block)
With `CORE_DEBUG_LEVEL=VERBOSE` the three strings stay compiled in, so raising the runtime level to VERBOSE reintroduces the loop-blocking flood this PR fixes. That's inherent to the stopgap — #1045 tracks the real non-blocking-transport cure. Optional: a one-line comment at the first `delta:` line noting these fire per delta batch and block on UART0, so a future maintainer raising the level understands the cause.

### Advisory
`CHANGELOG.md` nominally "lists each PR as merged," but the file is de-facto abandoned (newest entries ~2020). Not applied.

PR #1047, opened 2026-07-06
feat(signalk): SKPrefixListener for path-family subscriptions

dirkwa:
## Problem

`SKValueListener` and `SKMetadataListener` each match a single **exact** path. There's no way to observe a *family* of paths whose members appear and disappear at runtime and can't be enumerated in advance.

The canonical case is **`notifications.*`**: a Signal K server raises alarms on arbitrary paths (`notifications.mob`, `notifications.nmea.alarm.navigational.*`, engine alarms, etc.) that a client cannot know ahead of time. A UI that surfaces alarms has to watch the whole subtree, but today that requires registering a listener per exact path — impossible when the paths are dynamic.

## Approach

A minimal, listener-shaped addition that composes with the existing dispatch, rather than a new callback surface on `SKWSClient`:

1. **`SKListener::matches(path)`** — a virtual that defaults to the existing exact compare (`sk_path == path`). `process_received_updates` now calls `listener->matches(path)` instead of comparing paths inline, so a listener owns its own routing policy. Behaviour for existing exact-path listeners is unchanged.

2. **`SKPrefixListener`** — subscribes with the wildcard `<prefix>*` and overrides `matches()` to prefix-match. Because a prefix listener serves many paths, it emits `SKPathValue{path, value}` so the consumer knows which path a delta arrived on.

```cpp
auto nl = std::make_shared<SKPrefixListener>("notifications.");
nl->connect_to(new LambdaConsumer<SKPathValue>([](const SKPathValue& pv) {
  // pv.path e.g. "notifications.mob"; pv.value the notification object
}));
```

## Notes

- No change to `SKValueListener`/`SKMetadataListener`, and put listeners are untouched (`SKPutListener` doesn't derive from `SKListener`, so its loop keeps the direct compare).
- Fires through the normal `process_received_updates` path (on the event loop), same as every other listener.
- Relies on the server honouring a `path: "<prefix>*"` subscription, which the Signal K subscription protocol supports.

Verified on hardware (ESP32-P4): an `SKPrefixListener("notifications.")` reliably delivers `notifications.*` alarms to a UI overlay, and exact-path value/meta listeners continue to work unchanged alongside it.

### DIFF
--- a/src/sensesp/signalk/signalk_listener.h
+++ b/src/sensesp/signalk/signalk_listener.h
@@ -51,6 +51,18 @@ class SKListener : virtual public Observable, public FileSystemSaveable {
 
   int get_listen_delay() { return listen_delay; }
 
+  /**
+   * Whether a received delta on `path` should be routed to this listener.
+   *
+   * Defaults to an exact match against `sk_path`. Listeners that observe a
+   * *family* of paths (e.g. a prefix subscription like `notifications.*`,
+   * where the server publishes on paths the client cannot enumerate in
+   * advance) override this to widen the match. `SKWSClient::process_received_updates`
+   * calls it per delta instead of comparing paths directly, so the routing
+   * policy lives with the listener.
+   */
+  virtual bool matches(const String& path) const { return sk_path == path; }
+
   virtual void parse_value(const JsonObject& json) {}
 
   /**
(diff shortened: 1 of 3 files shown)


dirkwa:
Reopening from a branch scoped to just this change — the subscription-dedupe fix that was on the same branch is going out as its own PR.

PR #1048, opened 2026-07-06
feat(signalk): SKPrefixListener for path-family subscriptions

dirkwa:
## Problem

`SKValueListener` and `SKMetadataListener` each match a single **exact** path. There's no way to observe a *family* of paths whose members appear and disappear at runtime and can't be enumerated in advance.

The canonical case is **`notifications.*`**: a Signal K server raises alarms on arbitrary paths (`notifications.mob`, `notifications.nmea.alarm.navigational.*`, engine alarms, etc.) that a client cannot know ahead of time. A UI that surfaces alarms has to watch the whole subtree, but today that requires registering a listener per exact path — impossible when the paths are dynamic.

## Approach

A minimal, listener-shaped addition that composes with the existing dispatch, rather than a new callback surface on `SKWSClient`:

1. **`SKListener::matches(path)`** — a virtual that defaults to the existing exact compare (`sk_path == path`). `process_received_updates` now calls `listener->matches(path)` instead of comparing paths inline, so a listener owns its own routing policy. Behaviour for existing exact-path listeners is unchanged.

2. **`SKPrefixListener`** — subscribes with the wildcard `<prefix>*` and overrides `matches()` to prefix-match. Because a prefix listener serves many paths, it emits `SKPathValue{path, value}` so the consumer knows which path a delta arrived on.

```cpp
auto nl = std::make_shared<SKPrefixListener>("notifications.");
nl->connect_to(new LambdaConsumer<SKPathValue>([](const SKPathValue& pv) {
  // pv.path e.g. "notifications.mob"; pv.value the notification object
}));
```

## Notes

- No change to `SKValueListener`/`SKMetadataListener`, and put listeners are untouched (`SKPutListener` doesn't derive from `SKListener`, so its loop keeps the direct compare).
- Fires through the normal `process_received_updates` path (on the event loop), same as every other listener.
- Relies on the server honouring a `path: "<prefix>*"` subscription, which the Signal K subscription protocol supports.

Verified on hardware (ESP32-P4): an `SKPrefixListener("notifications.")` reliably delivers `notifications.*` alarms to a UI overlay, and exact-path value/meta listeners continue to work unchanged alongside it.

### DIFF
--- a/src/sensesp/signalk/signalk_listener.h
+++ b/src/sensesp/signalk/signalk_listener.h
@@ -51,6 +51,18 @@ class SKListener : virtual public Observable, public FileSystemSaveable {
 
   int get_listen_delay() { return listen_delay; }
 
+  /**
+   * Whether a received delta on `path` should be routed to this listener.
+   *
+   * Defaults to an exact match against `sk_path`. Listeners that observe a
+   * *family* of paths (e.g. a prefix subscription like `notifications.*`,
+   * where the server publishes on paths the client cannot enumerate in
+   * advance) override this to widen the match. `SKWSClient::process_received_updates`
+   * calls it per delta instead of comparing paths directly, so the routing
+   * policy lives with the listener.
+   */
+  virtual bool matches(const String& path) const { return sk_path == path; }
+
   virtual void parse_value(const JsonObject& json) {}
 
   /**
--- a/src/sensesp/signalk/signalk_prefix_listener.h
+++ b/src/sensesp/signalk/signalk_prefix_listener.h
@@ -1,0 +1,81 @@
+#ifndef SENSESP_SRC_SENSESP_SIGNALK_SIGNALK_PREFIX_LISTENER_H_
+#define SENSESP_SRC_SENSESP_SIGNALK_SIGNALK_PREFIX_LISTENER_H_
+
+#include <ArduinoJson.h>
+
+#include "sensesp/system/valueproducer.h"
+#include "signalk_listener.h"
+
+namespace sensesp {
+
+/**
+ * @brief A value+path pair emitted for a delta on a prefix-matched path.
+ *
+ * Unlike `SKValueListener` (one exact path, value only), a prefix listener
+ * serves a family of paths, so the consumer needs to know *which* path a
+ * delta arrived on. `path` is the full received path; `value` is a read-only
+ * view of that delta's `value` member — valid only for the duration of the
+ * observer call (it aliases the receive-queue document), so copy anything you
+ * need to retain.
+ */
+struct SKPathValue {
+  String path;
+  JsonVariantConst value;
+};
+
+/**
+ * @brief A ValueProducer that listens to every Signal K value delta whose
+ * path begins with a given prefix, emitting the `{path, value}` pair.
+ *
+ * `SKValueListener`/`SKMetadataListener` match a single exact path. Some
+ * consumers must observe a *family* of paths whose members appear and
+ * disappear at runtime and cannot be enumerated in advance — the canonical
+ * case is `notifications.*`, where a Signal K server raises alarms on
+ * arbitrary paths. This listener subscribes with the wildcard `<prefix>*`
+ * and routes any delta whose path starts with `<prefix>` (via the overridden
+ * `matches()`), so `SKWSClient::process_received_updates` fans matching
+ * deltas to it through the normal listener dispatch.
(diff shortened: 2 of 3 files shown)


mairas:
Sorry for the delay! A very useful addition. LGTM but I'll let Claude loose at the PR before merging.

dirkwa:
Sure, concider maybe adding CodeRabbit for this Repo too. 
I have the PR code already in production for a while and it's been a great addition.

mairas:
## Review

Nice, well-scoped addition. The `matches()` virtual is a clean extension point and behaviour for existing exact-path value/meta listeners is genuinely preserved (`matches()` defaults to the same exact compare, and put listeners are untouched). A few things worth addressing.

### 1. `SKPathValue::value` dangles after the delta is processed — correctness

`parse_value` (`signalk_prefix_listener.h:72`) emits `SKPathValue{path, json["value"]}`, where `json["value"]` is a `JsonVariantConst` aliasing the receive-queue document. `emit()` stores that struct into `ValueProducer::output_`, and `process_received_updates` then calls `received_updates_.pop_front()` (`signalk_ws_client.cpp:715`), freeing the backing document. From that point `output_.value` — and any copy a consumer kept — references freed memory.

The doc comment says the value is valid "only for the duration of the observer call", but that contradicts the `ValueProducer` contract every other producer honours: `get()` is public and returns `output_` at any time, and a consumer is free to retain the emitted struct (the exact "UI overlay marshalled onto the event loop" the PR targets). Confirmed with AddressSanitizer against ArduinoJson 7.4.1 — retaining a copy past the source document's lifetime is a `heap-use-after-free`.

`SKMetadataListener` already solved the identical problem correctly: `SKMetaView` owns the backing document via `std::shared_ptr<const JsonDocument> raw` (`signalk_metadata_listener.h:59`, `:150`), so copies are safe. `SKPathValue` should own its value the same way. I'll push a commit that does this (maintainer edits are enabled).

### 2. Temporary `String` per listener in the value hot path — efficiency (minor)

In the value branch, `path` is a `const char*` (`signalk_ws_client.cpp:698`), and `matches(const String&)` is now called per listener (`:703`). Each call constructs a throwaway `String` from the `const char*` — one heap alloc/free **per listener, per value delta**. The old `get_sk_path().equals(path)` used `String::equals(const char*)`, which allocated nothing.

Cheap fix: hoist a single `String path = ru.doc["path"].as<String>();` before the loop (the meta branch already does exactly this at `:684`) and pass that. One construction per delta instead of N.

### 3. No tests

`matches()` — the base exact default and the prefix override — is pure logic and easy to cover. The lifetime fix in #1 is also testable (retain a copy, drop the source doc, assert the value still reads). Worth adding alongside the fix.

### Minor notes

- **Server dependency undocumented.** The mechanism rests on the server honouring a partial-wildcard subscription (`path: "notifications.*"`); the client-side `matches()` only filters deltas the server already chose to send. Worth a line in the class doc noting the required server behaviour (you've verified it against signalk-server on hardware — good).
- **Notification latency.** Default `listen_delay=1000` and no subscription `policy` means alarms (MOB) can be delayed/batched up to ~1s under the server's default policy. Pre-existing SensESP behaviour, not introduced here, but for the canonical alarm use case a note steering users to lower the delay would help.
- **`startsWith` is a raw string prefix, not path-segment aware.** `SKPrefixListener("notifications")` (no trailing dot) would also match a hypothetical `notificationsX`. The examples correctly use the trailing dot — make that convention explicit in the `@param prefix` doc.
- **Empty prefix.** `prefix == ""` logs an error but then behaves as subscribe-to-everything (`path: "*"`, `startsWith("")` matches all). Either document that as intentional or reject it — the error message implies it's broken when it actually functions.

mairas:
Pushed d640d27 to address finding #1 (the dangling `SKPathValue::value`). `SKPathValue` now owns a `std::shared_ptr<const JsonDocument>` and exposes `value` as a view of it, mirroring `SKMetaView::raw`, so retained copies stay valid.

Verified: the fix is AddressSanitizer/UBSan-clean against ArduinoJson 7.4.1 (the pre-fix alias trips a `heap-use-after-free` on the same test), and the header compiles + links on-target (`pioarduino_esp32`).

Findings #2 (per-listener temporary `String` in `process_received_updates`) and #3 (tests) are left for you — happy to take them on too if you'd like.

dirkwa:
Super, thanks

PR #1049, opened 2026-07-06
fix(signalk): dedupe the subscription by path

dirkwa:
## Problem

`subscribe_listeners()` emits one subscribe entry per listener. When a value listener and a metadata listener are registered for the **same path** — the common gauge pattern shown in `SKMetadataListener`'s own docs (`set_units`/`set_zones` from meta, value from an `SKValueListener`) — the subscription frame contains two identical entries for that path.

`sendMeta=all` is a **connection-level** flag (`/signalk/v1/stream?...&sendMeta=all`), so a single subscription for a path already delivers both its value and meta deltas. The duplicate entry is pure waste that grows the subscribe frame — on a display subscribing to ~30 paths with both listeners, the frame doubles.

## Fix

Collapse listeners by path into one subscribe entry, keeping the smallest `period` so the fastest listener's cadence wins. No behaviour change for consumers — the server sends the same deltas — just a smaller, non-redundant subscription.

Verified on hardware (ESP32-P4): a 31-widget layout's subscribe frame dropped from 68 to 34 entries, and all value + metadata listeners keep working.

### DIFF
--- a/src/sensesp/signalk/signalk_ws_client.cpp
+++ b/src/sensesp/signalk/signalk_ws_client.cpp
@@ -6,6 +6,8 @@
 #include <ESPmDNS.h>
 #include <esp_http_client.h>
 
+#include <map>
+
 #include "sensesp/signalk/signalk_ws_delta_size.h"
 
 #ifdef SENSESP_SSL_SUPPORT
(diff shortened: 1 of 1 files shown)


mairas:
## Review

Correct, clean, and a real improvement — good to merge. No bug here; the notes below are all minor.

I verified the load-bearing premise rather than taking it on faith:

- The connection uses `stream?subscribe=none&sendMeta=all` (`signalk_ws_client.cpp:1409-1410`) and `send_meta_enabled_` defaults to `true`, so `sendMeta` is genuinely connection-level — one subscription entry per path delivers both value and meta deltas, and collapsing a value+meta listener pair on the same path loses nothing.
- Delta dispatch (`process_received_updates`) routes by the listener registry + `wants_meta()` + `matches()`, independent of the subscribe frame — so reducing subscribe entries can't change what any listener receives.
- No path is dropped: every listener's `sk_path` is inserted, so every subscribed path survives. Nothing depends on the frame's shape/ordering, and it's built once per connect.
- Compiles + links on-target (`pioarduino_esp32`), confirming `std::map<String,int>` (String's `operator<`) and the C++17 structured binding are fine on the modern core.

### Minor notes (non-blocking)

1. **Min-period-wins is now deterministic.** When listeners share a path with different periods, the fastest cadence applies to all of them. Reasonable and documented — and arguably better than before, since two entries for one path previously left the effective rate up to the server's duplicate-subscription merge semantics. Not a regression; just worth stating in the docs that a value listener with a deliberately large period sharing a path with a faster listener will now deliver at the faster rate.

2. **No automated test.** Same gap as #1048; verified on hardware (68→34 entries). The dedup is pure logic but entangled with the WS send + FreeRTOS, so it's on-target-only as written. It'd become host-testable if the collapse were extracted into a free function over `(path, period)` pairs — also cleaner separation. Optional follow-up.

3. **Nit (optional):** `find()` then `operator[]` does two map lookups; `try_emplace` + a conditional update does it in one. Negligible — runs once per connect.

PR #1050, opened 2026-07-15
feat(wifi): add multi-network WiFi client failover API

KEGustafsson:
## Summary

Adds a first-class multi-network WiFi client API so more than one STA network can be defined **in code**, tried in order with failover — the in-code equivalent of the multi-network list the WiFi web UI already manages.

`WiFiProvisioner` already stored up to `kMaxNumClientConfigs` (3) client networks and failed over between them, but that list could only be populated from a saved (web-UI) config. The builder exposed just a single `set_wifi_client(ssid, password)`. Defining more than one network in code required subclassing `WiFiProvisioner` and installing it via `set_network_provisioner_factory()` — which also skips the `wifi_provisioner_` wiring and so silently drops the WiFi web UI / scan REST handlers.

## What's new

- **`WiFiProvisioner`** gains a constructor taking a `std::vector<ClientSSIDConfig>`, seeding the preset list (skipping entries with an empty SSID/password and honoring the `kMaxNumClientConfigs` cap). The common bring-up is factored into a shared `init_wifi()` helper so both constructors behave identically. Presets still apply only when nothing is saved to flash.
- **`SensESPApp::set_wifi_clients(...)`** stores the list; `setup()` uses the new constructor for the default provisioner when it is non-empty, so `wifi_provisioner_` is set as usual and the WiFi web UI keeps working.
- **`SensESPAppBuilder::set_wifi_clients(...)`** convenience method.

## Usage

```c++
SensESPAppBuilder builder;
auto app = builder.set_hostname("my-device")
                  ->set_wifi_clients({
                      {"PrimarySSID", "password1"},
                      {"BackupSSID", "password2"},
                  })
                  ->get_app();
```

Like `set_wifi_client()`, these presets only seed a device with no saved configuration; a configuration saved through the web UI always wins.

## Backward compatibility

Fully backward compatible. `set_wifi_client()` / the single-SSID constructor and all existing behavior are unchanged.

## Documentation

- Doxygen `@brief`/`@param` on the new constructor and `set_wifi_clients()` builder method.
- `docs/pages/features/index.md`: method-reference table row, a usage snippet in the WiFi section, and a note on the saved-config-wins precedence.

## Testing

- Verified the change applies and builds cleanly.
- New code formatted with clang-format (Google style, per `.clang-format`); reformatting was limited to the lines this change introduces.

## Notes

- `CHANGELOG.md` was intentionally left untouched — it has not been maintained per-PR since 2020 (v0.4.3-era) while the project is at v3.x. Happy to add an entry if maintainers prefer.

### DIFF
--- a/docs/pages/features/index.md
+++ b/docs/pages/features/index.md
@@ -21,6 +21,7 @@ nav_order: 40
 | Method | Description |
 |--------|-------------|
 | `set_wifi_client(String ssid, String password)` | Set WiFi network credentials |
+| `set_wifi_clients(std::vector<ClientSSIDConfig> configs)` | Set multiple WiFi client networks tried in order with failover (up to 3) |
 | `set_wifi_access_point(String ssid, String password)` | Configure the WiFi soft-AP SSID and password |
 | `disable_wifi()` | Disable WiFi entirely (for Ethernet-only setups) |
 | `set_ethernet(config)` | Configure Ethernet provisioner (ESP32-P4 boards) |
@@ -206,6 +207,20 @@ auto app = builder.set_hostname("my-device")
                   ->get_app();
 ```
 
+To define more than one network in code (tried in order with failover, up
+to three), use `set_wifi_clients()` with a list of `ClientSSIDConfig`
+entries instead:
+
+```c++
+SensESPAppBuilder builder;
+auto app = builder.set_hostname("my-device")
+                  ->set_wifi_clients({
+                      {"PrimarySSID", "password1"},
+                      {"BackupSSID", "password2"},
+                  })
+                  ->get_app();
+```
+
 ### Ethernet (ESP32-P4)
 
 Native RMII Ethernet is supported on ESP32-P4 boards with an external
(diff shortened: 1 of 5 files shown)


mairas:
## Review

Correct, well-documented, and good to merge. No bugs — the notes below are all minor and none require a change before merging.

What I verified in context:

- **The refactor is behavior-preserving.** `init_wifi()` is byte-identical to the original single-SSID constructor's tail (client-list padding → mode select → AP/client start → captive-portal DNS), and the single-SSID constructor keeps its head verbatim and just delegates. Existing users get zero behavior change.
- **The new constructor is consistent with the old.** Same `!config_loaded` guard and AP block; `client_enabled_` is only seeded when presets are applied, matching the original — which relies on `from_json` restoring `client_enabled_` from flash. So "a saved config always wins" holds for both paths.
- **Preset seeding is correct.** The cap counts only valid entries (skipped empty entries don't consume a slot; `break` at `>= kMaxNumClientConfigs`), so up to 3 real networks seed. `kMaxNumClientConfigs = 3` matches the docs.
- **It solves the stated problem.** `setup()` uses the vector constructor for the default provisioner, so `wifi_provisioner_` stays wired and the WiFi web UI / scan REST handlers keep working — the exact thing the subclass+factory workaround silently dropped.
- **No overload ambiguity.** The two constructors differ in their 2nd parameter (`String` vs `std::vector<ClientSSIDConfig>`) with no cross-conversion; the only call sites are `setup()`'s own branches. Compiles + links on-target (`pioarduino_esp32`).

### Minor notes (non-blocking)

1. **Duplicated AP block.** The `if (!config_loaded) { ap_settings… }` head is copy-pasted across both constructors. The tail was factored into `init_wifi()`; the AP head could be too (e.g. a small `apply_ap_presets()` helper). Judgment call — fine as-is.
2. **`set_wifi_client()` + `set_wifi_clients()` together:** `setup()` silently prefers the multi-network list and ignores the single pair. Harmless; a one-line debug log when both are set would aid diagnosis.
3. **Open networks:** presets skip any entry with an empty password (same as the single-SSID path), so passwordless networks can't be seeded in code. Pre-existing limitation, not introduced here — worth a doc footnote if it's a supported use case.
4. **No automated test** — `WiFiProvisioner` is on-target only (`esp_wifi`), consistent with the project's model; verified via on-target build + device testing.
5. **Trivial:** the docs hardcode "3"; if `kMaxNumClientConfigs` ever changes, the prose drifts.

PR #1061, opened 2026-08-18
fix: make UIButton actually work in the web UI

humppafreak:
## Summary

`UIButton::add(name, title)` compiled and ran, but never did anything: it self-registered into a static registry, but nothing in the backend served that registry and nothing in the current Vite/Preact frontend rendered it. No button ever appeared in the web UI and no callback ever fired.

This PR wires it up end to end:

- **Backend** (`src/sensesp/net/web/button_handler.{h,cpp}`, new): `GET /api/buttons` lists the `UIButton` registry (`name`, `title`, `must_confirm`); `POST /api/buttons/<name>` looks up the button and calls `notify()`. Registered in `SensESPApp` alongside the other command handlers. Reuses the existing `check_origin` CSRF check (exposed from `base_command_handler.h` instead of duplicated).
- **Routing**: added a `"Control"` route (`/control` → `ControlPage`) to `add_routes_handlers()`, matching the existing Status/System/Log/WiFi/Signal K/Configuration pattern.
- **Frontend** (`frontend/src/pages/Control/index.tsx`, new): fetches `/api/buttons` on mount and renders one `ButtonCard` per entry (reusing the shared `ButtonCard` component from `components/Card.tsx`), respecting `must_confirm` via a confirm dialog, with toast feedback matching the System page's pattern. Wired into `App.tsx`'s route table.
- Regenerated the embedded frontend bundle (`src/sensesp/net/web/autogen/frontend_files.h`).

## Test plan

- [x] Native host test suite (`pio test -e native`) passes
- [x] Frontend type-checks, lints, and builds cleanly (`tsc --noEmit`, `eslint`, `vite build`); confirmed the built bundle contains the new `ControlPage`/`/api/buttons` code
- [x] ESP32 build compiles and flashes to real hardware
- [x] Manual on-hardware verification: registered a `UIButton`, confirmed it appears on `/control`, and confirmed clicking it fires the callback

---
_Generated by [Claude Code](https://claude.ai/code)_

## Summary

This PR adds end-to-end web UI support for `UIButton`.

- Adds `/api/buttons` to list registered buttons.
- Adds `POST /api/buttons/*` to trigger buttons with origin validation.
- Registers button handlers in `SensESPApp`.
- Adds the `/control` route and `ControlPage`.
- Supports confirmation dialogs, pending states, and success or error toasts.
- Regenerates the embedded frontend bundle.


mairas:
## Multi-persona review

**Scope:** all 8 changed files at 902e02a07, diffed against merge base 4ce19e310. The regenerated `frontend_files.h` was checked for consistency, not line by line: the embedded bundle contains the new ControlPage and `/api/buttons` code, and the federation chunk hashes are unchanged from the merge base.

**Reviewers:** 8 independent persona agents, each in a fresh context: correctness, testing, maintainability, project-standards (always on), plus security (new unauthenticated endpoints, user input in the URL path), api-contract (new REST routes, newly exported symbol), reliability (user callbacks in HTTP handler context), and adversarial (data-mutating endpoint). Findings below are merged and deduplicated; the Reviewers column shows who flagged each finding independently.

The feature itself is welcome and the end-to-end wiring is correct: route, page, registry serialization, and CSRF guard reuse all match the existing patterns, and the manual hardware verification is appreciated. The findings below are about execution context, one wire-format bug, and polish.

### P1 — High

| No. | Location | Finding | Reviewers | Conf. |
|---|---|---|---|---|
| 1 | `src/sensesp/net/web/button_handler.cpp:55` | `it->second->notify()` runs every user-registered callback inside the esp_http_server worker task. All other SensESP callbacks (sensors, transforms, timers) run on the ReactESP event loop, so button callbacks will be written without locking; running them on the httpd task races the event loop on any shared state. The httpd stack is 6144 bytes with ~2.7 KB measured headroom (see the comment in `http_server.h`), with no budget for user code, so a callback of real depth can overflow it and panic the device. A slow callback also stalls the single httpd task and freezes the whole web UI until it returns. The sibling state-changing endpoints already solve this: `/api/device/reset` and `/api/device/restart` send the response and defer the action with `event_loop()->onDelay(...)` (`base_command_handler.cpp:62,75`). **Fix:** `auto button = it->second; event_loop()->onDelay(0, [button]() { button->notify(); });` then send the response. ReactESP event queues are mutex-protected, so cross-task `onDelay` is safe. This changes the 200 to mean "click accepted" rather than "callback completed" — the same contract reset/restart already have. | reliability, adversarial, security, correctness, api-contract | 0.95 |

### P2 — Moderate

| No. | Location | Finding | Reviewers | Conf. |
|---|---|---|---|---|
| 2 | `src/sensesp/net/web/button_handler.cpp:55` | `notify()` is called through the map entry without holding a `shared_ptr`. A callback that calls `UIButton::clear_registry()` (documented for clean app restart) or re-registers the same name (`ui_buttons_[name] = new_cmd` overwrites, `ui_button.h:33`) destroys the button while its own `notify()` is still iterating `observers_`: use-after-free. The `shared_ptr` capture in the finding 1 fix closes this too. | adversarial | 0.72 |
| 3 | `src/sensesp/net/web/button_handler.cpp:58` | The POST success body is never sent. `httpd_resp_send(req, "{\"status\":\"ok\"}", 0)` passes an explicit length of 0; esp_http_server calls `strlen()` only for `HTTPD_RESP_USE_STRLEN` (-1). The response goes out as 200, `Content-Type: application/json`, `Content-Length: 0`. The shipped UI only checks `response.ok`, but any client that parses the advertised JSON gets an empty body. **Fix:** `httpd_resp_sendstr(req, "{\"status\":\"ok\"}")`, as the GET handler in the same file already does. | correctness, api-contract, adversarial | 0.95 |

### P3 — Low

| No. | Location | Finding | Reviewers | Conf. |
|---|---|---|---|---|
| 4 | `frontend/src/pages/Control/index.tsx:53` | `buttons` initializes to `[]`, so every visit renders "No controls have been registered by this device's firmware" until the fetch resolves — an affirmatively false claim during load on a slow link. Track a loading flag (or initialize state to `null`) and show the empty-state message only after a successful response with zero entries. | correctness, reliability, adversarial | 0.95 |
| 5 | `frontend/src/pages/Control/index.tsx:95` | The catch path sets `Error triggering "X": …` but the toast body already prefixes `Failed to trigger "X":`, so network failures render both prefixes concatenated. Set only the error itself in the catch block. | correctness, maintainability | 0.95 |
| 6 | `frontend/mock/api-routes.mock.ts:9` | The vite mock dev server defines every other endpoint the pages consume, but there is no `/api/buttons` mock and the `/api/routes` mock body has no Control entry, so the new page cannot be exercised off-device. Add an `api-buttons.mock.ts` (one `must_confirm: true` and one `false` entry) and add Control to the routes mock. The routes mock also omits the existing Log route — pre-existing drift, cheap to fix in the same edit. | testing | 0.85 |
| 7 | `src/sensesp/net/web/button_handler.cpp:18` | `must_confirm` is snake_case; the API's existing multi-word keys are camelCase (`componentName`, `loadPath` in `RouteDefinition::as_json`). The field name freezes on merge. Either rename to `mustConfirm` on both sides and regenerate the bundle, or accept snake_case deliberately. | api-contract | 0.85 |
| 8 | `src/sensesp/net/web/button_handler.cpp:8` | `add_button_list_handler` and `add_button_click_handler` are referenced nowhere outside this file and are not in the header, but have external linkage. The sibling `config_handler.cpp` puts file-local helpers in an unnamed namespace; do the same here. | project-standards | 0.75 |
| 9 | `src/sensesp/net/web/button_handler.cpp:43` | `char name_cstr[url_tail.length() + 1]` is a VLA (nonstandard extension per the Google style the contributing doc mandates) sized from the request URI, on the same frame that currently runs user callbacks. A registered button name is short by construction; use a fixed-size buffer and return 404 early for overlong tails. The same idiom pre-exists in `config_handler.cpp:116,167`. | reliability, project-standards | 0.72 |
| 10 | `src/sensesp/net/web/button_handler.cpp:38` | Third hand-rolled copy of the strip-prefix / strip-query / `urldecode2` sequence (after config GET and PUT, which already diverge from each other). Worth extracting a shared helper next to `urldecode2`; migrating the config handlers can follow separately. | maintainability | 0.72 |
| 11 | `src/sensesp/net/web/button_handler.h:1` | The filename collides with the existing `ButtonHandler` class (`src/sensesp/system/button.h`, the physical boot-button handler), so every grep for "button_handler" now hits two unrelated concepts. `ui_button_handler.{h,cpp}` would match `ui_button.h`. Cheap now, breaking after a release ships the header. | maintainability | 0.70 |
| 12 | `src/sensesp/ui/ui_button.h:33` | `UIButton::add()` silently overwrites duplicate names. While the registry was dead code this was invisible; now the overwritten button disappears from the UI and its callbacks never fire, with no diagnostic. An `ESP_LOGW` on duplicate registration would make the collision visible in the device log. (Unchanged code made user-visible by this PR.) | maintainability | 0.85 |
| 13 | `src/sensesp/net/web/autogen/frontend_files.h` | The bundle regeneration is folded into the feature commit. The repo convention is a standalone "rebuild embedded frontend bundle" commit, which the release workflow in AGENTS.md uses as its rebuild anchor; folding it in makes the release step re-run the build unnecessarily and buries a ~3700-insertion generated blob in a reviewable commit. Splitting it out is a rebase-only change. | project-standards | 0.72 |
| 14 | commit 902e02a07 | The commit is typed `fix:` but adds new API surface: two endpoints, a new page and route, and a newly exported symbol. The release tooling derives the version bump from commit types and would compute a patch bump. Retype as `feat:` (arguable the other way, since an existing documented API was made functional — but the endpoints are new). | project-standards | 0.65 |
| 15 | `src/sensesp/net/web/button_handler.cpp:14` | Advisory: the static registry is read from the httpd task (GET iterates, POST does `find()`) and written from the event-loop task, with no lock. Safe only under the unstated convention that buttons are registered in `setup()` before the server starts. Worth documenting on `UIButton::add()`; the finding 1 fix moves the write side toward the event loop anyway. The same convention already governs the ConfigItem and StatusPageItem registries. | adversarial | 0.62 |

### Pre-existing (action required, both in files this PR already touches)

| No. | Location | Finding | Routing |
|---|---|---|---|
| 16 | `src/sensesp/net/web/base_command_handler.cpp:58,74` | The reset and restart handlers pass the same explicit length of 0, so their message bodies ("Resetting device back to factory defaults…", "Restarting device") have never actually been transmitted. This is the idiom the new endpoint copied. Same one-line fix (`httpd_resp_sendstr` or `HTTPD_RESP_USE_STRLEN`). | fix here |
| 17 | `src/sensesp/net/web/base_command_handler.h:50` | The closing `#endif` comment names `SENSESP_NET_HTTP_COMMAND_HANDLER_H_`; the actual guard is `SENSESP_NET_WEB_BASE_COMMAND_HANDLER_H_`. | fix here |

### Residual risks (noted, not blocking)

- The Origin-vs-Host CSRF model is defeated by DNS rebinding, and `must_confirm` is a browser-side gate only: any LAN client can trigger any button with one curl POST, unauthenticated and unrated. Both match the existing posture of `/api/device/reset` and `/api/device/restart`, so they are design constants of the API, not regressions here — but this PR is the first to route arbitrary user callbacks through that posture, which raises what a successful forgery is worth.
- `urldecode2` converts `+` to space, so a button named with a literal `+` is unreachable for clients sending the path verbatim (the shipped UI is unaffected: `encodeURIComponent` emits `%2B`). Non-UTF-8 bytes in a button name break the round trip and 404 on every click.
- The handler parses the raw `req->uri` while the dispatcher matched the decoded URI; double-encoded prefixes mis-parse the name. Every traced outcome is a 404, never a wrong-button dispatch.
- The POST handler never drains a request body; a non-browser client sending one on a keep-alive socket can corrupt the next request. Shared with the pre-existing reset/restart handlers.
- Exporting `check_origin` from the header means user firmware defining its own `sensesp::check_origin` now collides at link time; the anonymous namespace previously made that impossible. Only one definition exists in-tree.
- The success toast means "the device accepted the click", not "the callback did what the user wanted" — same contract as the restart/reset cards.
- `UIButtonCard` passes `button.title` as both card title and button text, so each card shows the same string twice with an empty body. Worth a look at whether that rendering is intended.

### Testing gaps

- No test covers the new handlers: URI-tail parsing (encoded names, query strip, empty name, unknown name 404) or the GET JSON shape the frontend's `UIButtonInfo` depends on. `test/system/test_registries` covers only the registry itself.
- `check_origin` has no regression test, and this PR makes it a public symbol guarding a third endpoint family.
- ControlPage branches (load error, empty state, confirm cancel, pending disable) are untestable under the current jest setup — consistent with every existing page, an infrastructure gap rather than a PR-specific omission.

### Verdict

**Ready with fixes.** Findings 1–3 should land before merge; they amount to two small changes (the `onDelay` deferral with a `shared_ptr` capture covers findings 1 and 2, `httpd_resp_sendstr` covers finding 3). Findings 4–6 are small frontend fixes worth folding in, and the two pre-existing items are one-liners in files the PR already touches. The remaining P3s are at the author's and maintainers' discretion; finding 7 (field naming) is the one that becomes irreversible on merge, so it deserves a deliberate yes or no now.

mairas:
Applying the findings from the review above directly to this branch (maintainer edit; fix commits on top, no rewrite of the existing commit). Plan, by finding number from the review:

- Findings 1 and 2: send the response, then defer `notify()` to the event loop via `event_loop()->onDelay(0, ...)` with a `shared_ptr` capture, matching the reset/restart handlers.
- Findings 3 and 16: replace the length-0 `httpd_resp_send` calls with `httpd_resp_sendstr` in the button POST handler and the pre-existing reset/restart handlers.
- Finding 4: ControlPage tracks load state and shows the empty-registry message only after a successful fetch.
- Finding 5: the catch path sets only the error text; the toast supplies the prefix.
- Finding 6: add an `/api/buttons` dev-server mock and add the Control (and missing Log) entries to the routes mock.
- Finding 7: rename the API field `must_confirm` to `mustConfirm` on both sides (maintainer decision: camelCase, consistent with `componentName`/`loadPath`).
- Finding 8: move the file-local handler helpers into an unnamed namespace.
- Finding 9: replace the URI-sized VLA with a fixed-size buffer and an early 404 for overlong names.
- Finding 11: rename `button_handler.{h,cpp}` to `ui_button_handler.{h,cpp}` (maintainer decision: avoids the collision with the existing `ButtonHandler` class).
- Findings 12 and 15: `UIButton::add()` logs a warning on duplicate names, and its docs state that registration belongs in setup, before the HTTP server starts.
- Finding 17: correct the stale `#endif` guard comment.
- Rebuild the embedded frontend bundle as its own conventional commit.

Not addressed here: finding 10 (shared URI-tail helper — a refactor that should also migrate the config handlers, out of scope for this PR) and findings 13 and 14 (they concern the existing commit's structure and type; rewriting a contributor's commit is not warranted — noted for the maintainer at merge time, when release notes are derived).

mairas:
The fixes from the review above are implemented. GitHub does not allow maintainer pushes to organization-owned forks (BoatHacks), even with "allow edits by maintainers" enabled, so the commits could not be pushed to this PR's branch directly. They are published on `fix/ui-button-review-fixes` in this repository, stacked directly on top of 902e02a07 — @humppafreak, you can bring them into this PR with:

```
git checkout fix/ui-button-control-page-clean
git pull https://github.com/SignalK/SensESP.git fix/ui-button-review-fixes
git push
```

This fast-forwards your branch; no rebase and no conflicts.

## What changed, by review finding

- Findings 1 and 2 — `3227da440 fix(web): harden UIButton HTTP handlers`. The click handler now sends the response and defers `notify()` to the event loop (`event_loop()->onDelay(0, ...)`) with a `shared_ptr` capture, matching the reset and restart handlers. The same commit bounds the decoded name buffer (finding 9, fixed 64-byte buffer with an early 404 for longer tails), gives the file-local helpers internal linkage (finding 8), and renames the files to `ui_button_handler.{h,cpp}` (finding 11).
- Finding 3 — same commit: the POST success body is sent with `httpd_resp_sendstr`.
- Finding 16 — `805711033 fix(web): send reset and restart response bodies`: the pre-existing length-0 sends in `base_command_handler.cpp` now use `httpd_resp_sendstr` too.
- Finding 17 — `d33bb15ea chore(web): fix stale include guard comment`.
- Findings 12 and 15 — `4f02ca66e fix(ui): warn when a UIButton name is reused`: `UIButton::add()` logs a warning on duplicate names, and its docs state that registration belongs in setup, before the HTTP server starts serving.
- Findings 4 and 5 — `e305a2af5 fix(web): add Control page loading state`: the page shows the shared Card loading placeholder until the fetch settles, and the error toast no longer double-prefixes network failures.
- Finding 7 — `ed61ae2dd fix(web): use camelCase mustConfirm in buttons API`: the field is `mustConfirm` on both sides (maintainer decision).
- Finding 6 — `56b6cec91 chore(web): add Control page dev mocks`: `GET /api/buttons` and `POST /api/buttons/:name` mocks, plus the missing Control and Log entries in the routes mock.
- `f4748b9ef chore(web): rebuild embedded frontend bundle`: regenerated from the updated frontend source; the embedded bundle now contains only the new index chunk, and the federation chunk hashes are unchanged.

## Verification

- Native test suite: 25 of 25 test cases pass (`pio test -e native`).
- ESP32 test-suite compile, same command CI runs: passes (`pio test --without-uploading --without-testing -e pioarduino_esp32`).
- Frontend: `eslint` clean, `prettier` applied, `vite build` succeeds. `tsc --noEmit` reports 4 errors, all in files these fixes do not touch (`App.tsx` federation virtual module, the Home page svg import, two in the System page) — present independently of this work.
- clang-format clean on every edited C++ file.

## Not addressed, deliberately

- Finding 10 (shared URI-tail helper): a refactor that should also migrate the two config handlers; out of scope for this PR.
- Findings 13 and 14 (bundle regen folded into the feature commit; `fix:` type on a commit that adds API surface): both would require rewriting the existing commit on this branch. Left for the maintainer to weigh at merge time — the release tooling derives the version bump from commit types, so the minor-vs-patch question from finding 14 still stands.

mairas:
## Multi-persona re-review (round 2)

**Scope:** branch `fix/ui-button-review-fixes` (the contributor's commit plus the 8 review-fix commits) against main, merge base 4ce19e310. Same 8 personas as round 1, fresh contexts, primed with the round-1 disposition and told to verify the fixes and hunt regressions.

**The round-1 fixes check out.** Independently verified this round: the deferred `notify()` is safe against the installed ReactESP 3.3.1 source (timed queue is mutex-guarded, a 0 ms event fires on the next tick, the captured `shared_ptr` is released on the event-loop task with no lifetime hazard); the decode buffer is fail-closed (`urldecode2` never emits more bytes than it consumes, so nothing can overflow the 65-byte buffer or slip past the gate); the Control page state machine has no stuck or stale state, and the confirm-dialog and pending-flag races are closed; the rebuilt bundle is current (decompressed the embedded index chunk: `mustConfirm` present, zero `must_confirm`, loading branch and confirm dialog present) and the rebuild commit keeps the exact subject the release workflow anchors on.

Round 2 found one moderate defect in a round-1 fix, plus smaller items.

### P2 — Moderate

| No. | Location | Finding | Reviewers | Conf. |
|---|---|---|---|---|
| 1 | `src/sensesp/net/web/ui_button_handler.cpp:51` | The 64-byte gate measures the percent-**encoded** URI tail, but registry names are decoded and `UIButton::add()` enforces no length limit, while the frontend always sends `encodeURIComponent(name)`. Spaces triple to `%20` and non-ASCII UTF-8 expands 6–12 bytes per character, so a legitimately registered name — roughly 22+ non-ASCII characters, or 55 ASCII characters with a few spaces — appears in `GET /api/buttons` and renders in the UI, yet every click 404s. The justifying comment ("Longer names cannot match any registered button") is true only of decoded names. Memory safety is unaffected; this is a silent availability gap for a class of legal names, with no warning at registration. **Fix options:** bound the raw tail at `3 * kMaxButtonNameLength`, decode into a matching buffer, then apply the 64-byte limit to the decoded name (decoding never expands, so the buffer math stays safe); and/or warn or enforce the limit in `UIButton::add()` so the mismatch surfaces at registration. Reword the comment either way. | adversarial, correctness, security, reliability, api-contract, testing, maintainability | 1.00 |

### P3 — Low

| No. | Location | Finding | Reviewers | Conf. |
|---|---|---|---|---|
| 2 | `src/sensesp/net/web/ui_button_handler.cpp:77` | Advisory: each accepted POST queues one `DelayEvent` on ReactESP's unbounded timed queue (throwing `new`, no size cap — verified in ReactESP 3.3.1 source). A scripted flood (no Origin header passes `check_origin`) gets an immediate 200 per request with no back-pressure; with a slow callback the queue grows until allocation aborts, and even without OOM every accepted click eventually fires — a 30-second flood of a relay-toggling button replays hundreds of actuations after the flood stops, which the pre-fix inline model could not do. Extends the round-1 accepted no-rate-limiting residual with a new failure mode. Possible mitigation: coalesce per-button pending notifies, or reject when the queue is deep. | security, adversarial | 0.80 |
| 3 | `src/sensesp_app.h:15` | The renamed include (`ui_button_handler.h`) was left in `button_handler.h`'s old position, breaking the sorted include order that `.clang-format` (`SortIncludes: true`) and the contributing guide require — the round-1 "clang-format clean on edited files" pass did not cover this file. Move it after `static_file_handler.h`; do not reformat the whole file (the current formatter version also reflows three untouched regions). | project-standards | 0.90 |
| 4 | `src/sensesp/net/web/ui_button_handler.h:13` | The public header says the handler "dispatches clicks to the corresponding button's observers", which reads as synchronous; the accepted-then-deferred contract (200 = queued, callbacks run later on the event loop, matching reset/restart) is documented only inside the `.cpp`. Consumers polling for the callback's effect right after a 200 will see stale state. State the semantics in the header doc. | api-contract | 0.85 |
| 5 | `frontend/src/pages/Control/index.tsx:30` | The list fetch has no `AbortController`/cleanup and its error alert has no retry, while the in-tree sibling `ConfigCards` has both (abort on unmount, `isAbortError` filter, retry via a reload token). Harm today is small (Preact ignores setState on unmounted components; the un-aborted request briefly holds one of the ESP32 httpd's few sockets), but two copies of the same pattern now differ for no domain reason. Mirror the `ConfigCards` structure. | maintainability, adversarial | 0.85 |
| 6 | commit 805711033 | The sendstr fix commit also carries clang-format rewraps of untouched regions (include re-sort, comment re-indents in the info and log handlers). The contributing guide: "If you reformat existing code, always do that in separate commits." Either rebase the branch to split format noise from the fix, or waive the rule for this review-fix branch. | project-standards | 0.72 |

### Residual risks (carried or new, not blocking)

- The routing/handling parser differential (dispatcher matches the decoded URI, handler slices the raw one at offset 13) still lands only on 404s — verified again — but any future per-name authorization added before the lookup inherits the mismatch. `%00` still aliases (`foo%00bar` triggers `foo`); relevant only if names ever become an authorization key.
- The frontend re-enables a button when the HTTP response arrives, before the deferred callback runs; rapid sequential clicks queue multiple notifies. Each was a genuinely accepted click, consistent with the documented semantics, but callbacks are not deduplicated.
- The POST mock always succeeds, so the error-toast path cannot be exercised in the dev server without editing the mock. Mock/backend shape parity is by hand-copy; the `mustConfirm` rename was exactly the class of drift nothing would have caught.

### Testing gaps

- The name round trip (`encodeURIComponent` → dispatch → `urldecode2` → registry lookup) and the 64-vs-65-byte boundary have no test anywhere; a boundary test would have caught finding 1 directly. The parse logic has no host-testable seam today — the round-1-deferred shared URI-tail helper extraction is the natural vehicle, and should land with these boundary tests.
- No CI clang-format check exists, so include-order and formatting compliance rest on manual passes (finding 3 is the demonstration).

### Verdict

**Ready with fixes.** Finding 1 should land before merge — it silently breaks a class of legitimate button names and the fix is small. Findings 3 and 4 are trivial. Finding 2 stays advisory; findings 5 and 6 are maintainer's discretion.

mairas:
Round-2 findings are addressed; 4 more commits on `fix/ui-button-review-fixes` (same fast-forward pull as before brings everything in).

- Finding 1 — `1edc2b03e fix(web): bound button names by decoded length`: the handler now accepts encoded tails up to 192 bytes, decodes, and applies the 64-character limit to the decoded name, so names with spaces or non-ASCII characters click correctly. The limit is documented on `UIButton::add()`, and the misleading comment is gone. Same commit documents the accepted-click 200 semantics in the public header (finding 4).
- Finding 3 — `e9f7f91f9 chore: restore sorted include order in sensesp_app.h`.
- Finding 5 — `653117555 fix(web): abort Control page fetch on unmount`: ControlCards now mirrors ConfigCards — `AbortController` with cleanup, `isAbortError` filtering, and a Retry button on load failure.
- `c44581837 chore(web): rebuild embedded frontend bundle` — regenerated from the updated source (index chunk `index-BXYX4DWC.js`).

Maintainer decisions on the rest: finding 2 (unbounded accepted-click queue) stays advisory — it extends the accepted no-rate-limiting residual, and coalescing clicks would change semantics; finding 6 (format rewraps mixed into `805711033`) is waived for this review-fix branch to avoid force-pushing shared history.

Verification: native tests 25 of 25 pass; the CI-equivalent ESP32 test-suite compile passes; eslint and prettier clean; `vite build` succeeds; clang-format clean on the edited files — `sensesp_app.h` checked this time, with its three pre-existing formatter-drift regions deliberately left untouched.

mairas:
Superseded by [PR 1069](https://github.com/SignalK/SensESP/pull/1069).

The head branch of this PR lives in an organization-owned fork, and GitHub's "allow edits by maintainers" does not function for organization forks — verified with both SSH and token pushes — so the two review rounds' fixes could not land on this branch. PR 1069 carries this PR's implementation with the review fixes folded into the feature commit; it preserves the original authorship and credits @humppafreak. The review history stays in this thread.

@humppafreak: the feature is good work and merges via PR 1069 — this close is purely mechanical. For future PRs, a branch in a user-owned fork lets maintainers push review fixes directly.

PR #1062, opened 2026-08-18
fix: block DNS rebinding in check_origin

humppafreak:
`check_origin()` only compared the `Origin` header's authority against the request's own `Host` header. A DNS-rebinding page makes its own hostname resolve to the device's IP after the fact, so its `Origin` and `Host` headers still agree with each other on that attacker-controlled hostname — the `Origin==Host` comparison alone lets it straight through.

This adds a second check: `Host` must also name an address the device actually owns — its mDNS `.local` name, or the live IP of one of its network interfaces (WiFi station, soft-AP, Ethernet). Both checks must pass.

Addresses the CodeRabbit finding on #1061 (DNS-rebinding bypass of the Origin/Host CSRF check).

## Test plan
- `pio test -e native`: 25/25 passed (file isn't compiled for the native target)
- No ESP32 toolchain available in the environment this was written in to run `pio ci`/`pio test -e pioarduino_esp32`; change is confined to `base_command_handler.{h,cpp}` and only adds `WiFi.h`/`ETH.h`-based host validation — please confirm CI compiles it cleanly.

## Summary

This PR adds a web control page for registered `UIButton` components.

- Registers `ControlPage` in the frontend.
- Lists available buttons from `/api/buttons`.
- Supports optional confirmation, pending states, and success or error notifications.
- Adds GET and POST button handlers.
- Validates button actions with `check_origin()` to mitigate DNS-rebinding attacks.
- Requires matching `Origin` and `Host` headers for state-changing requests.
- Restricts `Host` to the device mDNS name or an active WiFi, soft-AP, or Ethernet address.
- Registers the button handlers during application setup.

PR #1075, opened 2026-09-11
fix: block DNS rebinding in check_origin

mairas:
Supersedes [#1062](https://github.com/SignalK/SensESP/pull/1062). That PR's head branch lives in the organization-owned fork `BoatHacks/SensESP`, and GitHub's "allow edits by maintainers" does not work for organization forks, so the rebase and review fixes could not land on its branch. This PR carries the DNS-rebinding fix only, rebased onto current `main`, and credits @humppafreak.

## What changed

`check_origin()` compared the `Origin` header's authority against the request's own `Host` header. A DNS-rebinding page makes its own hostname resolve to the device's IP after the fact, so its `Origin` and `Host` headers still agree with each other on that attacker-controlled hostname; the `Origin == Host` comparison alone lets it through.

This adds a second condition: `Host` must also name an address the device actually owns — its mDNS `.local` name, or the live IP of one of its network interfaces (WiFi station, soft-AP, Ethernet). Both conditions must hold.

## Why this is fix-only

PR [#1062](https://github.com/SignalK/SensESP/pull/1062) also carried the whole UIButton web Control page, because it was branched from the feature work in the now-closed [#1061](https://github.com/SignalK/SensESP/pull/1061). That feature has since merged on its own via [#1069](https://github.com/SignalK/SensESP/pull/1069), so the superseding change is just the `check_origin` hardening. This resolves the CodeRabbit major finding on [#1062](https://github.com/SignalK/SensESP/pull/1062) that asked to keep the DNS-rebinding fix independently reviewable and releasable.

## CodeRabbit findings from the superseded PR

- Major, "separate the Control feature from the DNS-rebinding fix": resolved by construction — this PR touches only `base_command_handler.{h,cpp}`.
- Minor, "remove or replace the echo comments": the three helper comments that restated the code are dropped; the one comment kept is the non-obvious reason the `0.0.0.0` guard exists.
- Minor, "clear the prior error before a new activation" on `Control/index.tsx`: that file belongs to the UIButton feature already merged in [#1069](https://github.com/SignalK/SensESP/pull/1069), not to this fix. The bug is still present on `main` and is filed as a separate follow-up rather than bundled here.

## Test plan

The change is confined to `base_command_handler.{h,cpp}`, which the native test target does not compile, so `pio test -e native` does not exercise it. I did not compile it against the ESP32 toolchain in this environment; the change uses only APIs already used elsewhere in the tree (`SensESPBaseApp::get_hostname()`, `WiFi.localIP()`, `WiFi.softAPIP()`, `ETH.localIP()`). Please confirm the ESP32 CI build compiles it cleanly.

https://claude.ai/code/session_01WYPBh2AySbQgeemaDHM8K2

## Summary

This PR hardens `check_origin()` against DNS rebinding.

- Requires the `Origin` authority to match `Host`.
- Accepts `Host` only when it matches the configured mDNS `.local` name or a live WiFi, soft-AP, or Ethernet address.
- Rejects `0.0.0.0`.
- Updates `check_origin()` documentation.
- Changes only `base_command_handler.{h,cpp}`.

Native tests do not compile the affected files. ESP32 toolchain compilation is not performed in the reported environment.


mairas:
@coderabbitai review

